PR #3, opened 2018-10-22
Fix package imports

matzew:
* package/import fixes
* pointing to Knative `pkg/event` API, for CloudEvents (thjat's in active development and started with the same code than the link in here)


kars7e:
Thanks Matthias!

PR #4, opened 2018-11-07
Make use of TrimPrefix for binary headers

rgee0:
Currently the Header processing in UnmarshalBinary makes use of `TrimLeft` which will have undesirable effects when a character in the cutset immediately follows the hyphen.  For example a header value of `CE-CloudEventsVersion` would be converted to `loudEventsVersion`.

This change amends TestHTTPMarshallerFromRequestBinaryBase64Success to add a header to exercise the identified issue and also changes TrimLeft to TrimPrefix within http.go.

Signed-off-by: Richard Gee <richard@technologee.co.uk>

### DIFF
--- a/v01/http.go
+++ b/v01/http.go
@@ -97,8 +97,8 @@ func (e *Event) UnmarshalBinary(req *http.Request) error {
 			continue
 		}
 
-		k = strings.TrimLeft(k, "Ce-")
-		k = strings.TrimLeft(k, "X-")
+		k = strings.TrimPrefix(k, "Ce-")
+		k = strings.TrimPrefix(k, "X-")
 		k = strings.Replace(k, "-", "", -1)
 
 		fieldName, ok := fieldNames[strings.ToLower(k)]
--- a/v01/http_test.go
+++ b/v01/http_test.go
@@ -24,6 +24,7 @@ func TestHTTPMarshallerFromRequestBinaryBase64Success(t *testing.T) {
 	header.Set("CE-MyExtension", "myvalue")
 	header.Set("CE-AnotherExtension", "anothervalue")
 	header.Set("CE-EventTime", "2018-04-05T17:31:00Z")
+	header.Set("CE-CloudEventsVersion", "0.1")
 
 	body := bytes.NewBufferString("This is a byte array of data.")
 	req := httptest.NewRequest("GET", "localhost:8080", ioutil.NopCloser(body))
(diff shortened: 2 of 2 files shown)


v01/http.go
```diff
@@ -97,8 +97,8 @@ func (e *Event) UnmarshalBinary(req *http.Request) error {
 			continue
 		}
 
-		k = strings.TrimLeft(k, "Ce-")
-		k = strings.TrimLeft(k, "X-")
+		k = strings.TrimPrefix(k, "Ce-")
+		k = strings.TrimPrefix(k, "X-")
```
duglin: Why are we trying to remove X- when line 96 would skip it?

v01/http.go
```diff
@@ -97,8 +97,8 @@ func (e *Event) UnmarshalBinary(req *http.Request) error {
 			continue
 		}
 
-		k = strings.TrimLeft(k, "Ce-")
-		k = strings.TrimLeft(k, "X-")
+		k = strings.TrimPrefix(k, "Ce-")
```
duglin: Given that TrimPrefix does an "if" to see if "Ce-" is there, and we're doing that on line 96, is there any reason we're not just using `k[3:]` and using `Ce-` on line 96?  Actually 96 is wrong, it should have a dash.

v01/http.go
```diff
@@ -97,8 +97,8 @@ func (e *Event) UnmarshalBinary(req *http.Request) error {
...
+		k = strings.TrimPrefix(k, "X-")
 		k = strings.Replace(k, "-", "", -1)
```
duglin: why do this? isn't this buggy since it would make "foo-bar" overwrite "foobar" ?

v01/http.go
```diff
@@ -97,8 +97,8 @@ func (e *Event) UnmarshalBinary(req *http.Request) error {
...
 
 		fieldName, ok := fieldNames[strings.ToLower(k)]
```
duglin: Sorry for extending beyond this PR, but... why do we setup fieldNames at all. It seems  like a lot of extra work when we could just lower-case the incoming property name and use target.FieldByName() below to find it? If it doesn't find it, it'll return an zero Value and we can just skip it then. This just feels like a lot of extra work with very little value.

v01/http.go
tenczar: In the larger scope, this loop is trying to operate on only header values that are marked as CloudEvent headers. If the header is not a CloudEvent header, it is ignored. If it is a CloudEvent header then I extract the raw property name and compare it to the well known CloudEvent fields to set that field on the Event object. If the raw property name is not in the map of known Event fields then it gets stuck in the (internal) extensions bag.

You are right, line 96 is missing a dash.

v01/http.go
tenczar: I think we can safely remove this because we decided that property names are just lowercase ASCII strings.

v01/http.go
duglin: yeah I get all of that, I just think we're doing it the hard way :-)
see: https://github.com/cloudevents/sdk-go/pull/4#discussion_r231731994  too

v01/http.go
tenczar: I did assuming that `FieldByName` was case sensitive. This allows me to map the case insensitive property name extracted from the header to its associated case sensitive version, which I then pass to `FieldByName`. If `FieldByName` is case insensitive then this can be drastically simplified.

duglin: See: https://golang.org/pkg/reflect/#Value.FieldByNameFunc

tenczar:
This has been approved for a long time. I'm going to merge it and address Doug's comments in a forthcoming PR.

v01/http.go
alexellis: @tenczar ^

tenczar: I agree that reflection will impact performance. I made the decision to use reflection to avoid a lot of boilerplate in the marshaller methods that checked for each well known property, make any necessary type conversions and set the associated field. The goal was to create marshaling code that would not need to be updated every time new fields were added to the spec.

PR #11, opened 2018-12-20
Updating marshalers to read and write from event struct and add builder

tenczar:
This change modifies the Marshaler and Converter interfaces to align with the API for json/xml marshaling. Users can now pass an event pointer to directly populate the event passed event. Also adds a builder for v02 events to help validate events.
Signed-off-by: Nick Tenczar <ntenczar@vmware.com>


n3wscott:
Sorry, this no longer applies to the new sdk, gonna close.

PR #14, opened 2019-02-08
Proposal: New SDK

n3wscott:
Rewrote the sdk to solve the issue of how to integrate with other transports other than http.

There will be a lot of discussion. This body of work is a result of lessons learned from working with the [knative/cloudevents](https://github.com/knative/pkg/tree/master/cloudevents) and attempting to rebase to `cloudevents/sdk-go` sdk.


n3wscott:
As a follow-up, I added samples: https://github.com/n3wscott/sdk-go/pull/1/

n3wscott:
> Dep tool is not the official thing. Please switch to `go mod`

Given `go mod` is still not on by default in golang version 1.11. I am inclined to not use `go mod` yet.

denismakogon:
Dep as well, no? Go mod meant to be the de-facto standard since 1.11

n3wscott:
Dep is the current de-facto for almost of golang projects I work in. Go mod is not in use yet and it has some serious drawbacks that are still being addressed.

tenczar:
Directly including source from a project's dependencies in a repo always bugged me. In this case it's weirder because the dependency is only necessary for running tests. I like Mark's suggestion.

n3wscott:
> Directly including source from a project's dependencies in a repo always bugged me. In this case it's weirder because the dependency is only necessary for running tests. I like Mark's suggestion.

I find checking out code that does not work more irritating. `dep` solves this problem, and committing the vendor directory after trimming prevents you from needing to `dep ensure`.

README.md
```diff
@@ -2,64 +2,94 @@
...
 
 ## Existing Go for CloudEvents
 
-Existing projects that added support for CloudEvents in Go are listed below. It's our goal to identify existing patterns
-of using CloudEvents in Go-based project and design the SDK to support these patterns (where it makes sense).
+Existing projects that added support for CloudEvents in Go are listed below. 
+It's our goal to identify existing patterns of using CloudEvents in Go-based 
+project and design the SDK to support these patterns (where it makes sense).
 - https://github.com/knative/pkg/tree/master/cloudevents
 - https://github.com/vmware/dispatch/blob/master/pkg/events/cloudevent.go
```
tenczar: We can probably remove this entry, at least until dispatch can be updated with all of these changes.

cmd/samples/complex/httptonats/main.go
```diff
@@ -0,0 +1,78 @@
...
+	fmt.Printf("forwarding...")
+
+	if err := r.Client.Send(event); err != nil {
+		fmt.Printf("forwarding failed: %s", err.Error())
+	}
+
+	fmt.Printf("----------------------------\n")
+}
+
+func _main(args []string, env envConfig) int {
```
tenczar: It would be nice for the main method to have a comment giving an overview of what the sample is trying to demonstrate.

cmd/samples/complex/sender/main.go
```diff
@@ -0,0 +1,158 @@
...
+		EventID:     uuid.New().String(),
+		EventType:   d.EventType,
+		EventTime:   &types.Timestamp{Time: now},
+		Source:      types.URLRef{URL: d.Source},
+		ContentType: d.ContentType,
+	}
+	return ctx
+}
+
+func _main(args []string, env envConfig) int {
```
tenczar: It would be nice for the main method to have a comment giving an overview of what the sample is trying to demonstrate.

n3wscott: I added a todo to add more documentation.

PR #15, opened 2019-02-13
Add Nats Transport

n3wscott:
This adds the [nats transport binding](https://github.com/cloudevents/spec/blob/v0.2/nats-transport-binding.md) to the sdk.


n3wscott:
oops wrong repo.

PR #21, opened 2019-02-21
Adding nop tests to make coverage show real results for empty dirs.

n3wscott:


n3wscott:
Not a big deal, merging.

PR #23, opened 2019-02-21
Adding validation method.

n3wscott:
Validate pre-flight events. 

This would work to inside the client before it starts encoding a given event, it would have a chance to call validate on the event and throw and error for the user.

Like:

```
err := myClient.Send(ctx, badEvent)
```

`err` here can tell the implementor what is missing in the current event and then react to it.


n3wscott:
Marked Work in Progress. Will do the other 2 versions and update.

n3wscott:
@markpeek Ready for Review. Thanks!

n3wscott:
FYI, total code coverage is at now at 75%.

PR #24, opened 2019-02-22
Make new client calls easy.

n3wscott:
I was not happy with how verbose the simple new client call was. Moving to using ClientOption based configuration.

Old code like: 

```go
c, err := client.NewHttpClient(context.TODO(), "http://localhost:8080/", cloudeventshttp.BinaryV02)
```

Becomes:

```go
c, err := client.NewHttpClient(
	client.WithTarget("http://localhost:8080/"),
	client.WithHttpEncoding(cloudeventshttp.BinaryV02), 
)
```

This becomes really useful when you are making a client that is intended to never send, but just receive, Target becomes optional. 


n3wscott:
Google Internal Reviewers: LGTM

Issue #22, opened 2019-02-21
Auto-generate ID when it is empty.


n3wscott:
It would be nice to do this:

```
		event := cloudevents.Event{
			Context: cloudevents.EventContextV02{
				Type:   "com.cloudevents.readme.sent",
				Source: source,
			},
			Data: data,
		}
```

Rather than this:

```
		event := cloudevents.Event{
			Context: cloudevents.EventContextV02{
				ID:     uuid.New().String(),
				Type:   "com.cloudevents.readme.sent",
				Source: source,
			},
			Data: data,
		}
```

PR #32, opened 2019-02-28
Adding Client Option for default ID and Time when using client.Send()

n3wscott:
fixes: https://github.com/cloudevents/sdk-go/issues/22

Attempting to reduce friction for the producer of events when using the client.

At the moment we need to do something like this to make a event context:

```go
func (d *Demo) context() cloudevents.EventContext {
	now := time.Now()
	ctx := cloudevents.EventContextV01{
		EventID:     uuid.New().String(),
		EventType:   d.EventType,
		EventTime:   &types.Timestamp{Time: now},
		Source:      types.URLRef{URL: d.Source},
		ContentType: &d.ContentType,
	}.AsV01()
	return ctx
}
```

But with this change, we can add new client options to enable defaulting:

```go
c, err := client.NewHTTPClient(
	client.WithTarget(env.HTTPTarget),
	client.WithHTTPEncoding(encoding),
	client.WithUUIDs(),
	client.WithTimeNow(),
)
```

And now the producer of the event has less to think about:

```go
func (d *Demo) context() cloudevents.EventContext {
	now := time.Now()
	ctx := cloudevents.EventContextV01{
		EventType:   d.EventType,
		Source:      types.URLRef{URL: d.Source},
		ContentType: &d.ContentType,
	}.AsV01()
	return ctx
}
```

### DIFF
--- a/cmd/samples/complex/sender/main.go
+++ b/cmd/samples/complex/sender/main.go
@@ -8,7 +8,6 @@ import (
 	cloudeventshttp "github.com/cloudevents/sdk-go/pkg/cloudevents/transport/http"
 	cloudeventsnats "github.com/cloudevents/sdk-go/pkg/cloudevents/transport/nats"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/types"
-	"github.com/google/uuid"
 	"github.com/kelseyhightower/envconfig"
 	"log"
 	"net/url"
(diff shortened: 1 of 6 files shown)


cmd/samples/complex/sender/main.go
```diff
@@ -77,11 +76,8 @@ func (d *Demo) Send() error {
 }
 
 func (d *Demo) context() cloudevents.EventContext {
-	now := time.Now()
```
markpeek: I like the addition of the default ID and Time. But for the purposes of this sample, should there be examples of both using the default (as best practice) and also the longer form for those that might need to assign a specific time?

n3wscott: The simple http sender does this, this is the complex example

Issue #31, opened 2019-02-28
There is no way for the receiver to easily return an error to the sender using the client.


n3wscott:
There is no return on the receive function. Need to think about how to propagate errors up from the handler without coupling it to the http request, because that will not work for nats.

PR #42, opened 2019-03-05
Add EventResponse for send/receive.

n3wscott:
This PR enables control over response codes for a transport (transport dependent). For HTTP this means letting the receiver control which HTTP status code is returned, as well as a response cloudevents.Event object. That response object will be passed through the client's defaulter chain if it the client is used to Send/Receive events ✨.
 
The signature of client.Receive has changed significantly, but in a followup PR I will reduce its complexity once again, and the README will not be wrong with the following:

```go
func Receive(event cloudevents.Event) {
	// do something with event.Context and event.Data (via event.DataAs(foo)
}
```

Until then, I want to leave the README in a broken state, because the samples do work and after the next PR, the README will be correct as well.

Fixes: https://github.com/cloudevents/sdk-go/issues/31


n3wscott:
Note: This is dependent on #45

n3wscott:
HTTP Receive Sample now outputs like:

```
----------------------------
Got Event Context: {CloudEventsVersion:0.1 EventID:de2128dc-3802-43e0-8cde-e8f5b6d7b7a4 EventTime:2019-03-07T21:31:07.350717Z EventType:com.cloudevents.sample.sent EventTypeVersion:<nil> SchemaURL: ContentType:0xc0001a2040 Source:{URL:{Scheme:https Opaque: User: Host:github.com Path:/cloudevents/sdk-go/cmd/samples/sender RawPath: ForceQuery:false RawQuery: Fragment:}} Extensions:map[]}
Got Data: &{Sequence:12 Message:Hello, Binary Encoding v0.1!}
Got Transport Context: {URI:/interesting/ Host:localhost:8080 Method:POST}
----------------------------
```

cmd/samples/http/receiver/main.go
```diff
@@ -32,14 +33,18 @@ type Example struct {
...
-func gotEvent(event cloudevents.Event) {
+func gotEvent(ctx context.Context, event cloudevents.Event, resp *cloudevents.EventResponse) error {
 	fmt.Printf("Got Event Context: %+v\n", event.Context)
 	data := &Example{}
 	if err := event.DataAs(data); err != nil {
 		fmt.Printf("Got Data Error: %s\n", err.Error())
 	}
 	fmt.Printf("Got Data: %+v\n", data)
+
+	fmt.Printf("Got Transport Context: %+v\n", http.TransportContextFrom(ctx))
```
n3wscott: look at this @Harwayne ^^

cmd/samples/http/responder/main.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func _main(args []string, env envConfig) int {
+	ctx := context.Background()
+
+	c, err := http.New(
+		cehttp.WithPort(env.Port),
+		cehttp.WithPath(env.Path),
+		client.WithUUIDs(),
+		client.WithTimeNow(),
```
n3wscott: This is REALLY subtle and awesome. I am setting client options on the http client wrapper, and using those event defaulters to default values on the response event. 💯

pkg/cloudevents/client/client.go
```diff
@@ -47,27 +50,40 @@ func (c *ceClient) Send(ctx context.Context, event cloudevents.Event) error {
...
 	return c.transport.Send(ctx, event)
 }
 
-func (c *ceClient) Receive(event cloudevents.Event) {
-	if c.receiver != nil {
-		c.receiver(event)
+// Receive is called from from the transport on event delivery.
+func (c *ceClient) Receive(ctx context.Context, event cloudevents.Event, resp *cloudevents.EventResponse) error {
+	if c.receive != nil {
+		err := c.receive(ctx, event, resp)
```
n3wscott: client.receive should be mappable to an event type or other interesting filtering. Look into this.

pkg/cloudevents/client/client.go
```diff
@@ -47,27 +50,40 @@ func (c *ceClient) Send(ctx context.Context, event cloudevents.Event) error {
...
+		// Apply the defaulter chain to the outgoing event.
+		if err == nil && resp != nil && resp.Event != nil && len(c.eventDefaulterFns) > 0 {
+			for _, fn := range c.eventDefaulterFns {
+				*resp.Event = fn(*resp.Event)
+			}
+			// Validate the event conforms to the CloudEvents Spec.
+			if err := resp.Event.Validate(); err != nil {
```
n3wscott: AND!!! outbound events are validated.

pkg/cloudevents/transport/http/transport.go
```diff
@@ -85,33 +93,55 @@ func (t *Transport) Send(ctx context.Context, event cloudevents.Event) error {
...
 		req.ContentLength = int64(len(m.Body))
-		return httpDo(ctx, &req, func(resp *http.Response, err error) error {
+		return httpDo(ctx, &req, func(resp *http.Response, err error) (*cloudevents.Event, error) {
 			if err != nil {
-				return err
+				return nil, err
 			}
 			defer resp.Body.Close()
 
+			body, _ := ioutil.ReadAll(resp.Body)
```
n3wscott: This is gathering the response and smashing it into an event if it can.

cmd/samples/http/receiver/main.go
Harwayne: That seems to include the HTTP host, path, and method. Nifty.

PR #44, opened 2019-03-05
Add circleci configuration for test automation

markpeek:
Add an initial circleci configuration to automate testing for PR's and showing test results.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,42 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/golang:1.11.5
+
+    environment:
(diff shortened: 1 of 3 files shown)


n3wscott:
looks good, should there be a `go mod tidy` step?

n3wscott:
Also thanks!! We really need some automation.

markpeek:
I added a step that would run `go mod tidy` and error if there is a difference. I'm assuming tidy might modify both go.mod and go.sum.

n3wscott:
>I added a step that would run go mod tidy and error if there is a difference. I'm assuming tidy might modify both go.mod and go.sum.

Thanks! We can try it out and roll with the pain if found. :D

Issue #43, opened 2019-03-05
No way to stop the client.


n3wscott:
Consider adding `Stop()` to the client interface.

Issue #33, opened 2019-02-28
There is almost no client unit test coverage. 


n3wscott:
Fix it. Mock the http server and do round trip like tests for the http client. Similar for nats.

n3wscott:
There is some now! It could get better.

PR #45, opened 2019-03-06
Add client.StopReceiver

n3wscott:
## Changes

* StartReceiver now returns a context. This is intended to be an fyi for the caller of the client. to extract data around how the transport was setup. If the client knows it is using an http transport, then it can fetch out the port or path that is being used and advertise it. This is demoed in the unit test for client.
* client.WithPort can now accept 0 as a valid port and the transport will pick a valid open port for you.
* The `transport.Sender` interface was modified to support Stop: `StartReceiver(context.Context) (context.Context, error)` and `StopReceiver(context.Context) error` are added. The logic on starting and stopping a transport is now inside the transport and removed from the client object code.

Fixes https://github.com/cloudevents/sdk-go/issues/43, https://github.com/cloudevents/sdk-go/issues/33

## client.Client API Changes

Modified Method:

```go
StartReceiver(ctx context.Context, fn Receiver) error
```

becomes

```go
StartReceiver(ctx context.Context, fn Receiver) (context.Context, error)
```

New Method: 

```go
StopReceiver(ctx context.Context) error
```

### DIFF
--- a/pkg/cloudevents/client/options.go
+++ b/pkg/cloudevents/client/options.go
@@ -112,7 +112,7 @@ func WithHTTPStructuredEncoding() Option {
 func WithHTTPPort(port int) Option {
 	return func(c *ceClient) error {
 		if t, ok := c.transport.(*http.Transport); ok {
-			if port == 0 {
+			if port < 0 {
 				return fmt.Errorf("client option was given an invalid port: %d", port)
 			}
 			t.Port = port
--- a/pkg/cloudevents/client/options.go
+++ b/pkg/cloudevents/client/options.go
@@ -115,7 +115,7 @@ func WithHTTPPort(port int) Option {
 			if port < 0 {
 				return fmt.Errorf("client option was given an invalid port: %d", port)
 			}
-			t.Port = port
+			t.Port = &port
 			return nil
 		}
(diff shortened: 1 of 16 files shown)


n3wscott:
Code Coverage is now `82.4%`

n3wscott:
FYI: @Harwayne

grantr:
>  If the client knows it is using an http transport, then it can fetch out the port or path that is being used and advertise it.

I'm not familar with how the client is used, so sorry if this is an ignorant question, but: how does the client know it's using an http transport? How does the client "decide" which transport is used?

n3wscott:
Per @grantr 's comment, I am going to make a follow-up PR to decouple the client from the transport per https://github.com/cloudevents/sdk-go/issues/46

pkg/cloudevents/client/options.go
```diff
@@ -112,10 +112,10 @@ func WithHTTPStructuredEncoding() Option {
 func WithHTTPPort(port int) Option {
 	return func(c *ceClient) error {
 		if t, ok := c.transport.(*http.Transport); ok {
-			if port == 0 {
+			if port < 0 {
```
markpeek: Should there be a uint16 check here for min and max?

PR #64, opened 2019-03-07
Removing todo list, moved to issues.

n3wscott:
Also, the link to cloudevents sdk for serverless gateway was not a sdk, removed that too. 


markpeek:
Awesome. Thanks for taking care of that.

PR #68, opened 2019-03-11
Adding unit tests where tests are lacking.

n3wscott:
This brings unit test coverage to `80.4%`

Fixes: https://github.com/cloudevents/sdk-go/issues/35

Signed-off-by: Scott Nichols <nicholss@google.com>

PR #69, opened 2019-03-11
Reworking client to be better future proof.

n3wscott:
Major Changes:

[x] Removed `client/[http, nats]` helper classes. It forced the loss of type safety for options and gives no hints on what the new method accepts. Better to be explicit. 
[x] Adds `client.NewDefault()`, which takes no options. This can be used for the standard cases of sending and receiving for normal HTTP situations.
[x] `client.StartReceiver` is now a blocking call.
[x] Add mutex around StartReceiver.


n3wscott:
https://github.com/cloudevents/sdk-go/pull/68 should go in first, there will be come conflicts.

pkg/cloudevents/transport/http/transport.go
```diff
@@ -171,26 +176,28 @@ func (t *Transport) StartReceiver(ctx context.Context) error {
 		}
 		t.realPort = listener.Addr().(*net.TCPAddr).Port
 
-		go func() {
-			log.Print(t.server.Serve(listener))
-			t.server = nil
+		// Shutdown
+		defer func() {
 			t.realPort = 0
+			t.server.Close()
```
grantr: Might be better to use graceful `Shutdown(ctx)` here, but I'm not sure how to choose a shutdown timeout.

pkg/cloudevents/transport/http/transport.go
```diff
@@ -171,26 +176,28 @@ func (t *Transport) StartReceiver(ctx context.Context) error {
...
+		select {
+		case <-ctx.Done():
+			return nil
+		case err := <-errChan:
+			return err
 		}
+
 	}
-	return nil
+	return fmt.Errorf("http server already started")
```
grantr: Now that there's a mutex around the entire (blocking) method, can this ever be reached?

n3wscott: nope!

cmd/samples/http/receiver/main.go
```diff
@@ -41,7 +40,7 @@ func gotEvent(ctx context.Context, event cloudevents.Event) error {
 	}
 	fmt.Printf("Got Data: %+v\n", data)
 
-	fmt.Printf("Got Transport Context: %+v\n", http.TransportContextFrom(ctx))
+	fmt.Printf("Got Transport Context: %+v\n", cloudeventshttp.TransportContextFrom(ctx))
```
markpeek: other samples are using cehttp...do we want consistency in this naming?

n3wscott: I have not made a choice on which is best.

pkg/cloudevents/transport/http/transport.go
```diff
@@ -154,50 +154,47 @@ func (t *Transport) StartReceiver(ctx context.Context) error {
...
+	errChan := make(chan error, 1)
+	go func() {
+		errChan <- t.server.Serve(listener)
+	}()
+
+	// wait for the server to return or ctx.Done().
+	select {
+	case <-ctx.Done():
+		// Try a gracefully shutdown.
+		return t.server.Shutdown(context.Background())
```
grantr: When called with the background context, `Shutdown()` can block indefinitely. Is that what you want here?

PR #77, opened 2019-03-19
Pass headers from transport request setup to out-bound request.

n3wscott:
This change adds the option for `httptransport.WithHeader(key, value)`. This will add a header to the outbound http request for custom http headers. 

Also added full pretty printing to help debug this for event and http TransportContext.

Supersedes https://github.com/cloudevents/sdk-go/pull/73

Thanks @matzew for pointing out the `req.Close` option, and for testing the client out!!

### DIFF
--- a/pkg/cloudevents/transport/http/options.go
+++ b/pkg/cloudevents/transport/http/options.go
@@ -56,6 +56,28 @@ func WithMethod(method string) Option {
 	}
 }
 
+// WithHeader sets an additional default outbound header for all cloudevents
+// when using an HTTP request.
+func WithHeader(key, value string) Option {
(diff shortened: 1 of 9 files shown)


pkg/cloudevents/transport/http/options.go
```diff
@@ -56,6 +56,28 @@ func WithMethod(method string) Option {
 	}
 }
 
+// WithHeader sets an additional default outbound header for all cloudevents
+// when using an HTTP request.
+func WithHeader(key, value string) Option {
```
n3wscott: WithHeader makes it possible to setup a header to be sent for the life of the client but not per Send. Add WithHeader to context as well.

PR #83, opened 2019-03-20
Adding Getter/Setter for Extension attributes

nachocano:
**Proposed Changes**

* Adding an ExtensionAs(string, interface{}) function as part of EventContext interface to easily read extension attributes. Only supports decoding string extensions. Might need to use something like DataAs decoding functionality to avoid duplicating code, and better support multiple extension types.
* Adding a SetExtension(string, interface{}) function to every EventContextV* struct. As it mutates the underlying map, we need a pointer receiver, thus I couldn't add it as part of the EventContext interface.


n3wscott:
LGTM

Thanks for adding that. I think it is a find API addition and gives the hooks for something more complicated later when we go back to support extensions.

And thanks for adding tests. 🏅

pkg/cloudevents/event.go
```diff
@@ -27,6 +27,10 @@ func (e Event) Type() string {
 	return e.Context.GetType()
 }
 
+func (e Event) ExtensionAs(name string, obj interface{}) {
```
markpeek: Please add doc strings to this and the other public methods in this PR.

nachocano: Done! Thanks!

nachocano:
> LGTM. Given the work @n3wscott has been doing on adding doc strings, could you add doc strings to the public methods in this change? I only made this comment on one of the occurrences Thanks!.

Sounds good! Just added them.

nachocano:
@n3wscott @markpeek I hate this signing stuff :)
I messed up, let me close this PR and create a new clean one with proper signatures.

PR #85, opened 2019-03-20
Adding headers to send+response.

n3wscott:
Fixes https://github.com/cloudevents/sdk-go/issues/61

### DIFF
--- a/cmd/samples/http/sender/main.go
+++ b/cmd/samples/http/sender/main.go
@@ -84,12 +84,13 @@ func _main(args []string, env envConfig) int {
 					},
 				}
 
-				if event, err := c.Send(context.Background(), event); err != nil {
+				if resp, err := c.Send(context.Background(), event); err != nil {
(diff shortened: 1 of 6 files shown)


cmd/samples/http/sender/main.go
```diff
@@ -84,12 +84,13 @@ func _main(args []string, env envConfig) int {
 					},
 				}
 
-				if event, err := c.Send(context.Background(), event); err != nil {
+				if resp, err := c.Send(context.Background(), event); err != nil {
 					log.Printf("failed to send: %v", err)
-				} else if event != nil {
-					fmt.Printf("Got Event Response Context: %+v\n", event.Context)
+				} else if resp != nil {
+					fmt.Printf("Response:\n%s\n", resp)
```
n3wscott: The response from send is not passed back a context, and thus does not have access to any transport specific fields that might be valuable to them, such as http headers that are sent in response to a Send.

Consider solving them with a function similar to receive for response.

PR #89, opened 2019-03-21
Adding ExtensionAs and SetExtension methods

nachocano:
**Proposed Changes**

* Adding an ExtensionAs(string, interface{}) function as part of EventContext interface to easily read extension attributes. Only supports decoding string extensions. Might need to use something like DataAs decoding functionality to avoid duplicating code, and better support multiple extension types.
* Adding a SetExtension(string, interface{}) function to every EventContextV* struct. As it mutates the underlying map, we need a pointer receiver, thus I couldn't add it as part of the EventContext interface.

### DIFF
--- a/pkg/cloudevents/event.go
+++ b/pkg/cloudevents/event.go
@@ -31,6 +31,11 @@ func (e Event) Type() string {
 	return e.Context.GetType()
 }
 
+// ExtensionAs returns Context.ExtensionAs(name, obj)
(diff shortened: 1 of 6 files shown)


nachocano:
@n3wscott @matzew Here is a cleaner version. Sorry for the mess up.
Thanks!

markpeek:
Thanks!

PR #92, opened 2019-03-21
Qualify stat names to avoid collisions

grantr:
[Standard practice](https://godoc.org/go.opencensus.io/plugin/ocgrpc) for OpenCensus measure names is to qualify them with project names so the names are unique. Since measure names are never exported (only view names), this is a safe change.

Also the measures are now exported so library users can create custom views with them This is also standard practice.

/cc @n3wscott

### DIFF
--- a/pkg/cloudevents/client/observability.go
+++ b/pkg/cloudevents/client/observability.go
@@ -7,14 +7,16 @@ import (
 )
 
 var (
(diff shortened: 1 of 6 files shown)


n3wscott:
Thanks a bunch for looking at this.

n3wscott:
Looks good to me.

PR #94, opened 2019-03-27
found a bug in how header comes from context, fixed.

n3wscott:
I was passing headers by ref, not by value and header values were getting duplicated.


csantanapr:
I don't see a test being added or modified to cover this bug

PR #99, opened 2019-04-03
Adding testing to support work to do outbound quote support.

n3wscott:


n3wscott:
Spec changed. No one implemented double quoted headers.

PR #100, opened 2019-04-08
Adding larger integration tests for http, and a test dir for integration tests.

n3wscott:
This is https://github.com/cloudevents/sdk-go/pull/99 without the quote support work.

Found and fixed a few issues in the code from writing the integration tests. 🎉


n3wscott:
@markpeek I would like to do a release after this lands, `0.5.0`. The import changes are pretty nice and I think need a minor bump.

PR #105, opened 2019-04-09
Adding event interface

n3wscott:
With this change I am starting to move away from concrete for the consumer of the SDK. There has been requests to be able to setup the event without ever needing to know the impl details of the context versions.

Both of these work and are equivalent:

```go
event := cloudevents.Event{
	Context: cloudevents.EventContextV02{
		Type:   "com.cloudevents.sample.sent",
		Source: *source,
	}.AsV02(),
	Data: data,
}
```

```go
event := cloudevents.NewEvent()
event.SetType("com.cloudevents.sample.sent")
event.SetSource("https://github.com/cloudevents/sdk-go/cmd/samples/sender")
event.Data = data
```

### DIFF
--- a/alias.go
+++ b/alias.go
@@ -59,6 +59,13 @@ var (
 	WithUUIDs          = client.WithUUIDs
 	WithTimeNow        = client.WithTimeNow
 
+	// Event Creation
+
+	NewEvent   = cloudevents.New
+	VersionV01 = cloudevents.CloudEventsVersionV01
+	VersionV02 = cloudevents.CloudEventsVersionV02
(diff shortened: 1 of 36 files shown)


n3wscott:
This brings code coverage to `85%`

n3wscott:
I need to validate these changes more until I want a release. But I agree
it will be a 0.6 release. Version churn is not so big of a deal pre release.

-Scott

markpeek:
Except as you approximate getting to 1.0. As in v0.99999 :-)

PR #106, opened 2019-04-09
Adding batch support

n3wscott:
WORK IN PROGRESS


matzew:
@n3wscott  I think we can close this now, see: https://github.com/cloudevents/sdk-go/blob/ec2ebeba5274b8056192ea6d2a526e2a9b55ecc6/pkg/cloudevents/transport/http/encoding.go#L37

n3wscott:
Batched mode is defined as a setting but it throws a not implemented error at master.

PR #107, opened 2019-04-10
Stop spewing request data into log stream

iancoffey:
In a lot of places, `transport.go` prints the full body of the response. This seems ripe with the possibility for the wrong data to end up in the wrong log stream / log drain 👎

This removes many of them. Thanks!


n3wscott:
Thanks!

PR #109, opened 2019-04-11
Encoding cleanup, adding event.DataEncoded flag, adding event.SetData

n3wscott:
There was an issue developing where the data payload was un-trustable to the codecs. I needed a way to set the data and have it pass through the event data encoding, which is not the same as the transport encoding. So now there is a new method on event:

```go
event.SetData(mydata)
```

And this results in the event data becoming encoded on the event according to the event's encoding settings.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,46 @@
+# Development
+
+This is a collection of topics related to development of the SDK.
+
+## Transcoding
+
+One of the goals of this sdk is to decouple the encoding from the transport as
+much as possible. This is done by the integrating interacting with the
+[Event][cloudevents.event] object. The [Transport][transport.transport]
+implementation interacts with a [Message][transport.message]. The
+[Codec][transport.codec] is responsible for converting Event to Message for the
+the transport implementation. And this process works in reverse when a transport
+is used to receive an event.
+
+Sending:
+
+```
+ Event -[via Codec]-> Message -> Transport
+```
+
+Receiving:
+
+```
+(Transport) -> Message -[via Codec]-> Event
+```
+
+The CloudEvents spec outlines the various ways encoding is allowed, and there
+are two levels of encoding.
+
+1. Encoding of the context and extension attributes of the event.
+1. Encoding of the data payload of the event.
+
+For this reason there is also a [DataCodec][datacodec.codec] that is responsible
+for converting an encoded data payload into the intended format. These formats
+tend to be `application/xml`, `text/xml`, `application/json`, and `text/json`.
+
+The DataCodec is invoked (as of this writing) when converting Event to Message,
+or when `event.DataAs` is called when receiving.
+
+^ This should be rewritten so that it is invoked only on Data() and SetData()
+
+[cloudevents.event]: ../pkg/cloudevents/event.go
+[transport.transport]: ../pkg/cloudevents/transport/transport.go
+[transport.message]: ../pkg/cloudevents/transport/message.go
+[transport.codec]: ../pkg/cloudevents/transport/codec.go
+[datacodec.codec]: ../pkg/cloudevents/datacodec/codec.go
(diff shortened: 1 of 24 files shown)


test/http/loopback_setters_test.go
```diff
@@ -0,0 +1,466 @@
...
+	testCases := map[string]struct {
+		event  func(string) *cloudevents.Event
+		resp   func(string) *cloudevents.Event
+		want   map[string]*cloudevents.Event
+		asSent map[string]*TapValidation
+		asRecv map[string]*TapValidation
+	}{
+		"Loopback": {
+			event: func(version string) *cloudevents.Event {
+				event := cloudevents.NewEvent(version)
```
n3wscott: Check this out-->

I am creating events of all versions in a test loop.

docs/README.md
```diff
@@ -0,0 +1,44 @@
+# Development
+
+This is a collection of topics related to development of the SDK.
+
+## Transcoding
+
+One of the goals of this sdk is to decouple the encoding from the transport as
+much as possible. This is done by the integrating interacting with the
```
markpeek: "integrating interacting" reads odd. Perhaps reword?

pkg/cloudevents/codec/jsoncodec.go
```diff
@@ -87,26 +98,20 @@ func jsonEncode(ctx cloudevents.EventContextReader, data interface{}) ([]byte, e
...
-			b["data"] = []byte(strconv.QuoteToASCII(string(buf)))
-		} else if mediaType == "" || mediaType == cloudevents.ApplicationJSON {
-			b["data"] = datab
-		} else if datab[0] != byte('"') {
-			b["data"] = []byte(strconv.QuoteToASCII(string(datab)))
+		mediaType, err := ctx.GetDataMediaType()
+		if err != nil {
+			return nil, err
+		}
+		isBase64 := ctx.GetDataContentEncoding() == cloudevents.Base64
```
markpeek: Would this chunk of code make more sense to be its own function?

pkg/cloudevents/event_interface.go
```diff
@@ -66,4 +66,7 @@ type EventWriter interface {
 
 	// SetExtension performs event.Context.SetExtension.
 	SetExtension(string, interface{})
+
```
markpeek: AI: alphabetize the Set operations at some point.

pkg/cloudevents/transport/http/codec_v02.go
```diff
@@ -78,13 +78,18 @@ func (v CodecV02) encodeBinary(e cloudevents.Event) (transport.Message, error) {
 		return nil, err
 	}
 
-	mediaType, err := e.Context.GetDataMediaType()
+	//mediaType, err := e.Context.GetDataMediaType()
```
markpeek: Any reason to comment out rather than remove?

test/http/loopback_setters_test.go
markpeek: Nice!

docs/README.md
n3wscott: Thanks! yeah, I was trying to say sdk integrator

pkg/cloudevents/codec/jsoncodec.go
n3wscott: This logic is integral for the structured json encoding. I added a comment to help make this clear.

pkg/cloudevents/event_interface.go
n3wscott: I have been attempting to keep them in the order the spec introduces them: https://github.com/cloudevents/spec/blob/master/spec.md

pkg/cloudevents/transport/http/codec_v02.go
n3wscott: 🤕 this was a hard refactor tbh

n3wscott:
Ready for another look, thanks!

PR #110, opened 2019-04-12
Don't open and close HTTP connections too often

bbrowning:
This adds a test to verify we aren't opening and closing HTTP
connections too often when sending events to a single host. The test
will fail in CI (it fails locally) until I push a subsequent change to
actually fix this issue.

The test tracks the count of opened HTTP connections and fails if it's
an amount more than double the expected concurrency. In practice, when
this is fixed properly, the number of opened connections should be
close to if not exactly the expected concurrency. In the case of this
test, that would be 200 new connections. As the code stands today,
when I run this on my laptop, we actually opening over 3000 new
connections in just the 1 second this test runs.

Related to #98

### DIFF
--- a/pkg/cloudevents/transport/http/transport.go
+++ b/pkg/cloudevents/transport/http/transport.go
@@ -175,7 +175,6 @@ func (t *Transport) obsSend(ctx context.Context, event cloudevents.Event) (*clou
 
 		req.Body = ioutil.NopCloser(bytes.NewBuffer(m.Body))
 		req.ContentLength = int64(len(m.Body))
-		req.Close = true
 
 		return httpDo(ctx, t.Client, &req, func(resp *http.Response, err error) (*cloudevents.Event, error) {
 			logger := cecontext.LoggerFrom(ctx)
--- a/pkg/cloudevents/transport/http/transport.go
+++ b/pkg/cloudevents/transport/http/transport.go
@@ -48,7 +48,7 @@ type Transport struct {
 	// If nil, the Transport will create a one.
 	Client *http.Client
 	// Req is the base http request that is used for http.Do.
-	// Only .Method, .URL, and .Header is considered.
+	// Only .Method, .URL, .Close, and .Header is considered.
 	// If not set, Req.Method defaults to POST.
 	// Req.URL or context.WithTarget(url) are required for sending.
 	Req *http.Request
@@ -153,6 +153,7 @@ func (t *Transport) obsSend(ctx context.Context, event cloudevents.Event) (*clou
 	if t.Req != nil {
 		req.Method = t.Req.Method
 		req.URL = t.Req.URL
+		req.Close = t.Req.Close
 		copyHeaders(t.Req.Header, req.Header)
 	}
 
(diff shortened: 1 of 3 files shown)


bbrowning:
This blew up as expected in CI, confirming we have an issue with the current behavior:
```
--- FAIL: TestStableConnectionsToSingleHost (1.04s)
    transport_test.go:126: too many new connections opened: expected 200, got 3619
```

bbrowning:
By not forcing every HTTP connection to close every time we send an event AND by tuning the net/http client properly (see the `makeStableClient` func) we can get good connection reuse when sending concurrently to a single host.

cc @n3wscott @matzew

pkg/cloudevents/transport/http/transport.go
```diff
@@ -175,7 +175,6 @@ func (t *Transport) obsSend(ctx context.Context, event cloudevents.Event) (*clou
 
 		req.Body = ioutil.NopCloser(bytes.NewBuffer(m.Body))
 		req.ContentLength = int64(len(m.Body))
-		req.Close = true
```
n3wscott: This was added because of kafka per @matzew's request.

n3wscott: Ben, can you do one thing for me:

keep this deleted and change lines 153-156 to this:

```go
if t.Req != nil {
		req.Method = t.Req.Method
		req.URL = t.Req.URL
		req.Close = t.Req.Close
		copyHeaders(t.Req.Header, req.Header)
	}
```

bbrowning: Yes, moving that to be configurable by the user makes sense.

n3wscott:
Thanks a bunch. Now @matzew can still set that setting using the request injection and it is off by default.

pkg/cloudevents/transport/http/transport_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+// An example of how to make a stable client under sustained
+// concurrency sending to a single host
+func makeStableClient(addr string) (*cehttp.Transport, error) {
+	ceClient, err := cehttp.New(cehttp.WithTarget(addr))
+	if err != nil {
+		return nil, err
+	}
+	netHTTPTransport := &http.Transport{
+		MaxIdleConnsPerHost: 1000,
```
dubee: @bbrowning, how does `MaxIdleConnsPerHost ` with a value of `1000` interact with `MaxIdleConns` with its default value of `100`?

My understanding is that `MaxIdleConns` is the number of connections to keep-alive across all hosts and `MaxIdleConnsPerHost` is the number of connections to keep alive per host. I'm guessing that if `MaxIdleConnsPerHost` has a higher value than `MaxIdleConns`, only `MaxIdleConns` will be used??

bbrowning: @dubee I had to dig into the source to figure this out, but both values get applied. It first checks `MaxIdleConnsPerHost` and then enforces `MaxIdleConns` by evicting the oldest connections from the cache. There are slightly different semantics in the errors returned between the two limits as well as the actual behavior - hitting the global limit still puts the new connection in the cache and just evicts the oldest. Hitting the per-host limit doesn't put the new connection in the cache at all.

For the purposes of this test, that doesn't seem to actually matter, but it's good to know when tweaking these values on the Knative side of things.

PR #111, opened 2019-04-12
Change readme to use the top level package.

n3wscott:


n3wscott:
This is cleanup prep work for a pending release next week.

PR #114, opened 2019-04-15
Document GetDataMediaType

grantr:
This function swallows a parsing error, and that seems worth documenting.

/cc @n3wscott


n3wscott:
LGTM Thanks!!

PR #122, opened 2019-05-06
fix no data bug.

n3wscott:
Found the get bytes method throws an error if there are no bytes to decode. No bytes is a valid option, return back nothing, not an error.


n3wscott:
That's fair. Let me add one.

*From: *Mark Peek <notifications@github.com>
*Date: *Mon, May 6, 2019, 4:37 PM
*To: *cloudevents/sdk-go
*Cc: *Scott Nichols, Author

*@markpeek* requested changes on this pull request.

n3wscott:
Added tests.

PR #126, opened 2019-05-21
Adding missing fields for binary AMQP.

n3wscott:


deissnerk:
I haven't tested the code in my environment, but the changes should solve the problem.
LGTM

PR #127, opened 2019-05-29
Fix crash with nil value in binary AMQP message

deissnerk:
An extension attribute with value `nil` is now ignored.


n3wscott:
LGTM! Thanks!

PR #138, opened 2019-06-10
update import statement order

thinkerou:
run the command: `goimports -m .`

### DIFF
--- a/pkg/cloudevents/transport/nats/transport.go
+++ b/pkg/cloudevents/transport/nats/transport.go
@@ -7,6 +7,7 @@ import (
 	"github.com/cloudevents/sdk-go/pkg/cloudevents"
 	context2 "github.com/cloudevents/sdk-go/pkg/cloudevents/context"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport"
+	"github.com/nats-io/go-nats"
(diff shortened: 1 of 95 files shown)


pkg/cloudevents/transport/nats/transport.go
```diff
@@ -3,10 +3,10 @@ package nats
 import (
 	"context"
 	"fmt"
+
 	"github.com/cloudevents/sdk-go/pkg/cloudevents"
 	context2 "github.com/cloudevents/sdk-go/pkg/cloudevents/context"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport"
-	"github.com/nats-io/go-nats"
```
n3wscott: this was deleted and I think it is needed, this is why the CI is breaking.

n3wscott:
Thanks a ton for the cleanup, it looks like one import was dropped that is breaking the CI. Add that back in and we can be in business :D

thinkerou:
@n3wscott fixed! thanks!

n3wscott:
LGTM, nice work, I use goland and never really see the imports unless I expand them.

thinkerou:
@n3wscott maybe you should configure import set. because I use vim and vscode.

Issue #131, opened 2019-05-30
Need a way to tell when a Transport is ready


alanconway:
I can't see any way to tell when a transport is "ready", which is critical if you need to query the transport state. For example in transport.http there is this comment:

// GetPort returns the port the transport is active on.
// .Port can be set to 0, which means the transport selects a port, GetPort
// allows the transport to report back the selected port.

That is actually not possible: while one goroutine is blocked in StartReceiver(), there's no way for another goroutine to tell when it can call GetPort() - it has to happen after the net.Listen() call.

Proposed solution (I'll write a PR if this seems acceptable), 2 versions the clean and compatible:

// clean but incompatible, adding a method to Transport interface invalidates existing transports:
interface Transport {
  ...
  // Ready returns a channel that closes when StartReceiver() has completed any 
  // set up (listening, connecting etc.) and is waiting to receive events. It always
  //  closes before StartReceiver() returns, including in the event of an error.
  Ready() <-chan struct{}
}

// backwards compatible version, separate interface. Transports are strongly encouraged to 
// implement it but existing transports won't be broken.
interface Readyer {
  Ready() <-chan struct{}
}

matzew:
> backwards compatible version, separate interface

seems like a safe harbor

alanconway:
I have a rethink on the solution to this: "Ready" is a very transport-specific concept, some transports might not even have such a phase. So instead of adding this to the Tranport interface, I propose to update the HTTP transport (and any others with similar issues) as follows:

l := net.Listen(...)
transport := http.New(WithListener(l))

That gives the user full control (if they wish) over the listener, including the ability to get a Listen(":0") address from l.Addr().  It also allows listening on specific interfaces on a multi-homed host (which is currently impossible) and listening specifically for tpc6 or tcp4 trafic. And it is backwards compatible :)

We'd want to encourage something similar on any transport that uses a listener, I'll have a look at the existing ones.

n3wscott:
you can already give port 0 in http options. 

Let me look more at this...

alanconway:
On Thu, Jun 6, 2019 at 4:50 PM Scott Nichols <notifications@github.com>
wrote:

> you can already give port 0 in http options.
>
You can, and there is a GetPort() method - the problem is there is no pontt
at which you can call that method.
```
go func() { x.StartReceiver(...) }
port := x.GoPort() // WRONG:  the concurrent StartReceiver() may not have
passed net.Listen() yet.
[...]

PR #139, opened 2019-06-12
Add WithListener to http transport, ensure GetPort() always returns active port.

alanconway:
Fix for #131 

GetPort() calls net.Listen() on demand so it will always return the
actual listening port if port 0 was requested.

WithListener() allows the user to provide their own net.Listener, allows:
- alternate networks (e.g. "tcp4" or "tcp6" to restrict IP versions)
- listening interface to be specified for multi-homed hosts
- non-TCP listeners - GetPort() won't work but everything else will.

Also fixed a cancellation race in StartReceiver that showed up in tests.

### DIFF
--- a/pkg/cloudevents/transport/http/options.go
+++ b/pkg/cloudevents/transport/http/options.go
@@ -2,6 +2,7 @@ package http
 
 import (
 	"fmt"
+	"net"
 	nethttp "net/http"
 	"net/url"
 	"strings"
@@ -145,7 +146,18 @@ func WithStructuredEncoding() Option {
 	}
 }
 
-// WithPort sets the port for for clients with HTTP transports.
+func checkListen(t *Transport, prefix string) error {
+	switch {
+	case t.Port != nil:
+		return fmt.Errorf("%v port already set", prefix)
+	case t.listener != nil:
+		return fmt.Errorf("%v listener already set", prefix)
+	}
+	return nil
+}
+
+// WithPort sets the listening port for StartReceiver.
+// Only one of WithListener  or WithPort is allowed.
 func WithPort(port int) Option {
 	return func(t *Transport) error {
 		if t == nil {
@@ -154,8 +166,27 @@ func WithPort(port int) Option {
 		if port < 0 {
 			return fmt.Errorf("http port option was given an invalid port: %d", port)
 		}
-		t.Port = &port
+		if err := checkListen(t, "http port option"); err != nil {
+			return err
+		}
+		t.setPort(port)
 		return nil
+	}
+}
+
+// WithListener sets the listener for StartReceiver.
+// Only one of WithListener or WithPort is allowed.
+func WithListener(l net.Listener) Option {
+	return func(t *Transport) error {
+		if t == nil {
+			return fmt.Errorf("http listener option can not set nil transport")
+		}
+		if err := checkListen(t, "http port option"); err != nil {
+			return err
+		}
+		t.listener = l
+		_, err := t.listen()
+		return err
 	}
 }
 
--- a/pkg/cloudevents/transport/http/options_test.go
+++ b/pkg/cloudevents/transport/http/options_test.go
@@ -1,7 +1,9 @@
 package http
 
 import (
+	"context"
 	"fmt"
+	"net"
 	"net/http"
 	"net/url"
 	"testing"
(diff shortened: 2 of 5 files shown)


pkg/cloudevents/transport/http/options_test.go
```diff
@@ -351,6 +354,55 @@ func TestWithPort(t *testing.T) {
...
 }
 
+// Force a transport to close its server/listener by cancelling StartReceiver
+func forceClose(tr *Transport) {
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() { tr.StartReceiver(ctx) }()
+	cancel()
+}
+
+func TestWithPort0(t *testing.T) {
```
n3wscott: nice tests.

pkg/cloudevents/transport/http/options_test.go
```diff
@@ -351,6 +354,55 @@ func TestWithPort(t *testing.T) {
...
+			}
+			if d := cmp.Diff(port, *tr.Port); d != "" {
+				t.Error(d)
+			}
+		})
+	}
+}
+
+func TestWithListener(t *testing.T) {
+	l, err := net.Listen("tcp", ":0")
```
n3wscott: how is this different than `SetPort(0)` ?

pkg/cloudevents/transport/http/options.go
```diff
@@ -154,11 +166,29 @@ func WithPort(port int) Option {
 		if port < 0 {
 			return fmt.Errorf("http port option was given an invalid port: %d", port)
 		}
-		t.Port = &port
+		if err := t.checkListen("http port option"); err != nil {
```
n3wscott: I think it would be safer to pass t to checkListen.

pkg/cloudevents/transport/http/options.go
```diff
@@ -145,7 +146,18 @@ func WithStructuredEncoding() Option {
 	}
 }
 
-// WithPort sets the port for for clients with HTTP transports.
+func (t *Transport) checkListen(prefix string) error {
```
n3wscott: I would expect this method to be:

```golang
func checkListen(t *Transport, prefix string) error
```

The reason these option functions exist in this location is to gain access to private members of the http transport struct.

pkg/cloudevents/transport/http/transport.go
```diff
@@ -296,28 +296,25 @@ func (t *Transport) StartReceiver(ctx context.Context) error {
 		t.handlerRegistered = true
 	}
 
-	addr := fmt.Sprintf(":%d", t.GetPort())
+	if err := t.listen(); err != nil {
```
n3wscott: can you make listen return the addr?

pkg/cloudevents/transport/http/transport.go
```diff
@@ -296,28 +296,25 @@ func (t *Transport) StartReceiver(ctx context.Context) error {
...
+		return err
+	}
+	addr := t.listener.Addr().String()
```
n3wscott: then this line goes away.

pkg/cloudevents/transport/http/transport.go
```diff
@@ -599,18 +598,39 @@ func (t *Transport) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+func (t *Transport) setPort(port int) {
+	if t.Port == nil {
+		t.Port = new(int)
 	}
-	return 8080 // default
+	*t.Port = port
+}
+
+// listen if not already listening, update t.Port
+func (t *Transport) listen() (err error) {
```
n3wscott: this method should be:

```golang
func (t *Transport) listen() (string, error)
```

returning the addr and the error.

pkg/cloudevents/transport/http/options_test.go
alanconway: I can say net.Listen("tcp6", "myhost:0") to listen for IPv6 connections only, and only on the interface identfied by myhost. This is a typical pattern for packages that listen: there may be a short cut  (e.g. net.http.ListenAndServe() but it's usually possible for the caller to provide the Listener for flexibility  (e.g. net.http.Serve(Listener))

pkg/cloudevents/transport/http/options.go
alanconway: There are no private members in Go, only exported and not exported. Nothing is private within a package.
I've no objection to changing it as a matter of style.

pkg/cloudevents/transport/http/transport.go
alanconway: will do

pkg/cloudevents/transport/http/transport.go
alanconway: done

pkg/cloudevents/transport/http/options.go
alanconway: done

n3wscott:
oh no! merge conflict. 

LGTM, will merge after the conflict is fixed.

alanconway:
Conflict fixed.

n3wscott:
LGTM!

Thanks!

Issue #140, opened 2019-06-13
Use of cloudevent HTTP transport as codec only


alanconway:
There are 2 use cases for cloudevents transport:

1. As message codec only: to translate Event to/from the native transport libraries message type (e.g. for HTTP this would be http.Request and http.Response) This is for applications that want full access to features of the underlying transport library (e.g. net/http) and decisions about dispatching, use of goroutines, memory management etc.

2. As a generic API for transport-neutral applications that want to allow replaceable transports. This is for applications that care most about isolating themselves from transport details. It is ok (when necessary) to limit access to the underlying transport and make implementation choices - the goal is encapsulation rather than total freedom.

The HTTP code provides 2. and is close to 1.  it is missing some convenience methods, e.g.

    func (m *Message) toRequest(r *http.Request)

This issue serves for HTTP, other issues referencing this can be raised for other transports.

n3wscott:
I am open to this idea. Some want this feature for sure. I get questions like "I have a req, can i haz cloudevent?"

alanconway:
What are your thoughts on the PR? I'd like to agree on a good API approach on HTTP before I run around and do something similar for AMQP or other transports. It is a zero-cost feature since the transport has to have this code internally anyway, and exporting it may even improve the structure of transport impls.

pgarneau:
We're developing a Kafka implementation internally and we want to use the CloudEvents format, but currently can't because there's no easy way to simply translate (bytes, string, etc) into a an Event object and manipulate it. We don't want to use the HTTP, Pubsub, etc receivers since we have our own for Kafka

n3wscott:
@pgarneau you can use the codecs directly if you choose.

n3wscott:
@alanconway I commented on #141, I think it is the correct direction with some tweaks for golang and to match the style of the project.

pgarneau:
@n3wscott Thanks! I'll give it a try and see where it takes me.

n3wscott:
@pgarneau if you want to add kafka transport to this sdk, that would be really awesome.

slinkydeveloper:
Done in sdk v2

PR #141, opened 2019-06-13
Issue #140: Add Message.ToRequest()/FromRequest

alanconway:
For issue #140 

Related changes/fixes:

Re-use Message as part of Response so a response can also be treated
as a Message.

Added/used copyHeadersEnsure() to ensure headers are not dropped on the
floor if copied into a nil Header map in a zero-initialized
http.Request or Message.

### DIFF
--- a/pkg/cloudevents/transport/http/message.go
+++ b/pkg/cloudevents/transport/http/message.go
@@ -23,7 +23,7 @@ type Message struct {
 // Response is an http transport response.
 type Response struct {
 	StatusCode int
-	Message    Message
+	Message
 }
 
 // CloudEventsVersion inspects a message and tries to discover and return the
@@ -90,23 +90,54 @@ func readAllClose(r io.ReadCloser) ([]byte, error) {
 	return nil, nil
 }
 
-// FromRequest initializes a Message from a http.Request
-func (m *Message) FromRequest(r *http.Request) (err error) {
-	copyHeadersEnsure(r.Header, &m.Header)
-	m.Body, err = readAllClose(r.Body)
+// NewMessage creates a new message from the Header and Body of
+// an http.Request or http.Response
+func NewMessage(header http.Header, body io.ReadCloser) (*Message, error) {
+	var m Message
+	err := m.Init(header, body)
+	return &m, err
+}
+
+// NewResponse creates a new response from the Header and Body of
+// an http.Request or http.Response
+func NewResponse(header http.Header, body io.ReadCloser, statusCode int) (*Response, error) {
+	resp := Response{StatusCode: statusCode}
+	err := resp.Init(header, body)
+	return &resp, err
+}
+
+// Copy copies a new Body and Header into a message, replacing any previous data.
+func (m *Message) Init(header http.Header, body io.ReadCloser) error {
+	m.Header = make(http.Header, len(header))
+	copyHeadersEnsure(header, &m.Header)
+	var err error
+	m.Body, err = readAllClose(body)
 	return err
 }
 
-// ToRequest copies a message to a http.Request.
-// Replaces r.Body, r.ContentLength and r.Method.
-// Updates r.Headers.
-func (m *Message) ToRequest(r *http.Request) {
-	copyHeadersEnsure(m.Header, &r.Header)
+func (m *Message) copyOut(header *http.Header, body *io.ReadCloser) {
+	copyHeadersEnsure(m.Header, header)
+	*body = nil
 	if m.Body != nil {
-		r.Body = ioutil.NopCloser(bytes.NewBuffer(m.Body))
-	} else {
-		r.Body = nil
+		copy := append([]byte(nil), m.Body...)
+		*body = ioutil.NopCloser(bytes.NewBuffer(copy))
 	}
-	r.ContentLength = int64(len(m.Body))
-	r.Method = http.MethodPost
+}
+
+// ToRequest updates a http.Request from a Message.
+// Replaces Body, ContentLength and Method, updates Headers.
+// Panic if req is nil
+func (m *Message) ToRequest(req *http.Request) {
+	m.copyOut(&req.Header, &req.Body)
+	req.ContentLength = int64(len(m.Body))
+	req.Method = http.MethodPost
+}
+
+// ToResponse updates a http.Response from a Response.
+// Replaces Body, updates Headers.
+// Panic if resp is nil
+func (m *Response) ToResponse(resp *http.Response) {
+	m.copyOut(&resp.Header, &resp.Body)
+	resp.ContentLength = int64(len(m.Body))
+	resp.StatusCode = m.StatusCode
 }
--- a/pkg/cloudevents/transport/http/message.go
+++ b/pkg/cloudevents/transport/http/message.go
@@ -1,7 +1,11 @@
 package http
 
 import (
+	"bytes"
 	"encoding/json"
+
+	"io"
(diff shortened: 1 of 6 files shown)


pkg/cloudevents/transport/http/message.go
```diff
@@ -77,3 +81,32 @@ func (m Message) CloudEventsVersion() string {
...
+func readAllClose(r io.ReadCloser) ([]byte, error) {
+	if r != nil {
+		defer r.Close()
+		return ioutil.ReadAll(r)
+	}
+	return nil, nil
+}
+
+// FromRequest initializes a Message from a http.Request
+func (m *Message) FromRequest(r *http.Request) (err error) {
```
n3wscott: I think this method signature should be:

```golang
func FromRequest(req *http.Request)  (*Message, error)
```

pkg/cloudevents/transport/http/message.go
```diff
@@ -77,3 +81,32 @@ func (m Message) CloudEventsVersion() string {
...
+	copyHeadersEnsure(r.Header, &m.Header)
+	m.Body, err = readAllClose(r.Body)
+	return err
+}
+
+// ToRequest copies a message to a http.Request.
+// Replaces r.Body, r.ContentLength and r.Method.
+// Updates r.Headers.
+func (m *Message) ToRequest(r *http.Request) {
+	copyHeadersEnsure(m.Header, &r.Header)
```
n3wscott: r could be nil

pkg/cloudevents/transport/http/message.go
```diff
@@ -77,3 +81,32 @@ func (m Message) CloudEventsVersion() string {
...
+func (m *Message) FromRequest(r *http.Request) (err error) {
+	copyHeadersEnsure(r.Header, &m.Header)
+	m.Body, err = readAllClose(r.Body)
+	return err
+}
+
+// ToRequest copies a message to a http.Request.
+// Replaces r.Body, r.ContentLength and r.Method.
+// Updates r.Headers.
+func (m *Message) ToRequest(r *http.Request) {
```
n3wscott: `r`  -> `req`

pkg/cloudevents/transport/http/message.go
n3wscott: `r` -> `req`

n3wscott: don't name the return

pkg/cloudevents/transport/http/transport.go
```diff
@@ -146,6 +145,16 @@ func copyHeaders(from, to http.Header) {
 	}
 }
 
+// Ensure to is a non-nil map before copying
+func copyHeadersEnsure(from http.Header, to *http.Header) {
```
n3wscott: `http.Header` is a `[][]string` and by definition it is already a pointer. No need to make it a pointer pointer.

pkg/cloudevents/transport/http/transport.go
```diff
@@ -575,12 +575,12 @@ func (t *Transport) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 		}
-		if len(resp.Header) > 0 {
-			copyHeaders(resp.Header, w.Header())
+		if len(resp.Message.Header) > 0 {
+			copyHeaders(resp.Message.Header, w.Header())
 		}
-		w.Header().Add("Content-Length", strconv.Itoa(len(resp.Body)))
-		if len(resp.Body) > 0 {
-			if _, err := w.Write(resp.Body); err != nil {
+		w.Header().Add("Content-Length", strconv.Itoa(len(resp.Message.Body)))
```
n3wscott: the comment on FromRequest implies this happens inside that function.

pkg/cloudevents/transport/http/message.go
```diff
@@ -18,8 +23,7 @@ type Message struct {
 // Response is an http transport response.
 type Response struct {
 	StatusCode int
-	Header     http.Header
-	Body       []byte
+	Message    Message
```
n3wscott: embed message:

```golang
type Response struct {
	StatusCode int
	Message
}
```

pkg/cloudevents/transport/http/message.go
n3wscott: or name this `MergeRequest`

pkg/cloudevents/transport/http/message.go
alanconway: done

pkg/cloudevents/transport/http/message.go
alanconway: Why? That has semantic meaning in Go and is necessary if you want to manipulate return values in a defer() statement. It can also be helpful as documentation if there are multiple same-type return values. 

None of that applies here, I'll change it (just means extra var statements in the body) but why the rule?

pkg/cloudevents/transport/http/message.go
alanconway: +1

pkg/cloudevents/transport/http/transport.go
alanconway: In this case there is - it's the classic C "a writable pointer parameter is a pointer to a pointer". The zero value of a map is nil, *not* an empty map. So if you have
    var r req http.Request
    copyHeadersEnsure(m, r.Header)
you will be passing a *nil map* to copyHeadersEnsure, not an existing map that can be updated. The only way a function like copyHeadersEnsure can change the value of r.Header to be a *different map*  (not just modify an existing non-nil map) is by passing a pointer. The nil map is different from any useable map.

I find it pretty annoying that the zero value of a map in Go is not a usable map, but I can kinda see why they did it from a performance perspective. The same holds for channels and other "pointer like" types - they are pointer like including in the sense that they can be nil, and you have to program for that.

pkg/cloudevents/transport/http/message.go
alanconway: I prefer to initialize in place because it gives the user more control over allocations. In a high-performance event bridge, it means the difference between a pool of re-usable Message objects and an allocation for every message, right on the critical path.  I've added a NewMessage() constructor with the signature you requested, but I'd rather keep in-place initialization.

pkg/cloudevents/transport/http/message.go
alanconway: That would panic. I should state that in the docs, but using a pointer as a way to provide "write access" to a parameter that is not allowed to be nil is a pretty common idiom, since Go has no other way to do that.

On a more general note, I see a lot of func (x *X) { if x == nil...} in the CE code. There are rare cases where it makes sense to operate on a nil pointer but they are... rare. IMO in most cases, calling a method on a nil pointer is a logic error and should result in an immediate panic, not get buried in a log file (if you're lucky) as a run-time error while the program to continues rolling towards certain doom. It's a debatable point, I like a good debate :)

alanconway:
Made some changes and renamed some things, @n3wscott let me know what you think.

pkg/cloudevents/transport/http/message_test.go
```diff
@@ -0,0 +1,109 @@
+/*
+Copyright 2019 The Knative Authors
```
n3wscott: Knative?

n3wscott:
Ping on the knative question

alanconway:
> Ping on the knative question

Doh, sorry - been working on  knative also wrong copyright. Will remove.

n3wscott:
LGTM

Thanks!

PR #142, opened 2019-06-13
Add ReceiveFunc wrapper - make Receiver from function.

alanconway:
This is analogous to http.HandleFunc which makes a http.Handler
from a function.

### DIFF
--- a/pkg/cloudevents/transport/transport.go
+++ b/pkg/cloudevents/transport/transport.go
@@ -2,6 +2,7 @@ package transport
 
 import (
 	"context"
+
 	"github.com/cloudevents/sdk-go/pkg/cloudevents"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/cloudevents/transport/transport.go
```diff
@@ -26,6 +27,14 @@ type Receiver interface {
 	Receive(context.Context, cloudevents.Event, *cloudevents.EventResponse) error
 }
 
+// ReceiveFunc wraps a function as a Receiver object.
+type ReceiveFunc func(ctx context.Context, e cloudevents.Event, er *cloudevents.EventResponse) error
+
+// Receive implements Receiver.Receive
+func (f ReceiveFunc) Receive(ctx context.Context, e cloudevents.Event, er *cloudevents.EventResponse) error {
```
n3wscott: `e` -> `event`
`er` -> `resp`

n3wscott: how does this work?

Can you add a test to show this working in practice?

n3wscott:
ok I get it now. this let's you use a function in place of a struct that implements that method because you want to use the Receive transport callback with just a function.

n3wscott:
LGTM

PR #146, opened 2019-06-15
migrate away from common json codec and use marshal methods.

n3wscott:
There use to be a codec that was shared for all transports to convert an event to json. This PR moves that codec and puts it behind the `event.(Un)MarshalJSON` method for golang, in an effort to stage some of the work to use codecs independently of the transports.


n3wscott:
cc @alanconway

alanconway:
+1 - it's less surprising and Event becomes a standard JSON(Un)Marshaller so it can be marshalled as part of a struct, map, to a json.Encoder/Decoder etc.  It also means the user doesn't have to take special steps to guess/know the spec version in advance - just unmarshal and then call GetSpecVersion() as with any other attribute.

n3wscott:
@alanconway there is a gotchya. This only works on structured representations of the event, not binary mode. Binary mode has components of the event in several places and it specific to each transport.

Structured more is more widely shared for all transports so I felt it did make sense to hook these together.

The decoder/transport still needs to look at its message and inspect the encoding before it can understand which route to decode it should take.

PR #152, opened 2019-07-02
Fix issues found by  golangci-lint

alanconway:
golangcli-lint combines multiple linters, including go vet. It has
good default settings - not much noise, I recommend it. Fixes are to:

- Remove dead/unreachable code
- Fix missing error checks:
  I mostly marked them intentional with `_ = foo()`, I think this is right
  One was in in a `defer` statement I fixed by returning it.
- Minor simplifications (from 'gosimple' linter)

### DIFF
--- a/pkg/cloudevents/observability/observer.go
+++ b/pkg/cloudevents/observability/observer.go
@@ -27,12 +27,11 @@ type Reporter interface {
 }
 
 type reporter struct {
-	ctx     context.Context
-	span    *trace.Span
-	on      Observable
-	start   time.Time
-	measure stats.Measure
-	once    sync.Once
+	ctx   context.Context
+	span  *trace.Span
+	on    Observable
+	start time.Time
+	once  sync.Once
 }
 
 // All tags used for Latency measurements.
--- a/pkg/cloudevents/transport/http/transport_test.go
+++ b/pkg/cloudevents/transport/http/transport_test.go
@@ -31,7 +31,7 @@ func startTestServer(handler http.Handler) (*http.Server, error) {
 		Addr:    listener.Addr().String(),
 		Handler: handler,
 	}
-	go server.Serve(listener)
+	go func() { _ = server.Serve(listener) }()
 	return server, nil
 }
 
@@ -135,7 +135,6 @@ func TestStableConnectionsToSingleHost(t *testing.T) {
 func TestMiddleware(t *testing.T) {
 	testCases := map[string]struct {
 		middleware []string
-		want       string
 	}{
 		"none": {},
 		"one": {
@@ -194,7 +193,7 @@ func makeRequestToServer(t *testing.T, tr *cehttp.Transport, responseText string
 	// Start the server.
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	go tr.StartReceiver(ctx)
+	go func() { _ = tr.StartReceiver(ctx) }()
 
 	// Give some time for the receiver to start. One second was chosen arbitrarily.
 	time.Sleep(time.Second)
--- a/pkg/cloudevents/transport/nats/codec_test.go
+++ b/pkg/cloudevents/transport/nats/codec_test.go
@@ -14,10 +14,6 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
-func strptr(s string) *string {
-	return &s
-}
-
 func TestCodecEncode(t *testing.T) {
 	sourceUrl, _ := url.Parse("http://example.com/source")
 	source := &types.URLRef{URL: *sourceUrl}
--- a/pkg/cloudevents/transport/nats/transport.go
+++ b/pkg/cloudevents/transport/nats/transport.go
@@ -112,7 +112,7 @@ func (t *Transport) HasConverter() bool {
 
 // StartReceiver implements Transport.StartReceiver
 // NOTE: This is a blocking call.
-func (t *Transport) StartReceiver(ctx context.Context) error {
+func (t *Transport) StartReceiver(ctx context.Context) (err error) {
 	if t.Conn == nil {
 		return fmt.Errorf("no active nats connection")
 	}
@@ -130,7 +130,6 @@ func (t *Transport) StartReceiver(ctx context.Context) error {
 		return fmt.Errorf("subject required for nats listen")
 	}
 
-	var err error
 	// Simple Async Subscriber
 	t.sub, err = t.Conn.Subscribe(t.Subject, func(m *nats.Msg) {
 		logger := context2.LoggerFrom(ctx)
@@ -154,7 +153,10 @@ func (t *Transport) StartReceiver(ctx context.Context) error {
 	})
 	defer func() {
 		if t.sub != nil {
-			t.sub.Unsubscribe() // TODO: create an error channel to pass this up
+			err2 := t.sub.Unsubscribe()
+			if err != nil {
+				err = err2 // Set the returned error if not already set.
+			}
 			t.sub = nil
 		}
 	}()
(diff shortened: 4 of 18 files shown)


pkg/cloudevents/observability/observer.go
```diff
@@ -27,12 +27,11 @@ type Reporter interface {
 }
 
 type reporter struct {
-	ctx     context.Context
-	span    *trace.Span
-	on      Observable
-	start   time.Time
-	measure stats.Measure
```
n3wscott: `measure stats.Measure` - I am suppressed this is not used. I think we need the measurement.

pkg/cloudevents/transport/amqp/codec_test.go
```diff
@@ -370,13 +370,6 @@ func TestCodecRoundTrip(t *testing.T) {
 					got.Data = data
 				}
 
-				if tc.wantErr != nil || err != nil {
```
n3wscott: the tests need to be expanded to non-happy path only. Please leave this?

pkg/cloudevents/transport/amqp/codec_test.go
```diff
@@ -508,13 +501,6 @@ func TestCodecAsMiddleware(t *testing.T) {
...
```
n3wscott: the tests need to be expanded to non-happy path only. Please leave this?

pkg/cloudevents/transport/http/codec_test.go
```diff
@@ -797,13 +797,6 @@ func TestCodecRoundTrip(t *testing.T) {
 					got.Data = data
 				}
 
-				if tc.wantErr != nil || err != nil {
```
n3wscott: the tests need to be expanded to non-happy path only. Please leave this?

pkg/cloudevents/transport/http/codec_test.go
```diff
@@ -939,13 +932,6 @@ func TestCodecAsMiddleware(t *testing.T) {
 						got.Data = data
 					}
 
-					if tc.wantErr != nil || err != nil {
```
n3wscott: the tests need to be expanded to non-happy path only. Please leave this?

pkg/cloudevents/transport/http/transport_test.go
```diff
@@ -136,7 +136,7 @@ func TestStableConnectionsToSingleHost(t *testing.T) {
 func TestMiddleware(t *testing.T) {
 	testCases := map[string]struct {
 		middleware []string
-		want       string
```
n3wscott: hm that is a weird change.

pkg/cloudevents/transport/nats/codec_test.go
```diff
@@ -240,13 +236,6 @@ func TestCodecRoundTrip(t *testing.T) {
 					got.Data = data
 				}
 
-				if tc.wantErr != nil || err != nil {
```
n3wscott: the tests need to be expanded to non-happy path only. Please leave this?

pkg/cloudevents/transport/nats/codec_test.go
```diff
@@ -378,13 +367,6 @@ func TestCodecAsMiddleware(t *testing.T) {
...
```
n3wscott: the tests need to be expanded to non-happy path only. Please leave this?

pkg/cloudevents/transport/nats/transport.go
```diff
@@ -112,7 +112,7 @@ func (t *Transport) HasConverter() bool {
 
 // StartReceiver implements Transport.StartReceiver
 // NOTE: This is a blocking call.
-func (t *Transport) StartReceiver(ctx context.Context) error {
+func (t *Transport) StartReceiver(ctx context.Context) (err error) {
```
n3wscott: do not name the return params.

pkg/cloudevents/types/urlref.go
```diff
@@ -51,7 +51,7 @@ func (u *URLRef) UnmarshalJSON(b []byte) error {
 // MarshalXML implements a custom xml marshal method used when this type is
 // marshaled using xml.Marshal.
 func (u URLRef) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
-	v := fmt.Sprintf("%s", u.String())
+	v := u.String()
 	return e.EncodeElement(v, start)
```
n3wscott: perhaps `return e.EncodeElement(u.String(), start)`

test/http/loopback.go
```diff
@@ -110,44 +110,3 @@ func ClientLoopback(t *testing.T, tc TapTest, topts ...cehttp.Option) {
 		assertTappedEquality(t, "http response", tc.asRecv, &resp)
 	}
 }
-
-// To help with debug, if needed.
-func printTap(t *testing.T, tap *tapHandler, testID string) {
```
n3wscott: please leave this, it was really helpful for debugging.

test/http/tap_handler.go
```diff
@@ -87,7 +87,7 @@ func (t *tapHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 func copyHeaders(from http.Header) http.Header {
 	to := http.Header{}
-	if from == nil || to == nil {
```
n3wscott: this will cause a npe

pkg/cloudevents/observability/observer.go
n3wscott: ah, looks like it was a migration leftovers. 👍

pkg/cloudevents/transport/amqp/codec_test.go
alanconway: will do, although right now it is unreachable.

pkg/cloudevents/transport/http/codec_test.go
alanconway: will do, although right now it is unreachable

pkg/cloudevents/transport/http/transport_test.go
alanconway: Very! The want member is not used, I've no idea why I replaced it with "_" instead of just removing it. Removed.

pkg/cloudevents/transport/nats/codec_test.go
alanconway: ok

pkg/cloudevents/transport/nats/transport.go
alanconway: This is necessary to  fix your error return problem at the end of the function. You had a comment:
    // TODO: create an error channel to pass this up
A channel is not necessary. A deferred function executes within the scope of the function that declares it, and can set the return values - but only if they are named.

pkg/cloudevents/transport/nats/codec_test.go
alanconway: ok

pkg/cloudevents/types/urlref.go
alanconway: Yes, better.

test/http/loopback.go
alanconway: ok

alanconway:
Fixed all requests except named return value required to return an error from a deferred function call.
Rebased and added some more fixes picked up from master - I think they are all uncontroversial

pkg/cloudevents/transport/nats/transport.go
n3wscott: Is that right?! Well thanks for that info. Great to know. LGTM

n3wscott:
Looks like you need to run `go fmt` 

Otherwise, LGTM, thanks for the work!

pkg/cloudevents/transport/nats/transport.go
alanconway: Also after a bit of googling I can't find any objections to named returns, and some articles on how they can be helpful: https://golang.org/doc/effective_go.html#named-results

IMO the only danger is the "naked" return, so I don't use it. I always list the return values even if they are the same variables that were named in parameters to avoid shadowing confusion, e.g.

```
func New() (t *Thing, err error) {
  if err := something(); err != nil {
     return // BAD 
     // naked return uses the variables which are still  (nil, nil)
     // The `err :=`  declaration shadows the err result.

     return t, err // GOOD
     // Explicit return evaluated in local context using inner err variable.
     // values are copied to the result variables before doing defers
 }
 ...
}
```

alanconway:
Fixed formatting, squashed into a single commit.

n3wscott:
LGTM! THANKS!!!

PR #155, opened 2019-07-10
Adding context to codec methods.

n3wscott:
Future work requires that context be available in the methods related to codecs. 

This also adds tracing to these methods in the context of the call stack.


n3wscott:
No code change. Just adding context methods.

PR #157, opened 2019-07-10
Context based encoding.

n3wscott:
There was a request to allow for the reuse of the client but be able to select the outbound encoding of an event. At the moment, we set this internally and then always output binary or structured. But there are methods that allow the selection of an encoding based on the event. This has been used, for example, to select the versioned codec for the same as the outbound event. [Wiring context](https://github.com/cloudevents/sdk-go/pull/155) allows for an override to be injected into context and then informs the selection of the codec.

### DIFF
--- a/cmd/samples/http/sender/main.go
+++ b/cmd/samples/http/sender/main.go
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	count = 1000
+	count = 1
 )
 
 type envConfig struct {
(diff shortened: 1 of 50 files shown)


n3wscott:
Related to https://github.com/knative/eventing/pull/1523/

cmd/samples/http/sender/main.go
```diff
@@ -44,13 +44,16 @@ func _main(args []string, env envConfig) int {
 		return 1
 	}
 
+	ctx := cloudevents.ContextWithEncoding(context.Background(), cloudevents.Structured)
+
 	seq := 0
 	for _, contentType := range []string{"application/json", "application/xml"} {
 		for _, encoding := range []cloudevents.HTTPEncoding{cloudevents.HTTPBinaryV01, cloudevents.HTTPStructuredV01, cloudevents.HTTPBinaryV02, cloudevents.HTTPStructuredV02} {
```
sayanh: I think we should add `HTTPBinaryV03`, `HTTPStructuredV03`, `HTTPBatchedV03` in the array

deissnerk:
LGTM

n3wscott:
Thanks for the reviews!

PR #158, opened 2019-07-11
Added an example of a CEV03 sender using a custom client

sayanh:


cmd/samples/http/sender-with-custom-client/main.go
```diff
@@ -0,0 +1,139 @@
...
+
+func _main(args []string, env envConfig) int {
+	source, err := url.Parse("https://github.com/cloudevents/sdk-go/cmd/samples/sender")
+	if err != nil {
+		log.Printf("failed to parse source url, %v", err)
+		return 1
+	}
+
+	// Add path to client certificate and key for HTTPS
+	cert, err := tls.LoadX509KeyPair("generated.crt", "generated.key")
```
n3wscott: can these paths and files be passed in? How do I generate them?

n3wscott:
Looks good, I have a question on how to create "generated.crt" and "generated.key" and then perhaps those paths can be passed in as env arguments (and defaulted to what they are now.)

cmd/samples/http/sender-with-custom-client/main.go
sayanh: Sorry about the name `generated`. Actually, to send events to a source which is TLS enabled, these certs are needed and this can be a valid use-case to use a custom client. Surely, we could pass as ENV variables.

cmd/samples/http/sender-with-custom-client/main.go
```diff
@@ -0,0 +1,141 @@
...
+func _main(args []string, env envConfig) int {
+	source, err := url.Parse("https://github.com/cloudevents/sdk-go/cmd/samples/sender")
+	if err != nil {
+		log.Printf("failed to parse source url, %v", err)
+		return 1
+	}
+
+	// Read certificate and key information from the environment for HTTPS
+	clientCert := os.Getenv("CLIENT_CERT")
+	clientKey := os.Getenv("CLIENT_KEY")
```
n3wscott: Can you migrate this to use the envconfig, noted above.

cmd/samples/http/sender-with-custom-client/main.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/kelseyhightower/envconfig"
+)
+
+const (
+	count = 1000
+)
+
+type envConfig struct {
+	// Target URL where to send cloudevents
+	Target string `envconfig:"TARGET" default:"https://localhost:8080" required:"true"`
```
n3wscott: like here: this allows you to validate that an env var is correct and set, so you can mark yours as required for this to work.

Re: "CLIENT_KEY" and "CLIENT_CERT"

cmd/samples/http/sender-with-custom-client/main.go
sayanh: done

cmd/samples/http/sender-with-custom-client/main.go
sayanh: absolutely

n3wscott:
LGTM

THANKS!!!!

PR #159, opened 2019-07-11
Remove bogus entry from go.sum

alanconway:
Seeing weird behavior where every `go build` was adding a line for
package github.com/gogo/protobuf to go.mod. `go tidy` would remove it
as unused, but the next `go build` would put it back.

I removed all the github.com/gogo/protobuf lines from go.sum and ran
`go tidy`. It put all but one of the lines back. Now everything works.

Weird.


n3wscott:
WERID!!! yeah I was having that same issue, and I was assuming it was because the tests used some lib that the cmd did not.

Thanks for figuring it out

LGTM

PR #163, opened 2019-07-16
Changes to keep `go vet` happy.

alanconway:
Various minor syntactic changes, no change in behavior.

Removed TestUnmarshal_nilEvent(t *testing.T)
Calling json.Unmarshal with a nil argument is always an error as
per the json package doc. `go vet` even reports it statically.
There is no need to verify that here, it's not special to cloudevents.

### DIFF
--- a/pkg/cloudevents/event_marshal_test.go
+++ b/pkg/cloudevents/event_marshal_test.go
@@ -489,19 +489,6 @@ func TestUnmarshal(t *testing.T) {
 	}
 }
 
-func TestUnmarshal_nilEvent(t *testing.T) {
-	wantErr := "json: Unmarshal(nil)"
-
-	err := json.Unmarshal(toBytes("{}"), nil)
-
-	if err != nil {
-		if diff := cmp.Diff(wantErr, err.Error()); diff != "" {
-			t.Errorf("unexpected error (-want, +got) = %v", diff)
-		}
-		return
-	}
-}
-
 func toBytes(body interface{}) []byte {
 	b, err := json.Marshal(body)
 	if err != nil {
--- a/pkg/cloudevents/transport/amqp/codec_test.go
+++ b/pkg/cloudevents/transport/amqp/codec_test.go
@@ -371,7 +371,7 @@ func TestCodecRoundTrip(t *testing.T) {
 					got.Data = data
 				}
 
-				if tc.wantErr != nil || err != nil {
+				if tc.wantErr != nil {
 					if diff := cmp.Diff(tc.wantErr, err); diff != "" {
 						t.Errorf("unexpected error (-want, +got) = %v", diff)
 					}
@@ -509,7 +509,7 @@ func TestCodecAsMiddleware(t *testing.T) {
 					got.Data = data
 				}
 
-				if tc.wantErr != nil || err != nil {
+				if tc.wantErr != nil {
 					if diff := cmp.Diff(tc.wantErr, err); diff != "" {
 						t.Errorf("unexpected error (-want, +got) = %v", diff)
 					}
--- a/pkg/cloudevents/transport/http/codec_test.go
+++ b/pkg/cloudevents/transport/http/codec_test.go
@@ -819,7 +819,7 @@ func TestCodecRoundTrip(t *testing.T) {
 					got.Data = data
 				}
 
-				if tc.wantErr != nil || err != nil {
+				if tc.wantErr != nil {
 					if diff := cmp.Diff(tc.wantErr, err); diff != "" {
 						t.Errorf("unexpected error (-want, +got) = %v", diff)
 					}
@@ -961,7 +961,7 @@ func TestCodecAsMiddleware(t *testing.T) {
 						got.Data = data
 					}
 
-					if tc.wantErr != nil || err != nil {
+					if tc.wantErr != nil {
 						if diff := cmp.Diff(tc.wantErr, err); diff != "" {
 							t.Errorf("unexpected error (-want, +got) = %v", diff)
 						}
--- a/pkg/cloudevents/transport/http/message.go
+++ b/pkg/cloudevents/transport/http/message.go
@@ -35,7 +35,10 @@ func (m Message) CloudEventsVersion() string {
 	if m.Header != nil {
 		// Try headers first.
 		// v0.1, cased from the spec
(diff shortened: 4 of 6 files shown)


pkg/cloudevents/transport/amqp/codec_test.go
```diff
@@ -371,7 +371,7 @@ func TestCodecRoundTrip(t *testing.T) {
 					got.Data = data
 				}
 
-				if tc.wantErr != nil || err != nil {
```
n3wscott: This does not catch the case where wantErr is nil and err is not.

pkg/cloudevents/transport/amqp/codec_test.go
```diff
@@ -509,7 +509,7 @@ func TestCodecAsMiddleware(t *testing.T) {
...
```
n3wscott: This does not catch the case where wantErr is nil and err is not.

pkg/cloudevents/transport/http/codec_test.go
```diff
@@ -819,7 +819,7 @@ func TestCodecRoundTrip(t *testing.T) {
 					got.Data = data
 				}
 
-				if tc.wantErr != nil || err != nil {
+				if tc.wantErr != nil {
 					if diff := cmp.Diff(tc.wantErr, err); diff != "" {
 						t.Errorf("unexpected error (-want, +got) = %v", diff)
```
n3wscott: This does not catch the case where wantErr is nil and err is not.

pkg/cloudevents/transport/http/codec_test.go
```diff
@@ -961,7 +961,7 @@ func TestCodecAsMiddleware(t *testing.T) {
 						got.Data = data
 					}
 
-					if tc.wantErr != nil || err != nil {
```
n3wscott: This does not catch the case where wantErr is nil and err is not.

pkg/cloudevents/event_marshal_test.go
```diff
@@ -489,19 +489,6 @@ func TestUnmarshal(t *testing.T) {
 	}
 }
 
-func TestUnmarshal_nilEvent(t *testing.T) {
```
n3wscott: why delete the test?

pkg/cloudevents/transport/http/message.go
```diff
@@ -35,7 +35,10 @@ func (m Message) CloudEventsVersion() string {
 	if m.Header != nil {
 		// Try headers first.
 		// v0.1, cased from the spec
-		if v := m.Header["CE-CloudEventsVersion"]; len(v) == 1 {
```
n3wscott: why change this?

pkg/cloudevents/transport/nats/codec_test.go
```diff
@@ -237,7 +237,7 @@ func TestCodecRoundTrip(t *testing.T) {
 					got.Data = data
 				}
 
-				if tc.wantErr != nil || err != nil {
```
n3wscott: This does not catch the case where wantErr is nil and err is not.

pkg/cloudevents/transport/nats/codec_test.go
```diff
@@ -375,7 +375,7 @@ func TestCodecAsMiddleware(t *testing.T) {
...
```
n3wscott: This does not catch the case where wantErr is nil and err is not.

test/http/loopback.go
```diff
@@ -151,3 +151,7 @@ func printTap(t *testing.T, tap *tapHandler, testID string) {
 		}
 	}
 }
+
+// Don't consider printTap as dead code even if not currently in use.
+// We want to keep it for possible future debugging.
+var _ = printTap
```
n3wscott: nice.

pkg/cloudevents/transport/amqp/codec_test.go
alanconway: That can't happen, on line 354 we have if err != nil { ... return } so if we get here err == nil.
That's why the linter is complaining - it's an impossible condition.

pkg/cloudevents/transport/amqp/codec_test.go
alanconway: See https://github.com/cloudevents/sdk-go/pull/163/files/1807d618dcaa5673fd594d50c78e8a89b7fd3039#r306909479

pkg/cloudevents/transport/http/message.go
alanconway: go vet is checking for incorrect capitalization of string literals used as HTTP headers. If you put the value in a variable it shuts up.

pkg/cloudevents/event_marshal_test.go
alanconway: It is not testing anything in this package - it's verifying the documented behavior of json.Unmarshal()
I'm hoping the json package has it's own tests :)

pkg/cloudevents/transport/http/codec_test.go
alanconway: See https://github.com/cloudevents/sdk-go/pull/163/files/1807d618dcaa5673fd594d50c78e8a89b7fd3039#r306909479

pkg/cloudevents/transport/http/codec_test.go
alanconway: See https://github.com/cloudevents/sdk-go/pull/163/files/1807d618dcaa5673fd594d50c78e8a89b7fd3039#r306909479

pkg/cloudevents/transport/nats/codec_test.go
alanconway: See https://github.com/cloudevents/sdk-go/pull/163/files/1807d618dcaa5673fd594d50c78e8a89b7fd3039#r306909479

pkg/cloudevents/transport/nats/codec_test.go
alanconway: See https://github.com/cloudevents/sdk-go/pull/163/files/1807d618dcaa5673fd594d50c78e8a89b7fd3039#r306909479

n3wscott:
LGTM

n3wscott:
Thanks @alanconway for the cleanup and the explanations.

Issue #162, opened 2019-07-12
Message interface for inter-transport use.


alanconway:
Right now the Message interface  can't be used without casting to a transport-specific interface. Its only operation (GetVersion) is useless without additional knowledge about the transport-specific implementation, from which you could just as well get the version directly.

Propose an enhanced Message interface that can be used independently of the transport that created it, in particular as a go-between when forwarding messages from one transport to another.

In particular this solves the following problems that can't be solved with the existing API:

- forward structured event encodings *without decoding* ; provided both transports are happy with the encoding format and don't need to validate or inspect the content. The resulting performance benefits are the main (only?) reason for using structured over binary bindings.
- negotiate reliable qualities of service between different transports with similar QoS levesl (at-most-once, at-least-once, exactly-once) Although each transport can use its own mechanisms to ensure the QoS there is a need to hand-over responsibility between the transports, else messages can be lost if the process doing the forwarding crashes.

Code and examples are nearly complete - coming soon as  PR.

alanconway:
Code and examples in PR #164. The idea is not to push this PR but to use it to focus discussion, we will need to agree on how/if to integrate the various ideas into the main transport package.

n3wscott:
When I was writing this originally I had never thought to pass a transport.Message to any other transport because the impl of the Message object is specific to a transport, and created after passing an Event through a Codec. The interface allows a codec to talk to a transport without being aware for the most part.

alanconway:
> When I was writing this originally I had never thought to pass a transport.Message to any other transport because the impl of the Message object is specific to a transport, and created after passing an Event through a Codec. The interface allows a codec to talk to a transport without being aware for the most part.

Yep, this proposal is to augment the interface so they become useful between transports. The codec still makes sense as a separate entity for users that want to go `Event -> http.Request` or `amqp.Message -> Event` and use their own http or amqp client/server code rather than the Transport/Sender/Receiver we provide.

alanconway:
Fixed by #164

Issue #160, opened 2019-07-12
Drop cloudevents.EventResponse


alanconway:
TL;DNR: The cloudevents/sdk-go API should follow message-oriented examples (AMQP, MQTT, Kafka) *not* RPC-oriented examples (HTTP, gRPC). That means separate, one-way Send and Receive operations with no implied coupling between sending one event and receiving another (no EventReponse). The only possible result from Send is an error indicating a delivery problem *not* a response containing data from a remote application.

# Details

"An  [event](https://github.com/cloudevents/spec/blob/v0.3/spec.md#event) is a data record expressing an occurrence and its context." It is *not* a request for service, it does not demand any "response". It simply informs the next link in the chain that something has happened.

We already have HTTP RPC in the cloud, what we need from cloudevents is an alternative loosely-coupled event/message-oriented model. The critical difference from RPC is that sending an event does not imply you will receive one, and receiving an event does not make you responsible for sending one. This is the key to "loose" coupling:

*Decoupled/asynchronous processing*: A sender can send events without knowing how *or when* they are processed. A receiver can process an event immediately, store it for later, forward it, generate new events etc. There's no expectation of matching request-response pairs that ties sender and receiver actions together.

*Routability*: events are easily routed, stored and forwarded, multicast or load-balanced by intermediaries between the sender and receiver. Routing RPC is more difficult because.

- you have to construct a "reverse route" to get the response back, dealing with problems like  one-way firewalls, no-longer-connected clients and requests/responses that get held up in queues.
- you have to fabricate exactly one response from 0, 1 or many (multicast) responses.

Of course the async eventing model introduces complexities that don't exist in RPC,  just as concurrent programming introduces complexities that don't exist in sequential programming. We need both.

It's not sufficient to say the EventResponse parameter is "optional". Its existence implies that API users can use it and API implementers should provide it. That's complicates implementations based on message-oriented middleware like AMQP, MQTT or Kafka, and it encourages API users to view cloudevents as a funny API for HTTP - which makes cloudevents useless, since we already have HTTP.

If there are some strong use cases for an RPC-like exchange of events then they should be implemented using standard HTTP RCP and standard HTTP APIs. The cloudevents APIs should be reserved for event-oriented use cases where we don't want to assume RPC-semantics, or  require implementations to support them.

n3wscott:
How do you suggest implementing filter functions with http and cloudevents?

n3wscott:
Knative uses this use case extensively: 

[...]

PR #164, opened 2019-07-16
Code and examples for proposed changes/additions to transport APIs

alanconway:
See Issu "Message interface for inter-transport use." #162

This PR is not intended to be pushed, it illustrates some ideas directly as working code and examples. As we agree on if/how the ideas  should be integrated, I'll submit PRs to merge them into the main transport package.

It is a little easier to read as godoc, for convenience I put it here https://alanconway.github.io/godoc/pkg/github.com/cloudevents/sdk-go/pkg/cloudevents/transport/x/index.html
or you can use godoc to generate it locally

### DIFF
--- a/pkg/cloudevents/transport/x/doc.go
+++ b/pkg/cloudevents/transport/x/doc.go
@@ -1,0 +1,49 @@
+/*
+
+Package transport/x shows proposed modifications to package transport.
+
+They are here for discussion purposes, anything that we can agree on
+should be merged into the transport package and may involve
+incompatible changes.
+
+Goals
+
+Support for intermediaries that receive and send events via different
+transports, for example adapters, importers, brokers or channels.
+
+Support for reliable messaging qualities of service between different
+transports
+
+Provide the option of blocking Receive vs. callback SetReceiver.
+There are many situations where the blocking style is easier to
+implement and work with.
+
+Overview of changes
+
+Sender/Receiver interfaces that can be implemented and used separately.
+Implementations can easily be combined into a single Transport that does both
+(e.g. for use by the client package)
+
+Message.Event() allows conversion from any message to the common
+in-memory Event representation, without knowing its transport/codec of
+origin. Explict use of Codecs is not required. (Transports should
+still provide a way for users to build/interpret transport messages if they
+want to bypass our transport APIs, but that's a separate issue.)
+
+Message.Structured() optionally allows a structured encoding to be
+forwarded without decoding. This is a special case, but an important
+one.
+
+Reliable messaging: Transport implementation must provide the actual
+reliability (resends, storing delivery state etc.) the Message
+interface allows hand-over responsibility between a Receiver and a
+Sender on different transports so we don't create a "hole" in the
+guarantee. The overall guarantee is the weaker of the sender and
+receiver.
+
+No reliability examples yet, wanted to get some feedback before going further.
+*/
+package x
+
+// FIXME(alanconway) review - check for "2" suffix
+// Finish comments.
--- a/pkg/cloudevents/transport/x/example_implementing_test.go
+++ b/pkg/cloudevents/transport/x/example_implementing_test.go
@@ -1,0 +1,70 @@
+package x_test
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+
+	"github.com/cloudevents/sdk-go/pkg/cloudevents"
+	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport"
+	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport/x"
+)
+
+// ExMessage is a json.RawMessage, which is just a byte slice
+// containing a JSON encoded event.
+type ExMessage struct{ json.RawMessage }
+
+func (m ExMessage) Structured() (string, []byte) {
+	return cloudevents.ApplicationCloudEventsJSON, []byte(m.RawMessage)
+}
+
+func (m ExMessage) Event() (e cloudevents.Event, err error) {
+	err = json.Unmarshal(m.RawMessage, &e)
+	return e, err
+}
+
+func (m ExMessage) Finish(error) {}
+
+// ExSender sends by writing JSON encoded events to an io.Writer
+type ExSender struct{ *json.Encoder }
+
+func NewExSender(w io.Writer) ExSender { return ExSender{json.NewEncoder(w)} }
+
+func (s ExSender) Send(ctx context.Context, m x.Message) error {
+	if t, b := m.Structured(); t != "" {
+		// Fast case: if the Message is already structured JSON we can
+		// send it directly, no need to decode and re-encode. Encoding a
+		// json.RawMessage to a json.Encoder() is just a byte-buffer copy.
+		return s.Encode(json.RawMessage(b))
+	} else {
+		// Some other message encoding. Decode as a generic cloudevents.Event
+		// and then re-encode as JSON
+		if e, err := m.Event(); err != nil {
+			return err
+		} else if err := s.Encode(e); err != nil {
+			return err
+		}
+		return nil
+	}
+}
+
+// ExReceiver receives by reading JSON encoded events from an io.Reader
+type ExReceiver struct{ *json.Decoder }
+
+func NewExReceiver(r io.Reader) ExReceiver { return ExReceiver{json.NewDecoder(r)} }
+
+func (sr ExReceiver) Receive(context.Context) (x.Message, error) {
+	var m ExMessage
+	err := sr.Decode(&m) // This is just a byte copy since m is a json.RawMessage
+	return m, err
+}
+
+// NewExTransport returns a transport.Transport which is implemented by
+// an ExSender and an ExReceiver
+func NewExTransport(r io.Reader, w io.Writer) transport.Transport {
+	return x.NewTransport(NewExSender(w), NewExReceiver(r))
+}
+
+// Example of implementing a transport including a simple message type,
+// and a transport sender and receiver.
+func Example_implementing() {}
--- a/pkg/cloudevents/transport/x/example_using_test.go
+++ b/pkg/cloudevents/transport/x/example_using_test.go
@@ -1,0 +1,128 @@
+package x_test
+
+import (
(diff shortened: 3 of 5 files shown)


alanconway:
This is a little easier to read as godoc, for convenience I generated it here https://alanconway.github.io/godoc/pkg/github.com/cloudevents/sdk-go/pkg/cloudevents/transport/x/index.html
or you can use godoc to generate it locally

alanconway:
@n3wscott ping

n3wscott:
Oh! I was traveling. Sorry I missed this. 

Let me look.

pkg/cloudevents/transport/x/example_implementing_test.go
```diff
@@ -0,0 +1,70 @@
...
+type ExSender struct{ *json.Encoder }
+
+func NewExSender(w io.Writer) ExSender { return ExSender{json.NewEncoder(w)} }
+
+func (s ExSender) Send(ctx context.Context, m x.Message) error {
+	if t, b := m.Structured(); t != "" {
+		// Fast case: if the Message is already structured JSON we can
+		// send it directly, no need to decode and re-encode. Encoding a
+		// json.RawMessage to a json.Encoder() is just a byte-buffer copy.
+		return s.Encode(json.RawMessage(b))
```
n3wscott: there is no impl for `Encode`. I am confused why you would need to encode message. In the current SDK, message is created by passing an event through a codec.

pkg/cloudevents/transport/x/doc.go
```diff
@@ -0,0 +1,49 @@
...
+There are many situations where the blocking style is easier to
+implement and work with.
+
+Overview of changes
+
+Sender/Receiver interfaces that can be implemented and used separately.
+Implementations can easily be combined into a single Transport that does both
+(e.g. for use by the client package)
+
+Message.Event() allows conversion from any message to the common
```
n3wscott: I don't think you can do this without understanding both the transport and the codec.

pkg/cloudevents/transport/x/doc.go
```diff
@@ -0,0 +1,49 @@
...
+Support for reliable messaging qualities of service between different
+transports
+
+Provide the option of blocking Receive vs. callback SetReceiver.
+There are many situations where the blocking style is easier to
+implement and work with.
+
+Overview of changes
+
+Sender/Receiver interfaces that can be implemented and used separately.
```
n3wscott: I would like to know how you can see re-implementing send/response for http.

pkg/cloudevents/transport/x/message.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// Received is called by a Sender when it gets acknowledgment of receipt
+	// (e.g. AMQP ACCEPT or MQTT PUBREC)
+	//
+	// The sender passes a finish() function that the original receiver
+	// must call when it get's the ack-of-the-ack (e.g. AMQP SETTLE, MQTT
+	// PUBCOMP)
+	//
+	// If sending fails, the sender must call Finish(err) with a non-nil error
+	// instead of Ack. ExactlyOnceMessage implementations must also be prepared to
```
n3wscott: Where is `Ack`?

pkg/cloudevents/transport/x/example_using_test.go
```diff
@@ -0,0 +1,128 @@
...
+// structured events, and to exchange delivery status between reliable
+// transports. Even transports using different protocols can ensure
+// reliable delivery.
+//
+func runIntermediary(r io.Reader, w io.WriteCloser) error {
+	for {
+		receiver := NewExReceiver(r)
+		sender := NewExSender(w)
+		for i := 0; i < count; i++ {
+			if m, err := receiver.Receive(context.TODO()); err != nil {
```
n3wscott: Oh, I see, the http transport could send and the resp could be queued and ready for the next Receive call.

n3wscott:
What do you think about making a branch in this repo to start staging this stuff and then we can work together on it, or a branch in your repo.

I think I still need to be convinced of the removal of codecs as an abstraction of how to turn events into messages. But codecs could be internal details of the transport Message objects rather than how I have it as a detail of the transport. 

Let's keep going on this idea. I think it could turn into a much easer adaptation for non-http transports.

Thanks for the detailed comments in the PR!

pkg/cloudevents/transport/x/example_implementing_test.go
alanconway: Encode here is the standard json.Encoder.Encode() - the naming is confusing. 

The new Message is an abstraction that can be passed *between* transports, so a Sender may be sending a Message that was encoded by a foreign Receiver. Message advertises if it already contains a structured encoding that the sender can potentially re-use with no re-encoding in the optimized case, but otherwise the sender will call Event() to extract (and possibly decode using some foreign codec)  the Event and re-encode it using its own Codec. I didn't implement an explicit Codec in the example, I just used json calls directly - I probably should for clarity.

Codec still has  role in creating the internal message format for each transport, and may be used directly by users that just want to create e.g. a HTTP message and then use their own client or server code. However for communicating from a Receiver to a Sender, the relevant codecs are used internally by each one, and the Message provides a uniform way to decode the message without knowing what Codec was used to create it.

pkg/cloudevents/transport/x/doc.go
alanconway: Correct - every Message implementation does understand the transport/codec it originated from - it is part of the transport implementation. Either Message.Event() would be implemented using  Codec.Decode() or vice-versa, either way the API is uniform.

Message.Event() == Codec.Decode(message)

The reason for introducing Message.Event() is to allow a Sender to decode  a message that originated from a different Receiver transport without needing to know how it was encoded. This also supports other Message implementations like EventMessage which is simply a Message interface wrapper for a locally-created Event object.

pkg/cloudevents/transport/x/doc.go
alanconway: I would like response events to go away, see https://github.com/cloudevents/sdk-go/issues/160
Can you comment on that issue with some details of the use-cases for response events? If it is a requirement it will certainly be possible to build it, but IMO it should be an optional extra mode of operation, not presented as the basic API. I would probably do something like:

type RequestMessage interface { Response(Message) }

basically it becomes another QoS alternative where we require a complete Message in response, not just some protocol-specific ack sequence. 

I'd like to  understand the use cases better though - if there is a real requirement for portable, routable response events then the cloudevents spec needs to be updated with a correlationId. The problem of routing responses may be bigger than a single client-server connection.

pkg/cloudevents/transport/x/message.go
alanconway: My bad, I went through some renaming spasms. That should say "Received"

n3wscott:
I have cut a [0.8.0 release](https://github.com/cloudevents/sdk-go/releases/tag/v0.8.0) to make room for these changes to come in.

pkg/cloudevents/transport/x/example_implementing_test.go
n3wscott: ah, ok so Message has to have some state in it to define in which transport the message is encoded so it can select the correct codec to convert it out. 

It seems like a really interesting feature to be able to move the Message between transports. Middleware will be much faster this way too!

alanconway:
> What do you think about making a branch in this repo to start staging this stuff and then we can work together on it, or a branch in your repo.
> 
> I think I still need to be convinced of the removal of codecs as an abstraction of how to turn events into messages. But codecs could be internal details of the transport Message objects rather than how I have it as a detail of the transport.

My bad - I'm not proposing the removal of Codec, I just didn't bother to write an explict one in the example - fixing.

> Let's keep going on this idea. I think it could turn into a much easer adaptation for non-http transports.
>
Yes - the biggest variable here is what we do about  #160. I think adding a RequestMessage interface might be a good compromise for the #160 use cases while removing responses from the main transport  API, but I would like to understand the requirement for response events better before I dive into that.

> Thanks for the detailed comments in the PR!

I'll update with explicit Codec and RequestMessage { Response() }to see how that looks. I'm thinking that a response-capable transport (HTTP) would do if m, ok :=  m.(RequestMessage) { respond } whereas response-ignorant transports would blindly Finish() the message, in which case the requester can decide if that's an error, or if it will manufacture a fake response for the ultimate clien.t

n3wscott:
Moving Message out of the transport also leads to interesting developments around sharing codec that are very common in strategy, like binary mode in AMQP, PubSub and HTTP all use the equivalent of headers to send the data, the Key value is different in each, but the strategy is the same...

pkg/cloudevents/transport/x/example_using_test.go
alanconway: Yep - key thing here is how to present the request/response correlation in the API. I'd like it not to be the first thing the users see, or perceived as the "normal" way of sending events, but I think we can find a way to provide it as an optional feature for transports (like HTTP) that are RCP underneath.

> Moving Message out of the transport also leads to interesting developments around sharing codec that are very common in strategy, like binary mode in AMQP, PubSub and HTTP all use the equivalent of headers to send the data, the Key value is different in each, but the strategy is the same...

Agreed, e.g. the `if Structured() else Event()` dance could be part of a generic library usable by all transports that allow structured tunneling.  Also the correct handling of different QoS options definitely should be captured generically, each transport just calls some relevant functions when it gets it's transport-specific ack indicator.

alanconway:
> Moving Message out of the transport also leads to interesting developments around sharing codec that are very common in strategy, like binary mode in AMQP, PubSub and HTTP all use the equivalent of headers to send the data, the Key value is different in each, but the strategy is the same...

Agreed, e.g. the `if Structured() else Event()` dance could be part of a generic library usable by all transports that allow structured tunneling.  Also the correct handling of different QoS options definitely should be captured generically, each transport just calls some relevant functions when it gets it's transport-specific ack indicator.

I'll probably do one or two more pushes in the current form to continue the discussion points so far, then try merging it into the main transport package and re-implementing some existing transports as you suggest.

n3wscott:
LGTM

Let's start this plan.

matzew:
:tada:

PR #165, opened 2019-07-17
Fix segfault if ce-specversion header missing

prestona:
If a HTTP request does not contain the `ce-specversion` header then the
process of converting the request to an event returns a `nil`
event and a non-fatal error value. Unfortunately, in this situation,
the code continued to process the request resulting in an attempt to
de-reference the `nil` event value.

Fix by adding an additional condition to the error handling. Now
conversions that yield a non-fatal error must also return a non-nil
event, otherwise a HTTP 400 (Bad Request) status is returned.


n3wscott:
oh nice catch!

/lgtm

PR #169, opened 2019-08-06
Add a method to disable tracing.

### LINKED ISSUE #168: tracing is too much
for network traces using opencensus and using the sdk, you get profile level tracing when you want only network level tracing.

Add a mode to tracing for consumers to turn off or adjust.

n3wscott:
For applications that want to use tracing to do network level traces and not profile traces, the sdk pushes too much data. 

Fixes #168 

cc: @Harwayne

### DIFF
--- a/alias.go
+++ b/alias.go
@@ -7,6 +7,7 @@ import (
 	"github.com/cloudevents/sdk-go/pkg/cloudevents"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/client"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/context"
+	"github.com/cloudevents/sdk-go/pkg/cloudevents/observability"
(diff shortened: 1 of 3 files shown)


markpeek:
@Harwayne give the 80% of trace lines comment, should tracing for this SDK default to "disabled" instead of "enabled"?

Harwayne:
> @Harwayne give the 80% of trace lines comment, should tracing for this SDK default to "disabled" instead of "enabled"?

I will definitely disable the tracing for my code. But there might be others that want this level of tracing. Whether my use is more typical or not, I don't know. I think those working on this repo have a better understanding of what the typical usage will be, so I'll defer to your judgement as to the default.

n3wscott:
Default is set to tracing off. It is really useful for profiling and perhaps not useful for normal application running.

Code updated.

PR #171, opened 2019-08-11
Enabling Pub/Sub to send and receive from multiple topics and subscriptions.

n3wscott:
In Knative, we need to be able to reuse a single CloudEvents client to listen to several subscriptions to allow the infrastructure to be cheaper and multi-tenant for a Google Cloud single secret.

This work enables a client to be configured with `n` Subscriptions and `StartReceiver` starts `n` go threads, and then blocks until the parent context has been canceled, canceling the threads and collecting the errors.

This change also enables a `topic` to be injected into the context of `client.Send`, redirecting the client to publish on that topic. This will allow a single client to target many topics and the transport deals with managing the topic instances. 

Creation applies to the entire client.

Deletion are supported inside the connection object but it has not been wired up into an easy way to use it. This is a less (if at all) used feature and will be enabled in the future.


cmd/samples/pubsub/multireceiver/main.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	log.Println("Created client, listening...")
+
+	if err := c.StartReceiver(ctx, receive); err != nil {
+		log.Fatalf("failed to start pubsub receiver, %s", err.Error())
+	}
+}
+
+/*
```
markpeek: Should these comments be at the top of the file to describe what it does and how to run it?

cmd/samples/pubsub/multisender/main.go
```diff
@@ -0,0 +1,77 @@
...
+	"context"
+	"log"
+	"os"
+	"strings"
+
+	"github.com/kelseyhightower/envconfig"
+
+	"github.com/cloudevents/sdk-go"
+	cecontext "github.com/cloudevents/sdk-go/pkg/cloudevents/context"
+	cloudeventspubsub "github.com/cloudevents/sdk-go/pkg/cloudevents/transport/pubsub"
```
markpeek: For consistency with other changes in this PR, this should be abbreviated "cepubsub".

cmd/samples/pubsub/multisender/main.go
```diff
@@ -0,0 +1,77 @@
...
+		if err != nil {
+			log.Printf("failed to send: %v", err)
+			os.Exit(1)
+		}
+	}
+
+	os.Exit(0)
+}
+
+/*
```
markpeek: Same comment about top of file.

cmd/samples/pubsub/receiver/main.go
```diff
@@ -64,6 +65,23 @@ func main() {
 	log.Println("Created client, listening...")
 
 	if err := c.StartReceiver(ctx, receive); err != nil {
-		log.Fatalf("failed to start nats receiver, %s", err.Error())
+		log.Fatalf("failed to start pubsub receiver, %s", err.Error())
 	}
 }
+
```
markpeek: same

pkg/cloudevents/transport/pubsub/internal/conneciton.go
```diff
@@ -0,0 +1,189 @@
+package internal
```
markpeek: filename should be "connection.go"?

PR #173, opened 2019-08-22
Trying to fix extensions... wip

n3wscott:


alanconway:
FYI https://github.com/cloudevents/spec/issues/508 - I think the spec needs some more baking regarding extensions.

alanconway:
Current issues that are affecting my testing: JSON marshal unmarshal turns extension values into raw JSON binary values, instead of canonical strings. Per latest spec and types.Value work, the approach should be to always use either canonical strings or the "preferred" go types defined in types/doc.go, and to use types.ValueOf() when in doubt to smooth over possible conversion issues.

n3wscott:
We have done a lot of work around extensions and this is no longer needed.

PR #177, opened 2019-08-26
Changing Send signature: Adding Context return param

nachocano:
- Send now returns a context, which each transport can populate with specific information. In the case of the HTTP transport, we are setting the StatusCode and Header of the responses for now.
- Updated UTs.


nachocano:
ping

nachocano:
ping

Issue #180, opened 2019-09-03
Improved AMQP transport


alanconway:
Required features:
- Refactor based on transport/x API - https://github.com/cloudevents/sdk-go/pull/164
- Configuration: client and sever modes (both sides), named links, anonymous relay.
- Delivery QoS: 0 (unreliable) and 1 (at-least-once)
- Codec: Event/AMQP translation for direct AMQP applications.

Nice-to-have (for this iteration):
- Security config: user/pass, SASL, TLS certificates.
- Request-response support (depends on https://github.com/cloudevents/sdk-go/issues/160)

Non-goals  (for this iteration):
- QoS 2 (exactly-once)
- Transactions

alanconway:
@n3wscott I'm thinking of renaming the transport/x package as pkg/message. That avoids clash with existing use  or "transport", fits well with ce spec use of "message" and reads pretty well in code as in "message.Sender", "message.Receiver". message.Message stutters a little but makes sense. What do you think?

n3wscott:
sounds good to me!
-Scott

alanconway:
@n3wscott existing AMQP codec uses string for timestamps and json-encodes all extensions - that doesn't match what it says in https://github.com/cloudevents/spec/blob/master/amqp-format.md which says to use the corresponding native AMQP types (timestamp, map etc.). OK if I fix, or am I missing something?

alanconway:
Switched package x name from "message" to "binding", it seemed a little odd in use, the package provides more than just the message.

alanconway:
Split out #226 for generic binding framework.

PR #181, opened 2019-09-04
Renamed pkg/cloudevents/transport/x to pkg/binding

alanconway:
Main types are binding.Message, binding.Sender and binding.Receiver which
reads well in code, corresponds to CE spec use of the term "binding", and
avoids clashes with existing package name "transport".

Part of #180 - The AMQP transport will be re-factored to use this package.


n3wscott:
LGTM

PR #187, opened 2019-09-13
Parse timestamp can return nil. Check for nil.

n3wscott:
@nachocano found a npe in his logs. I think this will fix.

### DIFF
--- a/pkg/cloudevents/transport/amqp/codec.go
+++ b/pkg/cloudevents/transport/amqp/codec.go
@@ -196,17 +196,17 @@ func (c Codec) fromHeaders(h map[string]interface{}, event *cloudevents.Event) e
 	}
 	delete(h, prefix+"source")
 
-	if t, ok := h[prefix+"time"].(string); ok { // TODO: time can be empty
-		timestamp := types.ParseTimestamp(t)
-		if err := ec.SetTime(timestamp.Time); err != nil {
-			return err
+	if t, ok := h[prefix+"time"].(string); ok {
+		if timestamp := types.ParseTimestamp(t); timestamp != nil {
+			if err := ec.SetTime(timestamp.Time); err != nil {
+				return err
+			}
 		}
 	}
 	delete(h, prefix+"time")
 
-	if t, ok := h[prefix+"schemaurl"].(string); ok {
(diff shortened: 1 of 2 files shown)


nachocano:
LGTM

pkg/cloudevents/transport/amqp/codec.go
```diff
@@ -196,17 +196,17 @@ func (c Codec) fromHeaders(h map[string]interface{}, event *cloudevents.Event) e
 	}
 	delete(h, prefix+"source")
 
-	if t, ok := h[prefix+"time"].(string); ok { // TODO: time can be empty
-		timestamp := types.ParseTimestamp(t)
-		if err := ec.SetTime(timestamp.Time); err != nil {
-			return err
+	if t, ok := h[prefix+"time"].(string); ok {
+		if timestamp := types.ParseTimestamp(t); timestamp != nil {
```
markpeek: Right now ParseTimestamp() doesn't return an error. It's debatable on what to do if there is a bad time being presented whether to fail entirely or, as in the case, silently drop. Perhaps open an issue?

pkg/cloudevents/transport/pubsub/codec.go
```diff
@@ -187,16 +187,16 @@ func (c Codec) fromAttributes(a map[string]string, event *cloudevents.Event) err
 	delete(a, prefix+"source")
 
 	if t := a[prefix+"time"]; t != "" {
```
markpeek: This looks suspiciously similar to the amqp codec code. Perhaps there needs to be helpers in the main code for the codecs to use.

n3wscott: they use a slightly different mapping ... grrrr

I have some ideas how to combine them and pass a mapping decoder ring for the future.

pkg/cloudevents/transport/amqp/codec.go
n3wscott: https://github.com/cloudevents/sdk-go/issues/188

n3wscott: good points, I was optimizing for moving fast but it is true that this could catch someone off guard... I caught myself off guard. :D

PR #194, opened 2019-09-20
Starting to add 1.0

n3wscott:
This PR adds the v1.0 context for conforming to the new release of CloudEvents 1.0-rc1 

This is the first of many updates to migrate the SDK to `v1.0`.


pkg/cloudevents/event_reader.go
```diff
@@ -81,15 +81,15 @@ func (e Event) DataMediaType() string {
...
+// DeprecatedDataContentEncoding implements EventReader.DeprecatedDataContentEncoding
+func (e Event) DeprecatedDataContentEncoding() string {
 	if e.Context != nil {
-		return e.Context.GetDataContentEncoding()
+		return e.Context.DeprecatedGetDataContentEncoding()
 	}
 	return ""
 }
 
-// DataContentEncoding implements EventReader.DataContentEncoding
```
markpeek: Likely this was an earlier type since it should be for Extensions().

n3wscott: Fixed, nice catch.

PR #204, opened 2019-09-23
Package convert implements CloudEvents type-system conversions.

### LINKED ISSUE #192: Provide uniform API for cloudevents type system
We need a uniform way to represent the conversions to/from string and to/from Go types for the types described in the cloudevents spec section "Type System".

I'm working on this as part of the improved AMQP binding, will reference this Issue from the relevant PR.

alanconway:
Implements the canonical string encodings specified by the CE spec.

Establishes a "preferred" Go type to represent each CE type and a set of conversions from other "compatible" Go types. Example: preferred type for CloudEvents Integer is int32, all Go scalar numeric types (excludes complex) can be converted to it, with range checks.

Fixes #192

NOTE: the package is under the new bindings/ hierarchy for now, probably it
should be merged into the existing "cloudevents/types" package. The binding
hierarchy is intended for more advanced/low-level use (implementing new
transports and the like) but the conversions are of wider interest. The CE spec
suggests that the canonical-string conversions should be a feature of any CE
library.

### DIFF
--- a/pkg/binding/convert/convert.go
+++ b/pkg/binding/convert/convert.go
@@ -1,0 +1,266 @@
+/*
+Package convert provides conversions specified by the CloudEvents type system.
+
+Each CloudEvents type has a preferred Go type, a set of convertible Go types,
+and a canonical string-encoding. This package provides conversions between these
+representations.
+
+In the table below "types" refers to package "github.com/cloudevents/sdk-go/pkg/cloudevents/types".
+
++----------------+---------+-----------------------------------+
+|CloudEvents Type|Go type  |Compatible types                   |
++----------------+---------+-----------------------------------+
+|Bool            |bool     |bool                               |
++----------------+---------+-----------------------------------+
+|Integer         |int32    |Any numeric type with value in     |
+|                |         |range of int32                     |
++----------------+---------+-----------------------------------+
+|String          |string   |string                             |
++----------------+---------+-----------------------------------+
+|Binary          |[]byte   |[]byte                             |
++----------------+---------+-----------------------------------+
+|URI             |url.URL  |url.URL, types.URIRef.             |
+|                |         |Must be an absolute URI.           |
++----------------+---------+-----------------------------------+
+|URI-Reference   |url.URL  |url.URL, types.URIRef              |
++----------------+---------+-----------------------------------+
+|Timestamp       |time.Time|time.Time, types.Timestamp         |
++----------------+---------+-----------------------------------+
+
+*/
+package convert
+
+import (
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"math"
+	"net/url"
+	"reflect"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/cloudevents/sdk-go/pkg/cloudevents/types"
+)
+
+// Bool converts from bool or a "true"/"false" string.
+func ToBool(v interface{}) (ret bool, err error) {
+	name := "Bool"
+	switch v := v.(type) {
+	case bool:
+		ret = v
+	case string:
+		switch v {
+		case "true":
+			ret = true
+		case "false":
+			ret = false
+		default:
+			err = valueErr(name, v)
+		}
+	default:
+		err = typeErr(name, v)
+	}
+	return ret, err
+}
+
+// Integer converts from any numeric value that fits in range of int32 or
+// a string in strconv.Atoi format.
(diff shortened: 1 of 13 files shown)


pkg/binding/convert/convert.go
```diff
@@ -0,0 +1,258 @@
...
+|Bool            |bool     |bool                               |
++----------------+---------+-----------------------------------+
+|Integer         |int32    |Any numeric type with value in     |
+|                |         |range of int32                     |
++----------------+---------+-----------------------------------+
+|String          |string   |string                             |
++----------------+---------+-----------------------------------+
+|Binary          |[]byte   |[]byte                             |
++----------------+---------+-----------------------------------+
+|URI             |url.URL  |url.URL, types.URLRef.             |
```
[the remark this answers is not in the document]
alanconway: OK will include. I think this package might be best merged into the types package, so we have a single place for CE type-system definitions. I have it separate for now just to facilitate my hacking.

alanconway:
I moved the conversions into cloudevents/types and deprecated ExtensionAs in favour of using them.
I think it works quite well.

pkg/cloudevents/types/convert.go
```diff
@@ -0,0 +1,248 @@
+package types
```
n3wscott: this might not matter so much for 1.0 because the types for extensions are only string, or string encoded.

pkg/cloudevents/event.go
```diff
@@ -31,7 +31,17 @@ func New(version ...string) Event {
 	return *e
 }
 
-// ExtensionAs returns Context.ExtensionAs(name, obj)
+// DEPRECATED: Access extensions directly via the e.Extensions() map.
```
n3wscott: The plan in the SDK was not to just expose the fields, but the concept of an extension as a grouping of many keys in various places. 

I would like to not deprecate ExtensionAs, it is the flexible future.

pkg/cloudevents/event_interface.go
```diff
@@ -30,14 +30,17 @@ type EventReader interface {
 	// Extension Attributes
 
 	// Extensions returns the event.Context.GetExtensions().
+	// Extensions use the CloudEvents type system, details in package cloudevents/types.
```
n3wscott: In 1.0. extensions are only `map[string]string`

pkg/cloudevents/eventcontext.go
```diff
@@ -31,13 +31,21 @@ type EventContextReader interface {
...
-	// provided type, or if the type is not a supported.
+	// DEPRECATED: Access extensions directly via the GetExtensions()
+	// For example replace this:
+	//
+	//     var i int
+	//     err := ec.ExtensionAs("foo", &i)
+	//
+	// With this:
+	//
+	//     i, err := types.ToInteger(ec.GetExtensions["foo"])
```
n3wscott: this would be `i, err := types.ToInteger(ec.GetExtensions()["foo"])`

pkg/cloudevents/eventcontext_v1.go
```diff
@@ -48,7 +48,7 @@ type EventContextV1 struct {
 
 	// Extensions - Additional extension metadata beyond the base spec.
 	// +optional
-	Extensions map[string]string `json:"-"`
```
n3wscott: this is an intended change. in 1.0, only string is supported unless I am reading the spec wrong now.

pkg/cloudevents/eventcontext_v1.go
```diff
@@ -62,26 +62,27 @@ func (ec EventContextV1) ExtensionAs(name string, obj interface{}) error {
 	}
 
 	// Only support *string for now.
-	switch v := obj.(type) {
```
n3wscott: this was a switch to be able to add to it later.

pkg/cloudevents/types/convert.go
n3wscott: ok I think I am just confused: https://github.com/cloudevents/spec/blob/master/spec.md#extension-context-attributes

pkg/cloudevents/event_interface.go
n3wscott: I am confused: https://github.com/cloudevents/spec/blob/master/spec.md#extension-context-attributes

so keeping type between binary http back to strong types is going to be very hard.

n3wscott:
LGTM

I am half sold on the marking ExtensionsAs changes, but they currently do not work as is so I would like to explore this new way.

PR #216, opened 2019-09-26
types.Value wrapper for extension values

alanconway:
Better implementation of conversions centered around type Value.

Thanks for prodding me, this is much neater than the previous version.
We can return this directly:

    Event.Extension(string) types.Value
    Event.SetExtension(string, types.Value)

Or we can continue to use interface{} for extension values and users can do
ValueOf() and v.Interface() to convert back and forth.


pkg/cloudevents/types/value.go
```diff
@@ -0,0 +1,210 @@
+package types
```
n3wscott: I am pretty excited about this....

pkg/cloudevents/types/value.go
```diff
@@ -0,0 +1,210 @@
...
+	case string:
+		return x
+	case []byte:
+		return base64.StdEncoding.EncodeToString(x)
+	case url.URL:
+		return x.String()
+	case time.Time:
+		return x.UTC().Format(time.RFC3339Nano)
+	case nil:
+		return "<nil>"
```
n3wscott: I think this will be an issue for nil things because you might want to omit empty, but it will not be empty with "<nil>"

n3wscott: We can make an issue to follow up.

pkg/cloudevents/types/doc.go
```diff
@@ -1,40 +1,32 @@
...
+ |Bool            |bool            |bool                               |
+ +----------------+----------------+-----------------------------------+
+ |Integer         |int32           |Any numeric type with value in     |
+ |                |                |range of int32                     |
+ +----------------+----------------+-----------------------------------+
+ |String          |string          |string                             |
+ +----------------+----------------+-----------------------------------+
+ |Binary          |[]byte          |[]byte                             |
+ +----------------+----------------+-----------------------------------+
+ |URI-Reference   |url.URL         |url.URL, types.URIRef, types.URI   |
```
n3wscott: convertible from string too.

n3wscott: And URI

pkg/cloudevents/types/doc.go
```diff
@@ -1,40 +1,32 @@
...
-convertible Go type or the CloudEvents string representation. The StringOf()
-function converts any value to the corresponding CloudEvents string form. The
-Normalize() function converts convertible values to the "normal" type, for
-example it converts any legal numeric type to int32.
+ +----------------+----------------+-----------------------------------+
+ |CloudEvents Type|Native Type     |Convertible From Types             |
+ +================+================+===================================+
+ |Bool            |bool            |bool                               |
+ +----------------+----------------+-----------------------------------+
+ |Integer         |int32           |Any numeric type with value in     |
```
n3wscott: Convertible needs to include from string because of http binary headers.

pkg/cloudevents/types/doc.go
```diff
@@ -1,40 +1,32 @@
...
+ |String          |string          |string                             |
+ +----------------+----------------+-----------------------------------+
+ |Binary          |[]byte          |[]byte                             |
+ +----------------+----------------+-----------------------------------+
+ |URI-Reference   |url.URL         |url.URL, types.URIRef, types.URI   |
+ +----------------+----------------+-----------------------------------+
+ |URI             |url.URL         |url.URL, types.URIRef, types.URI   |
+ |                |                |Must be an absolute URI.           |
+ +----------------+----------------+-----------------------------------+
+ |Timestamp       |time.Time       |time.Time, types.Timestamp         |
```
n3wscott: Convertible needs to include from string because of http binary headers.

pkg/cloudevents/types/value.go
```diff
@@ -0,0 +1,210 @@
...
+		return x, nil
+	case string:
+		return base64.StdEncoding.DecodeString(x)
+	default:
+		return nil, typeErr("Binary", x)
+	}
+}
+
+// ToURIRef returns a url.URL value, decoding from string if necessary.
+func (v Value) ToURIRef() (url.URL, error) {
```
n3wscott: If this is converting to `url.URL`, I think it should be `ToURL`. Or make two methods that produce `types.URI` and `types.URIRef`

n3wscott:
LGTM

pkg/cloudevents/types/doc.go
alanconway: fixed in final story.

pkg/cloudevents/types/doc.go
alanconway: +1 the final ToX functions all take string implicitly - I'll clarify the doc.

pkg/cloudevents/types/value.go
alanconway: I'm not keen on that. The empty string is not a missing value, it is a legitimate value of type string. I agree that some bindings may use/interpret empty strings to mean missing value but that should be handled by code specific to those bindings, not ambiguated in the type system. I'd prefer nil to stringify as a broken, unparseable thing like <nil>. The surrounding code should figure out that it has no value *before* it tries to convert the value it doesn't have :) If it doesn't that's a bug and better fail than cover it up.

Issue #195, opened 2019-09-20
Migrate away from DataContentEncoding, add data_base64.


n3wscott:
Ref:  https://github.com/cloudevents/spec/pull/492

PR #218, opened 2019-09-30
use []byte and not string for base64 json key.

n3wscott:
Related to https://github.com/cloudevents/sdk-go/issues/195

Leverage the native base64 encoding that the golang json lib uses.


alanconway:
LGTM

Issue #48, opened 2019-03-06
There be a uint16 check here for min and max on port


n3wscott:
Should there be a uint16 check here for min and max?

_Originally posted by @markpeek in https://github.com/_render_node/MDExOlB1bGxSZXF1ZXN0MjU4NzYxMTAx/pull_requests/unread_timeline_

PR #220, opened 2019-09-30
validate port to be in the valid port range.

n3wscott:
validate port to be in the valid port range.

fixes https://github.com/cloudevents/sdk-go/issues/48


alanconway:
LGTM 
That check will happen anyway as soon as you Dial or Listen but no harm to do it early

PR #224, opened 2019-09-30
New value conversion API (replace type Value)

alanconway:
CloudEvents types are represented by these 6 native Go types (or canonical strings):

    bool, int32, string, []byte, *url.URL, time.Time

Simple canonical string conversions following package strconv conventions:

    FormatX(X) string ParseX(string)(X, error)

Where X is one of Bool, Integer, String, Binary, URL, Time:

Flex conversion: accepts any native value or canonical string and converts to native:

    ToX(interface{}) (X, error)

Flex format: accepts any native value and produces the canonical string:

    Format(interface{}) string

Validation: validates and converts any convertible value into one of the preferred native types. This is for internal use and for bindings with their own type-mapping rules; the binding's type switch only need consider the 6 types above, and can ignore conversions.

    Validate(interface{}) (interface{}, error)

### DIFF
--- a/pkg/cloudevents/types/value.go
+++ b/pkg/cloudevents/types/value.go
@@ -10,201 +10,212 @@ import (
 	"time"
 )
 
-// StringOf returns the canonical string encoding of x.
-// x can be any type that is convertible to a CloudEvents attribute.
-func StringOf(x interface{}) (string, error) {
-	v, err := ValueOf(x)
-	return v.String(), err
+// FormatBool returns canonical string format: "true" or "false"
+func FormatBool(v bool) string { return strconv.FormatBool(v) }
+
+// FormatInteger returns canonical string format: decimal notation.
+func FormatInteger(v int32) string { return strconv.Itoa(int(v)) }
+
+// FormatBinary returns canonical string format: standard base64 encoding
+func FormatBinary(v []byte) string { return base64.StdEncoding.EncodeToString(v) }
+
+// FormatTime returns canonical string format: RFC3339 with nanoseconds
+func FormatTime(v time.Time) string { return v.UTC().Format(time.RFC3339Nano) }
+
+// ParseBool parse canonical string format: "true" or "false"
+func ParseBool(v string) (bool, error) { return strconv.ParseBool(v) }
+
+// ParseInteger parse canonical string format: decimal notation.
+func ParseInteger(v string) (int32, error) {
+	// Accept floating-point but truncate to int32 as per CE spec.
+	f, err := strconv.ParseFloat(v, 64)
+	if err != nil {
+		return 0, err
+	}
+	if f > math.MaxInt32 || f < math.MinInt32 {
+		return 0, rangeErr(v)
+	}
+	return int32(f), nil
 }
 
-// Value holds the value of a CloudEvents attribute.
-//
-// Value can hold the native type or the string representation.
-// The To...() methods extract the value as a native type.
-type Value struct{ v interface{} }
+// ParseBinary parse canonical string format: standard base64 encoding
+func ParseBinary(v string) ([]byte, error) { return base64.StdEncoding.DecodeString(v) }
 
-// Interface returns the value as an interface{}.
-//
-// The return value may be nil if the Value is empty, otherwise it will always
-// be one of the native types: bool, int32, string, []byte, url.URL, time.Time.
-func (v Value) Interface() interface{} { return v.v }
+// ParseTime parse canonical string format: RFC3339 with nanoseconds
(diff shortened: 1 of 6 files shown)


pkg/cloudevents/types/value.go
```diff
@@ -1,5 +1,7 @@
 package types
 
+// FIXME(alanconway) naming: URL, Time?
```
n3wscott: I like sticking with the native type names as you have it.

pkg/cloudevents/types/value.go
```diff
@@ -10,201 +12,212 @@ import (
...
-// ToURIRef returns a url.URL value, decoding from string if necessary.
-func (v Value) ToURIRef() (url.URL, error) {
-	switch x := v.v.(type) {
-	case url.URL:
-		return x, nil
+// ToURL returns a *url.URL value, parsing from string if necessary.
+func ToURL(v interface{}) (*url.URL, error) {
+	v, err := Validate(v)
+	switch v := v.(type) {
+	case *url.URL:
```
n3wscott: might be nice to also handle `url.URL`

alanconway:
Left that FIXME by mistake, glad Scott likes the way it is :) Will clean
up.

pkg/cloudevents/types/value.go
```diff
@@ -10,201 +10,212 @@ import (
...
+// ToTime returns a time.Time value, parsing from RFC3339 string if necessary.
+func ToTime(v interface{}) (time.Time, error) {
+	v, err := Validate(v)
+	switch v := v.(type) {
 	case time.Time:
-		return x, nil
+		return v, nil
 	case string:
-		ts, err := time.Parse(time.RFC3339Nano, x)
+		ts, err := time.Parse(time.RFC3339Nano, v)
```
n3wscott: you could types.ParseTimestamp and then return the inner time.Time? Then the CE time format does not have to leak in two places.

pkg/cloudevents/types/value.go
```diff
@@ -10,201 +10,212 @@ import (
...
+		return time.Time{}, convertErr("Time", v, err)
 	}
 }
 
-func rangeErr(v interface{}) error {
-	return fmt.Errorf("%v is out of range for Integer", v)
-}
-
-func typeErr(name string, v interface{}) error {
+func convertErr(name string, v interface{}, err error) error {
```
n3wscott: I like this, I wonder if we need a followup to add the IsConvertErr methods like k8s uses to see the type of error. like NotFoundError

pkg/cloudevents/types/value_test.go
```diff
@@ -14,60 +14,42 @@ import (
...
-	// Integer conversions are range-checked
-	_, err = types.ValueOf(math.MaxUint32)
-	fmt.Println(err)
-
-	// Consistent treatment for native or string-encoded values.
-	asInt := func(x interface{}) {
-		v, _ := types.ValueOf(x)
-		i, err := v.ToInteger()
-		fmt.Printf("asInt: %v (%v)\n", i, err)
+func Example() {
```
n3wscott: NEEEEAAAATTTTT

I am so jazzed about this.

pkg/cloudevents/types/value.go
alanconway: Wanted to bring that up - I don't like a ParseTimestamp that hides parse errors. I'd rather keep ParseTime and drop ParseTimestamp - the marshalling code can do
    ts, _ := ParseTime(s)
if it needs to ignore the error.

pkg/cloudevents/types/value.go
alanconway: We could add a concrete
   type ConvertError string
and return that from convertErr, users can use standard type casts or type switches to detect. I prefer that to an IsBlah function for cases where we can control the error type.
We can do the function as well if you have a strong preference: 
    func IsConvertErr(err e) bool { _, ok := e.(ConvertError); return ok }

pkg/cloudevents/types/value.go
n3wscott: ah interesting. As long as the time format parsing ends up in just one place, the type system seems like a good place.

pkg/cloudevents/types/value.go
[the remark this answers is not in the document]
alanconway: Leftover.

alanconway: +1

pkg/cloudevents/types/value.go
alanconway: It is handled - Validate() does all legal conversions, so the switch only needs to consider *url.URL and string. Aside from actual validation, that's the main role of Validate() - to simplify type switches so they only have to consider the preferred native types.

PR #226, opened 2019-10-01
Package binding/spec: specversion metadata for generic mappings.

### LINKED ISSUE #225: Generic binding interfaces and tools
Separating this from #180 as there is a significant library that is independent of AMQP and can be used to build other bindings.

alanconway:
Provides string attribute names for all versions, useful for mappings
like AMQP or HTTP that use prefixed CE attribute names in their encoding.

Fixes #225

### DIFF
--- a/pkg/binding/doc.go
+++ b/pkg/binding/doc.go
@@ -1,35 +1,36 @@
 /*
 
-Package binding provides interfaces for transport bindings.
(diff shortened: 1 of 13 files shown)


n3wscott:
This all LGTM, just waiting on the merge conflicts to be fixed to merge.

PR #230, opened 2019-10-01
Single point of parsing for Time

alanconway:
All time parsing calls types.ParseTime()

Return errors from time parsing

Added types.ConvertError error type for all conversion errors.

Improved error messages and tightened up some error handling.

### DIFF
--- a/pkg/cloudevents/types/timestamp.go
+++ b/pkg/cloudevents/types/timestamp.go
@@ -16,15 +16,12 @@ type Timestamp struct {
 }
 
 // ParseTimestamp attempts to parse the given time assuming RFC3339 layout
-func ParseTimestamp(t string) *Timestamp {
-	if t == "" {
-		return nil
+func ParseTimestamp(s string) (*Timestamp, error) {
+	if s == "" {
+		return nil, nil
 	}
-	timestamp, err := time.Parse(time.RFC3339Nano, t)
-	if err != nil {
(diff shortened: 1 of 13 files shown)


pkg/cloudevents/types/timestamp.go
```diff
@@ -67,17 +60,10 @@ func (t *Timestamp) UnmarshalXML(d *xml.Decoder, start xml.StartElement) error {
...
 
 // String outputs the time using layout RFC3339.
-func (t *Timestamp) String() string {
-	if t == nil {
-		return time.Time{}.UTC().Format(time.RFC3339Nano)
-	}
-
-	return t.UTC().Format(time.RFC3339Nano)
-}
+func (t Timestamp) String() string { return t.UTC().Format(time.RFC3339Nano) }
```
n3wscott: This should be the canonical form of the time. ToString(t) right?

alanconway: Good point.

alanconway: Fixed, removed all other refs to RFC399 outside value.go, also fixed usage from other packages.

n3wscott:
LGTM!!

Looks like the commit has to be signed?

alanconway:
Dammit, still haven't figured out how to make emacs do that for me. Should
be good now.

PR #235, opened 2019-10-08
Remove datacontent attr

matzew:
updating tests


matzew:
Follow up on #229 and related `datacontent` attribute changes

PR #237, opened 2019-10-10
Add datacodec support for "text/plain" media type.

### LINKED ISSUE #232: Add text/plain as a supported content-type in sdk-go
Currently, text/plain is not a supported content-type in sdk-go.
However, I don't see anything in the http transport binding saying that binary mode events can't be text/plain content-type. Can this be added to sdk-go?

alanconway:
Fixes #232

### DIFF
--- a/pkg/cloudevents/datacodec/codec_test.go
+++ b/pkg/cloudevents/datacodec/codec_test.go
@@ -11,6 +11,8 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
+func strptr(s string) *string { return &s }
+
 type Example struct {
 	Sequence int    `json:"id"`
 	Message  string `json:"message"`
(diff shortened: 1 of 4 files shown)


pkg/cloudevents/datacodec/codec_test.go
```diff
@@ -30,6 +32,11 @@ func TestCodecDecode(t *testing.T) {
 			wantErr:     `[decode] unsupported content type: "unit/testing-invalid"`,
 		},
 
+		"text/plain": {
+			contentType: "text/plain",
+			in:          "hello😀",
```
matzew: no cat emoji ? :cat:

alanconway: Holy moly, there's a whole library of unicode cats:
https://emojipedia.org/cat-face-with-wry-smile/
Somebody has waaay too much time on their hands.

PR #247, opened 2019-10-17
Allow setting Host in HTTP request

agrimmer:
Allow setting the Host in the HTTP request by using the request contained in transport.


agrimmer:
Note, setting the Host in the header is not sufficient, i.e. https://github.com/golang/go/issues/7682

n3wscott:
LGTM

PR #248, opened 2019-10-18
Added Requester and async ack support.

### LINKED ISSUE #227: Extend binding Sender to support request-response and async ack.
Something like:
```
Requester {
    Request(request Message) Receiver
}
// Sync use:
resp, err := requester.Request(req).Receive()
// Async use
rec, err := requester.Request(req)
... later
resp, err := rec.Receive()
```
Async sender will modify Sender inteface, concrete proposal coming in pull request.

alanconway:
Requester supports request/response pattern.

Async ack is provided by WithFinish() - Sender does async send by
default (depends on QoS and binding), WithFinish() allows the caller
to be notified when Finish() is called.

Also added ChanSender/ChanReceiver - useful for testing so might as well make
them exported.

Fixes #227

### DIFF
--- a/pkg/binding/chan.go
+++ b/pkg/binding/chan.go
@@ -1,0 +1,52 @@
+package binding
+
+import (
+	"context"
+	"errors"
+	"io"
+)
+
+// ChanSender implements Sender by sending on a channel.
+type ChanSender chan<- Message
+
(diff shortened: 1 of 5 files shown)


n3wscott:
LGTM

PR #259, opened 2019-11-25
Benchmark to compare transport/http to bindings/http

slinkydeveloper:
This benchmark emulates a typical message forwarding situation and doesn't pass through the network (the http client is mocked for this purpose), http methods are called directly and mocked.

This could be used as a comparison tool to improve further both transport/http & bindings/http

I've also included an example gnuplot script to graph the results


slinkydeveloper:
cc @alanconway

n3wscott:
LGTM

I look forward to using this :D

PR #260, opened 2019-11-26
Adding client defaulter for datacontenttype = application/json.

n3wscott:
It has become a problem to not have app/json set for most users. Adding a defaulter helper and also adding it to the default client.

### DIFF
--- a/pkg/cloudevents/client/defaulters_test.go
+++ b/pkg/cloudevents/client/defaulters_test.go
@@ -9,8 +9,10 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
+var versions = []string{"0.1", "0.2", "0.3", "1.0"}
+
 func TestDefaultIDToUUIDIfNotSet_empty(t *testing.T) {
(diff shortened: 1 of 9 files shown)


pkg/cloudevents/client/defaulters_test.go
```diff
@@ -6,60 +6,44 @@ import (
...
-		},
-	}
-	for n, tc := range testCases {
-		t.Run(n, func(t *testing.T) {
-
-			got := DefaultIDToUUIDIfNotSet(context.TODO(), tc.event)
-
-			if got.Context != nil && got.Context.AsV02().ID == "" {
+func TestDefaultIDToUUIDIfNotSet_empty(t *testing.T) {
+	for _, tc := range []string{"0.1", "0.2", "0.3", "1.0"} {
```
vaikas: would it be useful to hoist these into a var:
var versions = []string{"0.1", "0.2", "0.3", "1.0"}

Then in these loops use:
for _, version := range versions {

n3wscott:
Thanks Nacho!

Issue #262, opened 2019-11-27
binding to transfer binary events without allocating an intermediate 


alanconway:
The current binding.Message.Event() API creates an intermediate Event object which then must be encoded to be sent by a Sender. Provide an additional API that allows binary events to be transferred without an intermediate Event data structure (analogous to the way Message.Structured() avoids intermediate data structures for structured messages.)

Idea provided by @slinkydeveloper 
Initial proposal #263

alanconway:
Proposal in #263

slinkydeveloper:
I'm on it

PR #263, opened 2019-11-27
[WIP] Binary message transfer without intermediate Event.

alanconway:
Non-functioning code! This is an API proposal for #262

NOTE: the proposal is incompatible with current API (adds a method to Message)

If it needs to be made compatible, the Binary() method can be moved to a new
interface (e.g. BinaryMessage) and the user can cast.

### DIFF
--- a/pkg/binding/interfaces.go
+++ b/pkg/binding/interfaces.go
@@ -2,7 +2,10 @@ package binding
 
 import (
 	"context"
+	"errors"
+	"io"
 
+	"github.com/cloudevents/sdk-go/pkg/binding/spec"
 	ce "github.com/cloudevents/sdk-go/pkg/cloudevents"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/binding/interfaces.go
```diff
@@ -49,6 +62,27 @@ type Message interface {
...
+// EventBuilder receives values to build a binary event representation.
+type EventBuilder interface {
+	// Data receives an io.Reader for the data attribute.
+	Data(io.Reader)
+
+	// Set a standard attribute.
+	//
+	// The value can either be the correct type for the attribute, or a canonical
+	// string encoding. See package cloudevents/types
+	Set(attributeId spec.Kind, value interface{})
```
n3wscott: The thing I do not like about this is the builder interface does not get to expose the correct types in the api. It is a hidden detail that gets smashed up in a switch as internal details. So you get runtime issues rather that compile errors.

n3wscott: This is why I have `SetTime` and `Time` etc on `Event`.

alanconway: We could remove Set() from builder and embed EventContextWriter instead for standard attributes. The idea of using interface{} was to avoid parsing and re-encoding canonical string values if both sides use that format, but its probably not a big enough deal to compromise type safety. I'm fine with using EventContextWriter.

n3wscott:

PR #266, opened 2019-11-27
Introducing file based Client Overrides

n3wscott:
### WithOverrides

`WithOverrides` allows you to create a set of files on the filesystem that will be watched, read, and mutate the outbound event.

File names will be used as the extension attribute name (only extensions are allowed, no first-class attributes name).
File contents will be used as the value of the extension attribute.

For example:

```bash
$ tree cmd/samples/http/overrides/extensions/
cmd/samples/http/overrides/extensions/
├── baz
└── foo

$ cat cmd/samples/http/overrides/extensions/baz
bar

$ cat cmd/samples/http/overrides/extensions/foo
42
```

`baz` and `foo` will be added to the CloudEvents extensions  with "bar" and "42", respectively.

See the [overrides](./cmd/samples/http/overrides) sample for a working example.


n3wscott:
I want to revert the k8s thing and add symlink support. Pending...

alanconway:
What's the motivation? At first glance it seems like it's outside the remit of a CloudEvents library to  populate the events from external data sources. Danger of scope creep?

n3wscott:
> What's the motivation? At first glance it seems like it's outside the remit of a CloudEvents library to populate the events from external data sources. Danger of scope creep?

It is slight scope creep, but it is tied into the client section (and totally optional to use) and it is useful enough that I think it can be a general client extension for this lib. The motivation is static decoration is fairly common in client usage, this makes that easier to do for those clients that are using a file to decorate the outbound event.

n3wscott:
Ready.

ti:
@alanconway 
why BinaryV1 as V03()?

case BinaryV1, StructuredV1, BatchedV1:
		ca := event.Context.AsV1()

PR #267, opened 2019-12-03
fix version convert bug

ti:


ti:
@alanconway
why BinaryV1 as V03()?

case BinaryV1, StructuredV1, BatchedV1:
ca := event.Context.AsV1()

n3wscott:
Thanks! Looks like you need to sign your commit to merge.

n3wscott:
Ping?

Issue #268, opened 2019-12-05
Improve NATS transport


ripienaar:
Today the NATS transport is very limited - no ability to use TLS, credentials, users, passwords etc. There's a whole pile of features one might want in real world.

So rather than try to map all the many nats options 1:1 here, would you object to me adding a option to the transport where a fully configured NATS connection can be passed in as an option? This way users can handle all the setup, authentication, tls etc on their own and pass in a suitabily prepared connector?

We can add commonly used options into the transport but honestly it'll be forever going round in circles that way, passing in a connection is path of least resistance and something people do elsewhere in the NATS eco system

ripienaar:
The proposed change is in a WIP PR https://github.com/cloudevents/sdk-go/pull/269/files

ripienaar:
I also added a little helper so one can pass in any nats option in stead of the extra work to pre-make a nats connection.

ripienaar:
Have simplified this down to just taking nats options and not a prepared connection.

ripienaar:
this should tick off https://github.com/cloudevents/sdk-go/issues/62

n3wscott:
Is this closable now?

ripienaar:
yup@

PR #269, opened 2019-12-05
Improve NATS transport flexibility (#268)

ripienaar:
This allow users to prepare a NATS connection and supply it to the
transport, this allow the plethora of TLS, Authentication, Credentials
and Logging options to be handled without overly expanding this library


pkg/cloudevents/transport/nats/transport.go
```diff
@@ -37,15 +38,18 @@ type Transport struct {
 
 // New creates a new NATS transport.
 func New(natsServer, subject string, opts ...Option) (*Transport, error) {
```
derekcollison: Might make natsServer -> natsURL?

pkg/cloudevents/transport/nats/options.go
```diff
@@ -10,3 +14,15 @@ func WithEncoding(encoding Encoding) Option {
 		return nil
 	}
 }
+
+// WithConnOptions supplies NATS connection options that will be used when setting
+// up the internal NATS connection
+func WithConnOptions(opts ...nats.Option) Option {
```
derekcollison: Just WithOptions imo.

pkg/cloudevents/transport/nats/transport.go
```diff
@@ -19,11 +19,12 @@ const (
...
 
-// Transport acts as both a http client and a http handler.
+// Transport acts as both a NATS client and a NATS handler.
 type Transport struct {
-	Encoding Encoding
-	Conn     *nats.Conn
-	Subject  string
+	Encoding    Encoding
+	Conn        *nats.Conn
+	ConnOptions []nats.Option
```
derekcollison: Just Options imo.

pkg/cloudevents/transport/nats/transport.go
```diff
@@ -37,15 +38,18 @@ type Transport struct {
 
 // New creates a new NATS transport.
 func New(natsServer, subject string, opts ...Option) (*Transport, error) {
-	conn, err := nats.Connect(natsServer)
+	t := &Transport{
+		Subject:     subject,
+		ConnOptions: []nats.Option{},
```
derekcollison: Just Options vs ConnOptions.

pkg/cloudevents/transport/nats/options.go
ripienaar: There are Nats related options and transport related options like the encoding so dropping the `Conn` might be confusing?

derekcollison: But these are specific to NATS yes?

ripienaar: no, in the same set of options we have `WithEncoding` which is a transport thing so `WithOptions` would be confusing, `WithConnOptions` makes it clear the ones passed there related to nats when scanning docs? and more might come in time thats unrelated to nats - since ultimately this config section configures the transport first

derekcollison: ok fair.

pkg/cloudevents/transport/nats/transport.go
```diff
@@ -36,16 +37,19 @@ type Transport struct {
 }
 
 // New creates a new NATS transport.
-func New(natsServer, subject string, opts ...Option) (*Transport, error) {
-	conn, err := nats.Connect(natsServer)
+func New(natsURL, subject string, opts ...Option) (*Transport, error) {
+	t := &Transport{
+		Subject:     subject,
+		ConnOptions: []nats.Option{},
```
derekcollison: Leave this nil.

pkg/cloudevents/transport/nats/options.go
```diff
@@ -10,3 +14,15 @@ func WithEncoding(encoding Encoding) Option {
...
+	return func(t *Transport) error {
+		for _, o := range opts {
```
derekcollison: t.ConnOptions = append(t.ConnOptions, opts...)

t.ConnOptions can be nil too.

n3wscott:
AWESOME!!

LGTM, I will wait to see if you want to address https://github.com/cloudevents/sdk-go/pull/269#discussion_r354917108 before merging.

ripienaar:
I'm good, personally I prefer setting things to what they claim to be so happy to accept Dereks approve as is :P

PR #274, opened 2019-12-17
Fix extensions encoding

### LINKED ISSUE #273: Extensions marshal to JSON broken
Blocking #270

When using `event.SetExtension` the marshal to JSON of an event including extensions with `URL`. E.g.:

```json
{"data":"hello","datacontenttype":"text/json","dataschema":"http://example.com/schema","exbinary":"AAECAw==","exbool":true,"exint":42,"exstring":"exstring","extime":"2020-03-21T12:34:56.78Z","exurl":{"Scheme":"http","Opaque":"","User":null,"Host":"example.com","Path":"/schema","RawPath":"","ForceQuery":false,"RawQuery":"","Fragment":""},"id":"full-event-0","source":"http://example.com/source","specversion":"1.0","subject":"topic","time":"2020-03-21T12:34:56.78Z","type":"com.example.FullEvent"}
```

This is because `SetExtension` populates the underlying map not with wrapped types (provided by `types` module), but with go types. `net.URL` doesn't provide a meaningful marshal/unmarshal, which leads to this error.

Tests inside `event_marshal_test.go` are broken because they set correct values to the map directly, and not using `SetExtension` method.

## Reproducer

```go
func TestMarshalv1Ext(t *testing.T) {
	now := types.Timestamp{Time: time.Now().UTC()}
	sourceUrl, _ := url.Parse("http://example.com/source")
	source := &types.URIRef{URL: *sourceUrl}

	schemaUrl, _ := url.Parse("http://example.com/schema")
	schema := &types.URI{URL: *schemaUrl}

	testEvent := cloudevents.Event{
		Context: cloudevents.EventContextV1{
			Type:            "com.example.test",
			Source:          *source,
			DataSchema:      schema,
			ID:              "ABC-123",
			Time:            &now,
			DataContentType: cloudevents.StringOfApplicationJSON(),
		}.AsV1(),
		Data: DataExample{
			AnInt:   42,
			AString: "testing",
		},
	}

	testEvent.SetExtension("ex1", int32(42))
	testEvent.SetExtension("ex2", "testing")
	testEvent.SetExtension("ex4", schema)

		event   cloudevents.Event
		want    []byte
		wantErr *string
	}{
		"v1.0 cased extensions": {
			event: testEvent,
			want: toBytes(map[string]interface{}{
				"specversion":     "1.0",
				"datacontenttype": "application/json",
				"data": map[string]interface{}{
					"a": 42,
					"b": "testing",
				},
				"id":   "ABC-123",
				"time": now.Format(time.RFC3339Nano),
				"type": "com.example.test",
				"ex1":  42,
				"ex2":  "testing",
				"ex4": "http://example.com/schema",
				"dataschema": "http://example.com/schema",
				"source":     "http://example.com/source",
			}),
		},
	}
	for n, tc := range testCases {
		t.Run(n, func(t *testing.T) {

			gotBytes, err := json.Marshal(tc.event)

			if tc.wantErr != nil || err != nil {
				if diff := cmp.Diff(*tc.wantErr, err.Error()); diff != "" {
					t.Errorf("unexpected error (-want, +got) = %v", diff)
				}
				return
			}

			// so we can understand the diff, turn bytes to strings
			want := string(tc.want)
			got := string(gotBytes)

			if diff := cmp.Diff(want, got); diff != "" {
				t.Errorf("unexpected event (-want, +got) = %v", diff)
			}
		})
	}
}
```

## Proposed actions

- `Extension` map should be populated with wrapped types, so they have the proper marshal/unmarshal function
- Tests inside `event_marshal_test.go` should use `SetExtension`

Bonus:

- make `Extension` map private, since there are accessors get, set, put

slinkydeveloper:
Fixes #273 

## Breaking changes

* Now extension map doesn't contain (and should not contain anymore) `net.URL` & `time.Time` (and pointers). **Users that wants to modify this map manually should keep in mind these changes**
* Unmarshalled extension types from json are now coherent between different spec versions (they are always unmarshalled as go types)

### DIFF
--- a/pkg/cloudevents/event_marshal.go
+++ b/pkg/cloudevents/event_marshal.go
@@ -9,6 +9,8 @@ import (
 	"strconv"
 	"strings"
 
+	errors2 "github.com/pkg/errors"
+
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/observability"
 )
 
@@ -222,11 +224,17 @@ func (e *Event) JsonDecodeV02(body []byte, raw map[string]json.RawMessage) error
 
 	if len(raw) > 0 {
 		extensions := make(map[string]interface{}, len(raw))
+		ec.Extensions = extensions
 		for k, v := range raw {
 			k = strings.ToLower(k)
-			extensions[k] = v
+			var tmp interface{}
+			if err := json.Unmarshal(v, &tmp); err != nil {
+				return err
+			}
+			if err := ec.SetExtension(k, tmp); err != nil {
+				return errors2.Wrap(err, "Cannot set extension with key "+k)
+			}
 		}
-		ec.Extensions = extensions
 	}
 
 	e.Context = &ec
@@ -263,11 +271,17 @@ func (e *Event) JsonDecodeV03(body []byte, raw map[string]json.RawMessage) error
 
 	if len(raw) > 0 {
 		extensions := make(map[string]interface{}, len(raw))
+		ec.Extensions = extensions
 		for k, v := range raw {
 			k = strings.ToLower(k)
-			extensions[k] = v
+			var tmp interface{}
+			if err := json.Unmarshal(v, &tmp); err != nil {
+				return err
+			}
+			if err := ec.SetExtension(k, tmp); err != nil {
+				return errors2.Wrap(err, "Cannot set extension with key "+k)
+			}
 		}
-		ec.Extensions = extensions
 	}
 
 	e.Context = &ec
@@ -312,15 +326,17 @@ func (e *Event) JsonDecodeV1(body []byte, raw map[string]json.RawMessage) error 
 
 	if len(raw) > 0 {
 		extensions := make(map[string]interface{}, len(raw))
+		ec.Extensions = extensions
 		for k, v := range raw {
 			k = strings.ToLower(k)
-			var tmp string
+			var tmp interface{}
 			if err := json.Unmarshal(v, &tmp); err != nil {
 				return err
 			}
-			extensions[k] = tmp
+			if err := ec.SetExtension(k, tmp); err != nil {
+				return errors2.Wrap(err, "Cannot set extension with key "+k)
+			}
 		}
-		ec.Extensions = extensions
 	}
 
 	e.Context = &ec
--- a/pkg/cloudevents/extensions.go
+++ b/pkg/cloudevents/extensions.go
@@ -27,4 +27,4 @@ func caseInsensitiveSearch(key string, space map[string]interface{}) (interface{
 	return nil, false
 }
 
-var IsAlphaNumericLowercaseLetters = regexp.MustCompile(`^[a-z0-9]+$`).MatchString
+var IsAlphaNumeric = regexp.MustCompile(`^[a-zA-Z0-9]+$`).MatchString
--- a/pkg/cloudevents/types/value.go
+++ b/pkg/cloudevents/types/value.go
@@ -68,23 +68,23 @@ func Format(v interface{}) (string, error) {
 		return v, nil
 	case []byte:
 		return FormatBinary(v), nil
-	case url.URL:
+	case URI:
 		return v.String(), nil
-	case *url.URL:
+	case URIRef:
 		// url.URL is often passed by pointer so allow both
 		return v.String(), nil
-	case time.Time:
-		return FormatTime(v), nil
+	case Timestamp:
+		return FormatTime(v.Time), nil
 	default:
 		return "", fmt.Errorf("%T is not a CloudEvents type", v)
 	}
 }
 
 // Validate v is a valid CloudEvents attribute value, convert it to one of:
-//     bool, int32, string, []byte, *url.URL, time.Time
+//     bool, int32, string, []byte, types.URI, types.URIRef, types.Timestamp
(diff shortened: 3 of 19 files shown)


slinkydeveloper:
cc @alanconway @n3wscott

pkg/binding/spec/spec.go
```diff
@@ -108,7 +108,7 @@ func (v *version) SetAttribute(c ce.EventContextWriter, name string, value inter
 	name = strings.ToLower(name)
 	var err error
 	if strings.HasPrefix(name, v.prefix) { // Extension attribute
-		value, err = types.Validate(value)
+		value, err = types.ValidateAndWrap(value)
```
alanconway: I would be inclined to just change the Validate() function rather than introduce a new function. The purpose of Validate() is to convert any "convertible" value into the "standard" Go form. As you've pointed out the "standard" form form URL and Time values should be the type wrappers since the native types lack proper marshalling - so you're fixing a bug in Validate(), rather than adding a new feature. I'd update the doc comments in the types package to reflect that.

pkg/cloudevents/event_marshal.go
```diff
@@ -224,7 +224,11 @@ func (e *Event) JsonDecodeV02(body []byte, raw map[string]json.RawMessage) error
 		extensions := make(map[string]interface{}, len(raw))
 		for k, v := range raw {
 			k = strings.ToLower(k)
-			extensions[k] = v
+			var tmp interface{}
+			if err := json.Unmarshal(v, &tmp); err != nil {
+				return err
+			}
+			extensions[k] = tmp
```
alanconway: Should be ec.SetExtension(k, tmp) rather than direct map?

pkg/cloudevents/extensions.go
```diff
@@ -27,4 +27,4 @@ func caseInsensitiveSearch(key string, space map[string]interface{}) (interface{
 	return nil, false
 }
 
-var IsAlphaNumericLowercaseLetters = regexp.MustCompile(`^[a-z0-9]+$`).MatchString
+var IsAlphaNumericLowercaseLetters = regexp.MustCompile(`^[a-zA-Z0-9]+$`).MatchString
```
alanconway: Rename: IsAlphaNumeric.

pkg/cloudevents/types/value.go
```diff
@@ -131,6 +131,63 @@ func Validate(v interface{}) (interface{}, error) {
 	return nil, fmt.Errorf("invalid CloudEvents value: %#v", v)
 }
 
+// Validate v is a valid CloudEvents attribute value, convert it to one of:
+//     bool, int32, string, []byte, types.URI, types.URIRef, types.Timestamp
+func ValidateAndWrap(v interface{}) (interface{}, error) {
```
alanconway: See comment above - replace Validate() with this rather than adding a new function.

pkg/binding/spec/spec.go
slinkydeveloper: My bad, I thought `types.Validate` was used somewhere else

alanconway:
Nope, I introduced it exactly for handling extension attributes. Thanks for
fixing it up!

slinkydeveloper:
Any idea why the ci fails? I've tried to run go fmt but there are no changes

alanconway:
> Any idea why the ci fails? I've tried to run go fmt but there are no changes

Nope, go fmt passed for me too.  You might be able to fix the CI script so it shows the output by hacking on .circleci/config.yml

slinkydeveloper:
The problem was inside the CI script. To check if code is formatted, there is a check on the output of `go fmt`, if there isn't any, then the step succeds. The problem is that `go fmt` could potentially print that is solving dependencies, like:

```
#!/bin/bash -eo pipefail
go fmt ./...

go: finding github.com/pkg/errors v0.8.1
```

Disabling go modules while running `go fmt` does the trick.

BTW I haven't added any new dependency

pkg/cloudevents/extensions.go
```diff
@@ -27,4 +27,4 @@ func caseInsensitiveSearch(key string, space map[string]interface{}) (interface{
 	return nil, false
 }
 
-var IsAlphaNumericLowercaseLetters = regexp.MustCompile(`^[a-z0-9]+$`).MatchString
+var IsAlphaNumeric = regexp.MustCompile(`^[a-zA-Z0-9]+$`).MatchString
```
n3wscott: the spec says only lowercase are supported.

n3wscott:
```markdown
CloudEvents attribute names MUST consist of lower-case letters ('a' to 'z') or digits ('0' to '9') from the ASCII character set. Attribute names SHOULD be descriptive and terse and SHOULD NOT exceed 20 characters in length.
```

https://github.com/cloudevents/spec/blob/master/spec.md#attribute-naming-convention

slinkydeveloper: Yes but `setExtension` make it lowercase https://github.com/cloudevents/sdk-go/blob/2676232a051b22008883dc9b95906842f017398b/pkg/cloudevents/eventcontext_v1.go#L82

slinkydeveloper: And this also makes the behaviour consistent with other context implementations where this is not required

n3wscott:
LGTM!!

PR #278, opened 2019-12-30
close gorouting when context cancel

qianlnk:
startreceiver with context cancel

when I cancel, I find the go routing func not exit, and it always get error "long poll request returned error".

### DIFF
--- a/pkg/cloudevents/transport/http/transport.go
+++ b/pkg/cloudevents/transport/http/transport.go
@@ -395,14 +395,12 @@ func (t *Transport) longPollStart(ctx context.Context) error {
 				logger.Errorw("long poll request returned error", err)
 				uErr := err.(*url.Error)
 				if uErr.Temporary() || uErr.Timeout() {
-					time.Sleep(time.Second * 2)
 					continue
 				}
 				// TODO: if the transport is throwing errors, we might want to try again. Maybe with a back-off sleep.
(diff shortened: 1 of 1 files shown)


pkg/cloudevents/transport/http/transport.go
```diff
@@ -383,19 +383,26 @@ func (t *Transport) longPollStart(ctx context.Context) error {
...
 		for {
+			if isClosed {
+				return
+			}
+
 			if resp, err := t.LongPollClient.Do(req); err != nil {
 				logger.Errorw("long poll request returned error", err)
 				uErr := err.(*url.Error)
 				if uErr.Temporary() || uErr.Timeout() {
+					time.Sleep(time.Second * 2)
```
n3wscott: not sure why you would want to add sleep statements here. Is this for debugging and forgotten?

qianlnk: when get error here， it retry "DO" too many times. then you will get too many error log, so I think here should retry after some time.

n3wscott: We can't add seep commands to the codebase like this.

qianlnk: OK, I've deleted the sleep code.

n3wscott:
LGTM

PR #279, opened 2020-01-08
Optimize pubsub receiver and add unit tests

yolocs:
* Allow setting `pubsub.ReceiveSettings`
* Default to 1000 for `NumGoroutine`
* Add unit tests for internal.Connection


yolocs:
Test failures don't seem related to this change.

go.mod
```diff
@@ -20,6 +20,8 @@ require (
 	go.uber.org/multierr v1.1.0 // indirect
 	go.uber.org/zap v1.10.0
 	golang.org/x/sync v0.0.0-20190423024810-112230192c58
+	google.golang.org/api v0.15.0
+	google.golang.org/grpc v1.26.0
```
nachocano: was this added because of the UT?

pkg/cloudevents/transport/pubsub/internal/connection.go
```diff
@@ -36,6 +36,8 @@ type Connection struct {
 	subWasCreated  bool
 	subOnce        sync.Once
 
+	ReceiveSettings *pubsub.ReceiveSettings
```
nachocano: add a comment?

nachocano: will these be configurable when setting up the transport?

pkg/cloudevents/transport/pubsub/internal/connection.go
```diff
@@ -49,6 +51,15 @@ const (
 	DefaultRetentionDuration = 25 * time.Hour
 )
 
+var DefaultReceiveSettings = pubsub.ReceiveSettings{
+	// Pubsub default receive settings will fill in other values.
+	// https://godoc.org/cloud.google.com/go/pubsub#Client.Subscription
+
+	// Override the default number of goroutines.
+	NumGoroutines: 1000,
```
nachocano: maybe add a comment of why 1k. Say it's a magic number, but has shown performance improvements empirically, and such...

nachocano:
Thanks for adding this @yolocs !! 
Just some nits... 
I'm curious about how this is configured from the calling code. Should we add it to the transport struct as well? https://github.com/cloudevents/sdk-go/blob/620617c81ca710e4ea931bb06036947603524674/pkg/cloudevents/transport/pubsub/transport.go#L32....

go.mod
yolocs: Yes.

pkg/cloudevents/transport/pubsub/internal/connection.go
yolocs: Missed that. Added now.

pkg/cloudevents/transport/pubsub/internal/connection.go
yolocs: Done

pkg/cloudevents/transport/pubsub/transport.go
```diff
@@ -42,6 +43,9 @@ type Transport struct {
 
 	// PubSub
 
+	// ReceiveSettings is used to configure Pubsub pull subscription.
+	ReceiveSettings *pubsub.ReceiveSettings
```
nachocano: I think you still need a function that returns an Option to set this up. Similar to how the other fields are set?

yolocs: Didn't realize there is an option. Done.

nachocano:
looks good to me! 
@n3wscott would have to stamp it

Issue #281, opened 2020-01-14
Allow Message instance to be reused across several Senders


slinkydeveloper:
Starting from https://github.com/cloudevents/sdk-go/pull/270, after a message is sent using a `Sender`, the `Message` is left in an inconsistent state and can't be reused to be forwarded again to another `Sender`.

Investigate how the `Message` instance can be forwarded to more `Sender`s

slinkydeveloper:
Solved by https://github.com/cloudevents/sdk-go/pull/292

PR #270, opened 2019-12-05
Bindings optimization to support binary -> binary + some refactoring

slinkydeveloper:
Fixes  #262
Depends on #274

## Overview

Starting from @alanconway work (https://github.com/cloudevents/sdk-go/pull/263) I've reorganized a bit the `Message` interface to allow both binary -> binary message passing and structured -> structured. 

The `Encoder` pattern allows to the receiver message to extract informations and to the sender to encode these informations in the right request structure.

I modified both HTTP and AMQP binding implementations to support new interfaces.

## Benchmark

From https://github.com/cloudevents/sdk-go/tree/master/test/benchmark/http

8Kb payload (left is https://github.com/cloudevents/sdk-go/pull/274, right is this patch)
8k

16Kb payload (left is https://github.com/cloudevents/sdk-go/pull/274, right is this patch)
16k

32Kb payload (left is https://github.com/cloudevents/sdk-go/pull/274, right is this patch)
32k

PR #282, opened 2020-01-15
Enable a binding.Message to be consumed more times

### LINKED ISSUE #280: Memory pools in bindings
Investigate the usage of memory pools in bindings to buffer event payloads

slinkydeveloper:
Fix #280 #281

## Reading binding.Message several times

With https://github.com/cloudevents/sdk-go/pull/270, every binding.Message can be consumed only one time, meaning that after the message is visited one time with the various `Encoder`s, it can't be visited again.

This PR allows a Message to be consumed several times. It re-enforces and document that behaviour, implementing it in `bindings/http.Message`, buffering the request body. To efficiently implement the buffering phase [it uses memory pooling](https://github.com/valyala/bytebufferpool). This obviously introduces a performance regression but allows retries and broadcasting the message.

Master 8Kb & 32Kb:

This PR **without** Memory pooling 8Kb & 32Kb:

This PR **with** memory pooling 8Kb & 32Kb:

## Let more `binding.Sender` to process the message

By contract `Sender.Finish()` **must** invoke `Message.Finish(error)`. To allow different `Sender`s to process the `Message` without prematurely _finishing_ it, this PR includes `WithAcksBeforeFinish()` method to decorate the `Message` instance to invoke `Finish` only after N acks are invoked. (Note: only the last error is take in account for calling the final `Finish`)

## Extra

This PR also adds two scripts to simplify benchmark execution across different git revisions.

### DIFF
--- a/pkg/binding/example_implementing_test.go
+++ b/pkg/binding/example_implementing_test.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"encoding/json"
 	"io"
-	"io/ioutil"
 
 	"github.com/cloudevents/sdk-go/pkg/binding"
 	"github.com/cloudevents/sdk-go/pkg/binding/format"
@@ -30,7 +29,7 @@ import (
 type ExMessage json.RawMessage
 
 func (m ExMessage) Structured(b binding.StructuredEncoder) error {
-	return b.SetStructuredEvent(format.JSON, bytes.NewReader([]byte(m)))
+	return b.SetStructuredEvent(format.JSON, m)
 }
 
 func (m ExMessage) Binary(binding.BinaryEncoder) error {
@@ -46,9 +45,22 @@ func (m ExMessage) Event(b binding.EventEncoder) error {
 	return b.SetEvent(e)
 }
 
+func (m ExMessage) IsEmpty() bool {
+	return m == nil
+}
+
+func (m ExMessage) Bytes() []byte {
+	return []byte(m)
+}
+
+func (m ExMessage) Reader() io.Reader {
+	return bytes.NewReader(m.Bytes())
+}
+
 func (m ExMessage) Finish(error) error { return nil }
 
 var _ binding.Message = (*ExMessage)(nil)
+var _ binding.MessagePayloadReader = (*ExMessage)(nil)
 
 // ExSender sends by writing JSON encoded events to an io.Writer
 // ExSender supports transcoding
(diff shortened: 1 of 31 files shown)


alanconway:
See #283 for a sketch of my alternative proposal. Building this feature into the HTTP binding doesn't help with other bindings, the work has to be repeated. My proposal is to have in-memory message implementations (EventMessage and StructMessage) The benefits as I see it:
1. Don't repeat this work in every binding - a central solution can work with any NxM binding combination.
2. Separates lifecycle - you can keep re-trying to send an in-memory message after the original HTTP or whatever connection has died and cleaned up resources.
3. Not just multicast: you can store in-memory messages in queues, topics, maps, tables - you can keep them for as long as you want for retries and reliability etc. etc.
4. Consistency: binding messages and in-memory messages implement the same Message and Encoder interfaces so anything can talk to anything - fast, direct forwarding between a pair of transports with no intermediate copies, or via intermediate queue/topic/multicast data structures - you can connect anything to anything and still keep copying to the minimum required to solve the problem.

pkg/binding/example_implementing_test.go
```diff
@@ -29,7 +29,7 @@ import (
 type ExMessage json.RawMessage
 
 func (m ExMessage) Structured(b binding.StructuredEncoder) error {
-	return b.SetStructuredEvent(format.JSON, m)
+	return b.SetStructuredEvent(format.JSON, &m)
```
alanconway: Change to use *m is not necessary. json.RawMessage is just a type cast for []byte, and in Go  slices are semantically pointers (actually pointer pairs) - passing or assigning a slice does not copy the underlying data. You need to use the copy() or append() built-in functions to copy the data. So passing/assigning ExMessage by value doesn't copy any data.

pkg/binding/example_implementing_test.go
```diff
@@ -46,9 +45,22 @@ func (m ExMessage) Event(b binding.EventEncoder) error {
 	return b.SetEvent(e)
 }
 
+func (m *ExMessage) IsEmpty() bool {
+	return m == nil
```
alanconway: If you get rid of  the pointer (m ExMessage) then this is (len(m)  == 0) if you keep the pointer then it should be (m == nil || len(*m) == 0)

pkg/binding/interfaces.go
```diff
@@ -84,13 +86,26 @@ var ErrNotStructured = errors.New("message is not in structured mode")
 // ErrNotBinary returned by Message.Binary for non-binary messages.
 var ErrNotBinary = errors.New("message is not in binary mode")
 
+// MessagePayload allows to read a message payload or as a byte array or as a reader
+// Message implementers must re
+type MessagePayloadReader interface {
```
alanconway: I think this is an undue burden on the Message implementer. If they provide an io.Reader then io.Copy to a bytes.Buffer is maximally efficient, and the streaming case is supported. io.Reader is a standard Go idiom (net.http and many other standard libs) so I don't think we should hide it in a non-standard interface.

n3wscott:

PR #283, opened 2020-01-15
WIP: Copying messages for multicast

alanconway:
Copy messages for multicast

StructMessage and EventMessage are intended to be full implementations of the
Message interface that are independent of any transport and can be used to hold
a copy of a message in memory independently of the life of an incoming transport
message (e.g. for multicast, queuing etc.) CopyMessage copies an incoming
Message, the copy can be read many times and is independent of the original
message lifecycle (e.g. you might Finish() the original before you're done
multicasting the copy depending on policy)

They are a lot like your MockXMessages but they are not mocks, they're complete
Message implementations. Possibly there's some code overlap that can be cleaned up.

Please review the code carefully. I like what you've done to reduce copying but
there was a lot to take in, so this is a bit rushed.

One question: I intended for 2 types of message representation - Binary and
Structured. Message.Event() and EventMessage were really just implementations of
Binary encoding - not meant to be a third representation. I think you could drop
Message.Event() now that you have ToEvent(Message) so bindings only have to
implement the methods Structured() and Binary().

### DIFF
--- a/pkg/binding/copy_message.go
+++ b/pkg/binding/copy_message.go
@@ -1,0 +1,26 @@
+package binding
+
+// CopyMessage reads m once and creates an in-memory copy using EventMessage or
+// StructMessage depending on the encoding of m.  The returned copy is not
+// dependent on any transport and can be read many times.
+//
+// NOTE: Calling Finish() on the returned message calls m.Finish()
+func CopyMessage(m Message) (Message, error) {
+	// Try structured first, it's cheaper.
+	sm := StructMessage{}
+	err := m.Structured(&sm)
+	switch err {
+	case nil:
+		return &sm, nil
+	case ErrNotStructured:
+		break
+	default:
+		return nil, err
+	}
+	em := EventMessage{}
+	err = m.Event(&em)
+	if err != nil {
+		return nil, err
+	}
+	return WithFinish(em, func(err error) { _ = m.Finish(err) }), nil
+}
--- a/pkg/binding/format/format.go
+++ b/pkg/binding/format/format.go
@@ -24,6 +24,14 @@ type Format interface {
 	Unmarshal([]byte, *ce.Event) error
 }
 
+// UnknownFormat allows an event with an unknown format string to be forwarded,
+// but Marshal() and Unmarshal will always fail.
+type UnknownFormat string
+
+func (uf UnknownFormat) MediaType() string                 { return string(uf) }
+func (uf UnknownFormat) Marshal(ce.Event) ([]byte, error)  { return nil, unknown(uf.MediaType()) }
+func (uf UnknownFormat) Unmarshal([]byte, *ce.Event) error { return unknown(uf.MediaType()) }
+
 // Prefix for event-format media types.
 const Prefix = "application/cloudevents"
 
--- a/pkg/binding/copy_message_test.go
+++ b/pkg/binding/copy_message_test.go
@@ -3,6 +3,7 @@ package binding_test
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/cloudevents/sdk-go/pkg/binding"
@@ -10,7 +11,7 @@ import (
 	"github.com/cloudevents/sdk-go/pkg/cloudevents"
 )
 
-type toEventTestCase struct {
+type copyMessageTestCase struct {
 	name         string
 	isStructured bool
 	isBinary     bool
@@ -18,11 +19,11 @@ type toEventTestCase struct {
 	want         cloudevents.Event
 }
 
-func TestToEvent(t *testing.T) {
-	tests := []toEventTestCase{}
+func TestCopyMessage(t *testing.T) {
+	tests := []copyMessageTestCase{}
 
 	for _, v := range test.Events() {
-		tests = append(tests, []toEventTestCase{
+		tests = append(tests, []copyMessageTestCase{
 			{
 				name:         "From structured with payload/" + test.NameOf(v),
 				isStructured: true,
@@ -70,11 +71,16 @@ func TestToEvent(t *testing.T) {
 	for _, tt := range tests {
 		tt := tt // Don't use range variable in Run() scope
 		t.Run(tt.name, func(t *testing.T) {
-			got, isStructured, isBinary, err := binding.ToEvent(tt.message)
+			copy, err := binding.CopyMessage(tt.message)
 			require.NoError(t, err)
-			require.Equal(t, tt.isStructured, isStructured)
-			require.Equal(t, tt.isBinary, isBinary)
-			test.AssertEventEquals(t, test.ExToStr(t, tt.want), test.ExToStr(t, got))
+			// The copy can be read any number of times
+			for i := 0; i < 3; i++ {
+				got, isStructured, isBinary, err := binding.ToEvent(copy)
+				assert.NoError(t, err)
+				require.Equal(t, tt.isStructured, isStructured)
+				require.Equal(t, tt.isBinary, isBinary)
+				test.AssertEventEquals(t, test.ExToStr(t, tt.want), test.ExToStr(t, got))
+			}
 		})
 	}
 }
--- a/pkg/binding/event_message.go
+++ b/pkg/binding/event_message.go
@@ -1,15 +1,71 @@
 package binding
 
-import ce "github.com/cloudevents/sdk-go/pkg/cloudevents"
+import (
+	"bytes"
+
+	"github.com/cloudevents/sdk-go/pkg/binding/spec"
+	ce "github.com/cloudevents/sdk-go/pkg/cloudevents"
+)
+
+var specs = spec.New()
 
 // EventMessage type-converts a cloudevents.Event object to implement Message.
 // This allows local cloudevents.Event objects to be sent directly via Sender.Send()
 //     s.Send(ctx, binding.EventMessage(e))
 type EventMessage ce.Event
 
-func (m EventMessage) Event(builder EventEncoder) error   { return builder.SetEvent(ce.Event(m)) }
-func (m EventMessage) Structured(StructuredEncoder) error { return ErrNotStructured }
-func (m EventMessage) Binary(BinaryEncoder) error         { return ErrNotBinary }
-func (EventMessage) Finish(error) error                   { return nil }
+func (m EventMessage) Event(builder EventEncoder) error {
+	return builder.SetEvent(ce.Event(m))
+}
+
+func (m EventMessage) Structured(StructuredEncoder) error {
+	return ErrNotStructured
+}
+
+func (m EventMessage) Binary(b BinaryEncoder) (err error) {
+	c := m.Context
+	var sv spec.Version
+	sv, err = specs.Version(c.GetSpecVersion())
+	if err != nil {
+		return err
+	}
+	set := func(k spec.Kind, v interface{}) {
+		if err == nil {
+			attr := sv.AttributeFromKind(k)
+			if attr != nil {
+				err = b.SetAttribute(attr, v)
+			}
+		}
+	}
+	set(spec.SpecVersion, c.GetSpecVersion())
+	set(spec.Type, c.GetType())
+	set(spec.Source, c.GetSource())
+	set(spec.Subject, c.GetSubject())
+	set(spec.ID, c.GetID())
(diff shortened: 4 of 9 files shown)


alanconway:
@slinkydeveloper for you - check it carefully it was rushed. Probably some overlap with you changes that can be cleaned up.

pkg/binding/copy_message.go
```diff
@@ -0,0 +1,26 @@
...
+		break
+	default:
+		return nil, err
+	}
+	em := EventMessage{}
+	err = m.Event(&em)
+	if err != nil {
+		return nil, err
+	}
+	return WithFinish(em, func(err error) { _ = m.Finish(err) }), nil
```
slinkydeveloper: Binding the original message lifecycle to the copied message is probably what you don't want to do if you want to decouple the lifecycles of the two messages (eg like you explained in this comment point 2 https://github.com/cloudevents/sdk-go/pull/282#issuecomment-574906328)

pkg/binding/format/format.go
```diff
@@ -24,6 +24,14 @@ type Format interface {
 	Unmarshal([]byte, *ce.Event) error
 }
 
+// UnknownFormat allows an event with an unknown format string to be forwarded,
+// but Marshal() and Unmarshal will always fail.
+type UnknownFormat string
```
slinkydeveloper: Why did you added this new format? Any specific reason?

alanconway:
> Your solution looks good for what regards the _structured_ encoding. However IMO going through the `EventMessage` for binary messages is a step backward compared to #270, since all the optimization done for the message transferring in binary -> binary (using the `BinaryEncoder`) is neutralized in the "fanout" case. 

There's no difference between structured and fanout: in both cases you incur the extra cost of creating one EventMessage (using the binary optimizations if relevant) followed by the cost of writing N output messages. If N==1 then it is faster to go direct, but for N>1 it *may* be faster to read the EventMessage, and even if it isn't the overhead drops off as N gets bigger. 

Given the potential need to route/replay etc. multicast messages after the incoming connection has dropped, my gut tells me the EventMessage is worth it -  but it depends on relative cost of reading the input message vs. the EventMessage.

> 
> Brain dump: Maybe we can implement a solution similar to yours, some kind of union type that every Message implementation should contain to keep the whole event (in case of structured) or the event payload, that can be:
> 
>     * buffered or unbuffered
> 
>     * can be accessed or as a byte array, or as `io.Reader`. In the latter, message is buffered and that buffer has its own lifecycle and it's eventually reused
> Then, `Message` interface could have an accessor method to retrieve it so we can create a method called `BufferMessage` that invokes it and transforms this union type from unbuffered to buffered.

At this point I'd trade simplicity for an optimization that might actually be a pessimization in some cases:
- binding creates needless internal copy that is never used.
- binding offers repeatable read method that is less efficient than caching as EventMesage and reading that. 
I'd like to see the binding implement the simplest, fastest "read once" strategy and let the application decide if it needs caching for whatever reason (read many, lifecycle etc.)

I think this is straying into the realms of "premature optimization" - a more complex API can itself have negative performance impact because people who have to spend time learning/implementing a complex  API have less time to think about performance.

pkg/binding/copy_message.go
alanconway: Right, my gut tells me that in cases where you need to buffer for one reason (mult-write) you will probably need to buffer for others (retry after incoming message is gone etc.) so there's' limited value in optimizing the special case of buffer-for-multi-write-only as it probably won't come up much in reality. I could be wrong but lets fix that when there's a real use case and the creation of a single extra EventMessage really is the hot spot on the critical path.

pkg/binding/format/format.go
alanconway: Your changes pass a Format in places where we don't actually know if we can locally encode/decode it with a Format. Previously I had passed a string in such places, but I like the type hint of of passing a Format - so this is a special Format that allows you to pass a string through APIs that take a Format but fails if you actually try to encode or decode.

slinkydeveloper:
>  but for N>1 it may be faster to read the EventMessage than multi-read the original incoming message.

So are you saying that having a single buffered message, which is read multiple times to dispatch the message to N senders, could be actually slower than copying the input message N times to create one message per sender?

slinkydeveloper:
> binding creates needless internal copy that is never used

What copy?

> binding offers repeatable read method that is less efficient than caching as EventMesage and reading that.

I get this one, it was one of my original concerns too with this design. Initially i thought that binding implementer should be responsible for implementing the retry mechanism _after_ the message is read and translated to the "final" request structure (eg the `HttpRequest`). Perhaps now I should reconsider that choice...

> I'd like to see the binding implement the simplest, fastest "read once" strategy and let the application decide if it needs caching for whatever reason (read many, lifecycle etc.)

I agree with this concern, perhaps i should reconsider some choice...

alanconway:
> > but for N>1 it may be faster to read the EventMessage than multi-read the original incoming message.
> 
> So are you saying that having a single buffered message, which is read multiple times to dispatch the message to N senders, could be actually slower than copying the input message N times to create one message per sender?

There are the folllowing operations involved:
- RI - read from input Message
- WE - write to EventMessage
- RE - read from EventMessage
- WO - write to output message

A) With no intermediate copy the cost is N*(RI+WO)
B) With an intermediate EventMessage copy the cost is: RI + WE + N*(RE+WO) 

If N=1 then clearly A < B : (RI+WO) < RI + WE + RE + WO

However in the case N>1 it depends on the relative cost of RI and RE. If it is more expensive to read from the inbound message than from the trivial EventMessage  (e.g. inbound read is parsing JSON or YAML, decoding AMQP or whatever) then at some value of N, B will become cheaper than A.

alanconway:
> > binding creates needless internal copy that is never used
> 
> What copy?

Suppose I have a JSON binding that does a JSON parse every time I read from the Message. I can do multiple reads, but there'll be a parse every time so for N>1 it will pretty quickly become more expensive than parsing once into an EventMessage and then reading from there N times.

Now you could demand that every binding support *efficient* read-many - but in that case the JSON binding would have to do the exact equivalent of creating an EventMessage internally on the first read, and worse - the JSONMessage can't know if the copy is needed or not, so it will make one even when we are actually going direct to the outbound message - defeating that direct no-copy optimization. 

So I'd prefer the JSONMessage to be a simple JSON parser and let the application decide if can stream the message directly input-to-output or if it needs to make a copy for lifecycle reasons. The tricky case  is when N>1 and the app doesn't need a copy for lifecycle reasons. Nobody knows for sure if it will be more efficient to make a temp copy or do N direct copies. At this point the optimization benefits are uncertain so  I would opt for simplicity - just always do a copy, it's safer.

> I get this one, it was one of my original concerns too with this design. Initially i thought that binding implementer should be responsible for implementing the retry mechanism _after_ the message is read and translated to the "final" request structure (eg the `HttpRequest`). Perhaps now I should reconsider that choice...

If the fan-out is all to the same binding then that makes perfect sense and you should go direct to HttpMessage. However as we want to build more interesting event routers, bridges etc. then fan-out to multiple different kinds of bindings is sure to come up (e.g. mix of HTTP1 + HTTP2, different structured encodings, AMQP etc. etc.) 

Using EventMessage is a strategy when you need a binding-neutral,, in-memory representation. That won't always be the case (you've pointed out some good examples) but it will definitely come up a lot.
The key is to keep it transparent so input Message and output Message are simple, and focused on the details of their encoding/protocol. That way you can pair an Message X and Y, where either, both or neither of X and Y are an EventMessage.

slinkydeveloper:
> B) With an intermediate EventMessage copy the cost is: RI + WE + N*(RE+WO)

I agree with the idea that we should go through an intermediate buffered message representation, respectively one for structured and one for binary, and that this intermediate representation should be "readable" by different type of `Sender`s at same time, although I'm worried that `Event` metadata marshalling/unmarshalling could bring us back to beginning (but this is a mere speculation, i need to test it to be sure). Even if that's the case, probably optimizing `EventMessage` could be the way to go more than creating a new intermediate representation (maybe with lazy metadata parsing?).

Anyway, the problem I see in this solution is that, using the actual visitors I've implemented, to read the `StructuredMessage` the "reader" doesn't receive directly the `[]byte`, but it receives an `io.Reader` to the underlying `[]byte`. So the ops cost is what you explained, while the actual copies ( so memory allocations) **when the binding creates the final structure using directly a []byte** (eg `AMQPMessage`) are 1 (the initial buffering) + N (for each send). And that's the reason why I've introduced this kind of "union type" here: https://github.com/cloudevents/sdk-go/pull/282/files#diff-4e71329bf72eef98f921c6239a5e8072R91 . So if the sender final structure accepts []byte, it just reuses the cached one more than going through an additional copy. Hence my question, does it makes sense this additional copy?

slinkydeveloper:
Anyway I've opened a ticket to better investigate the various use cases, the actual one covers just one use case https://github.com/cloudevents/sdk-go/issues/285

alanconway:
On Tue, Jan 21, 2020 at 7:20 AM Francesco Guardiani <
notifications@github.com> wrote:

> B) With an intermediate EventMessage copy the cost is: RI + WE + N*(RE+WO)
>
> I agree with the idea that we should go through an intermediate buffered
> message representation, respectively one for structured and one for binary,
> and that this intermediate representation should be "readable" by different
> type of Senders at same time, although I'm worried that Event metadata
> marshalling/unmarshalling could bring us back to beginning (but this is a
> mere speculation, i need to test it to be sure). Even if that's the case,
> probably optimizing EventMessage could be the way to go more than
> creating a new intermediate representation (maybe with lazy metadata
> parsing?).
>
> Anyway, the problem I see in this solution is that, using the actual
> visitors I've implemented, to read the StructuredMessage the "reader"
> doesn't receive directly the []byte, but it receives an io.Reader
>
The standard io.Copy() operation is already optimized for the case where
either or both sides is actually a bytes.Reader/Writer. E.g. if you re-use
a bytes.Buffer as the target of io.Copy from a sequence of bytes.Reader
sources, then the buffer will grow exactly once for each time the source is
bigger than the target - your classic "grow to demand" buffer. So io.Reader
already provides a good abstraction for arbitrary data sources that is
optimized for byte-array sources.


> to the underlying []byte. So the ops cost is what you explained, while
> the actual copies ( so memory allocations) *when the binding creates the
> final structure using directly a []byte* (eg AMQPMessage) are 1 (the
> initial buffering) + N (for each send). And that's the reason why I've
> introduced this kind of "union type" here:
> https://github.com/cloudevents/sdk-go/pull/282/files#diff-4e71329bf72eef98f921c6239a5e8072R91
> . So if the sender final structure accepts []byte, it just reuses the
> cached one more than going through an additional copy. Hence my question,
> does it makes sense this additional copy?
>
Copying thru an EventMessage using io.Reader on both sides won't make any
extra copies or incur any extra costs if the target uses io.Copy and a
bytes.Buffer (or any impl of bytes.Writer). The io and bytes packages do a
pretty good job of optimizing down to simple memory copies when that is
possible - but it depends on using them consistently. That's why I prefer
to use them directly rather than hide them in new interfaces.

slinkydeveloper:
Ok so I've ran some tests of [this branch](https://github.com/cloudevents/sdk-go/pull/282) (your code + minor nits + [code to adapt the bench](https://github.com/slinkydeveloper/sdk-go/blob/message-buffer/test/benchmark/http/main.go#L94)) vs my branch (+ [code to adapt to bench](https://github.com/slinkydeveloper/sdk-go/blob/multi_sender/test/benchmark/http/main.go#L85))

Below part of the results (left is this branch, right is my branch):

Ns/ops with parallelism 4 and payload 1kb

Schermata da 2020-01-22 18-08-28

Ns/ops with parallelism 4 and payload 8kb

Schermata da 2020-01-22 18-11-11

Allocations with parallelism 4 and payload 1kb

Schermata da 2020-01-22 18-12-14

Allocations with parallelism 4 and payload 8kb

Schermata da 2020-01-22 18-06-26

Apart the binary to structured case that seems somehow broken (i'll further investigate on it to see what we can do about it), in multi sender case my branch constantly outperform this branch of 20/30% both on memory allocs that ns/ops. But my branch has pooling, so that's something we should consider among the others.

I'll try to move some pieces, like adding pooling, and get back with other results to see how we can get the two solutions closer.

slinkydeveloper:
@alanconway I've opened a new PR where, starting from your work here, I add pooling and hiding the `CopyMessage` result message implementations, allowing us to efficiently implement it. Let me know WDYT

n3wscott:

PR #284, opened 2020-01-20
Fix: defeat http.Client memory leak

AlexStocks:
u should check t.Client is nil or not again in mutex lock

### DIFF
--- a/pkg/cloudevents/transport/http/transport.go
+++ b/pkg/cloudevents/transport/http/transport.go
@@ -170,7 +170,9 @@ func (t *Transport) Send(ctx context.Context, event cloudevents.Event) (context.
 func (t *Transport) obsSend(ctx context.Context, event cloudevents.Event) (context.Context, *cloudevents.Event, error) {
 	if t.Client == nil {
 		t.crMu.Lock()
-		t.Client = &http.Client{}
+		if t.Client == nil {
+			t.Client = &http.Client{}
+		}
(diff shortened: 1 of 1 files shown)


pkg/cloudevents/transport/http/transport.go
```diff
@@ -170,7 +170,9 @@ func (t *Transport) Send(ctx context.Context, event cloudevents.Event) (context.
 func (t *Transport) obsSend(ctx context.Context, event cloudevents.Event) (context.Context, *cloudevents.Event, error) {
 	if t.Client == nil {
 		t.crMu.Lock()
-		t.Client = &http.Client{}
+		if t.Client == nil {
```
n3wscott: this code can only get here if `t.Client == nil`, see line 171

AlexStocks: however, this is not thread safe. It there are two goroutines exec this func concurrently, the first create the t.Client at line 173, and the other is blocked at line 172. Later on the first goroutine release the lock  at line 174, and the second goroutine will get the lock and create the t.Client again at line 173.

n3wscott: ah I see, can you kill line 171?

n3wscott: well... nevermind, no need to toggle that lock over and over....

n3wscott:
LGTM

PR #289, opened 2020-01-23
Fixing PubSub Codec to only support v1 string extensions 

nachocano:
- v1 should support string extensions. In the pubsub codec, we had some legacy code from 0.3, which supported maps, and other types as well.
- This caused a panic when using it from a receive adapter code, trying to decode a "int64" extension: 
panic: cannot convert 1.57947362937361e+15 to int32: out of range

Signed-off-by: Nacho Cano <nachoacano@gmail.com>

### DIFF
--- a/pkg/cloudevents/transport/pubsub/codec_v1.go
+++ b/pkg/cloudevents/transport/pubsub/codec_v1.go
@@ -2,7 +2,6 @@ package pubsub
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


pkg/cloudevents/transport/pubsub/codec_v1.go
```diff
@@ -118,27 +117,13 @@ func (v CodecV1) toAttributes(e cloudevents.Event) (map[string]string, error) {
...
-		}
-		encoded, err := json.Marshal(v)
+		k = strings.ToLower(k)
+		cstr, err := types.Format(v)
 		if err != nil {
-			return nil, err
+			return a, err
 		}
-		a[prefix+k] = string(encoded)
+		a[prefix+k] = cstr
```
n3wscott: don't use this. you want to do `e.SetExtension()`

n3wscott: oooh sorry sorry, this is pushing to the pubsub side. 

This is ok.

n3wscott:
LGTM

Issue #290, opened 2020-01-23
SDK should not panic


nachocano:
- Currently the SDK is panicking in multiple functions, without giving the chance to the caller code to decide what to do. In most of the cases, those errors are recoverable, and they shouldn't, for example, take down a pod. 
- The main issue is in https://github.com/cloudevents/sdk-go/blob/master/pkg/cloudevents/event_writer.go, with all those methods.
-  I argue that instead of panicking, those methods should bubble up the error, so that calling code can decide what to do. This will imply changing the signature of all these publicly exposed APIs, and therefore change all of the client code using them (eventing, eventing-contrib, etc.)
- I think this is a serious issue as oftentimes we might be panicking during execution.
- Panic should be used instead for things that are unrecoverable. A malformed event is not, client code should probably just log the error... 

Thoughts?
/cc @n3wscott 
/cc @grantr

n3wscott:
The intention was that these are all convenience methods and I wanted it to be a simple set, set, set vs set check error, set check err... 

You can recover: https://blog.golang.org/defer-panic-and-recover

nachocano:
I understand the intention, but I still believe we shouldn't be doing this.
We should not be panicking in library code.

That recover thing is kind of an ugly hack IMO, and probably costly.

Maybe some Builder would help with having just one check for error in
client code. But I actually don't mind set check error as long as my
program doesn't crash.

grantr:
https://eli.thegreenplace.net/2018/on-the-uses-and-misuses-of-panics-in-go/ has some interesting quotes and discussion about the Go philosophy on panics vs errors. Rob Pike said:

> We want you think of panics as, well, panics! They are rare events that very few functions should ever need to think about. If you want to protect your code, one or two recover calls should do it for the whole program. If you're already worrying about discriminating different kinds of panics, you've lost sight of the ball.

I suspect Rob would frown on these functions translating errors into panics.

The practical problem here is that these functions do not suggest in either signature or documentation that they can fail. This led to processes crashing for unknown reasons. If these methods had been documented as panicking or had a signature that returned an error, then their user would know that they'd need to handle those errors, but as it is now, there's no indication that they'll cause a crash until the crash occurs.

If these methods stick around in their current panicking form, their error behavior needs to be carefully and loudly documented. I'd also recommend suffixing them with `OrDie` so anyone looking at code that calls them gets a hint of what they'll do.

n3wscott:
[...]

PR #294, opened 2020-01-28
Rework panic out of the api.

n3wscott:
fixes: https://github.com/cloudevents/sdk-go/issues/290

Before the SDK would panic on the builder like methods for setting fields on the event as it constructed if there was a clear syntactic violation. 

This change moves the bubbling of those errors for when they call event.Validate(), the errors are assembled and sent as a single error.

A user can clear an error if they `event.Set<Field>(correctedValue)` again on the same field.

### DIFF
--- a/pkg/cloudevents/event.go
+++ b/pkg/cloudevents/event.go
@@ -13,11 +13,25 @@ type Event struct {
 	Data        interface{}
 	DataEncoded bool
 	DataBinary  bool
+	FieldErrors map[string]error
 }
 
 const (
 	defaultEventVersion = CloudEventsVersionV02
 )
+
+func (e *Event) fieldError(field string, err error) {
+	if e.FieldErrors == nil {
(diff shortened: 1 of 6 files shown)


pkg/cloudevents/event.go
```diff
@@ -13,12 +13,26 @@ type Event struct {
 	Data        interface{}
 	DataEncoded bool
 	DataBinary  bool
+	FieldErrors map[string]error
```
nachocano: users will check whether `event.FieldErrors != nil`, right?

nachocano: discussed offline

nachocano:
LGTM

pkg/cloudevents/event.go
n3wscott: so no need to call Validate on caller code then... that is done in the SDK in client.Send

nachocano: sycned offline

PR #302, opened 2020-02-07
Rework binding APIs algorithms and public methods

### LINKED ISSUE #286: Direct use of Encoders should be possible
Both existing bindings, HTTP and AMQP, do not export their Encoders. So they can only be instantiated within their packages. The only place I found, was the according `Send()` method of the Senders. How can I use the Encoders directly, if I prefer to handle the transport directly? In HTTP I might want to use something else then POST, or I might want to implement a special retry logic or react to return codes which do not seem to be exposed.

### LINKED ISSUE #296: Implement binding transcoders to read/add/update metadata/extensions
Now the encoding process is quite static and doesn't allow any plug-in mechanism to read, add & update a metadata or an extension (eg tracing extension is a reasonable example). This will cause to fallback always to event for users that needs to perform some transformations on the metadata.
An attempt to solve this problem is already there: https://github.com/cloudevents/sdk-go/blob/master/pkg/binding/transcoder.go#L8, but it only allows to update an already existing metadata.
We need to cover the following use-cases:

* Read a specific meta
* Add a specific meta if not existing
* Update an existing specific meta if existing
* Remove a specific meta

Those operations should be done directly when the message is transformed (using one of the three methods of Message: Binary, Structured and Event), so they can be potentially used or when the message is buffered or when the message is translated to the final protocol struct (like `HttpRequest`).
Probably those operations can't be implemented for structured encoding, which should be transformed to Event (like in [`VersionTranscoder`](https://github.com/cloudevents/sdk-go/blob/master/pkg/binding/transcoder/version.go#L12))

### LINKED ISSUE #299: Define how to handle fallback for non-cloudevents messages in bindings
There are some use cases where there is the need to handle both cloudevents messages and non-cloudevents messages, wrapping the latter in valid cloudevents. We must define how we can handle this use case in bindings APIs

### LINKED ISSUE #300: Define the encoders invocation order in binding.Translate

slinkydeveloper:
* Removed EventEncoder, as it was duplicating a lot of code in bindings implementations. Now the binding implementer must implement two encoders more than four.
* Moved some methods and `EventMessage` from `binding/event` to `binding`
* `Message` now has two additional methods:
  * `GetParent()` to return the encapsulated event, if any. binding implementer usually just returns `nil` for this
  * `Encoding()` to return the message encoding. This forces the binding implementer to check, when the message is created, the message encoding, allowing to implement fallback strategies for non cloudevents messages (closes #299)
* `BinaryEncoder` has two additional methods to control the lifecycle of the encoding: `Start()` and `End()`
* Now bindings module provides three main algorithms to use encoders and perform message transformations: 
  * `ToEvent()` to go from any message to event, optionally performing a transformation in the middle
  * `RunDirectEncoders()` to perform a direct encoding of a structured or binary message, optionally performing a transformation in the middle
  * `Encode()` which combines the previous two. It tries to use RunDirectEncoders, otherwise it goes through the Event data structure and then re-encode back using always the same specified encoders (closes #300)
* Reworked both http and amqp implementations, also offering public methods to use the protocol encoders (closes #286)
* Implemented facilities to simplify add/remove/update of metadata & extensions (closes #296)
* Added `context.Context` to a bunch of methods, to tweak the encoding algorithms: `Encode()`, `RunDirectEncoders()`, `ToEvent()`, `Message.Structured()`, `Message.Binary()`, `BinaryEncoder.Start()`, `StructuredEncoder.SetStructuredEvent()`, `CopyMessage()`, `BufferMessage()`


pkg/binding/buffering/binary_buffer_message.go
```diff
@@ -21,35 +20,52 @@ type binaryBufferedMessage struct {
 	body       *bytebufferpool.ByteBuffer
 }
 
+func (m *binaryBufferedMessage) Init() error {
```
n3wscott: for all these methods, I have found passing ctx helps for a bunch of things like tracing, metrics, and a hook to extend singleton state into methods that don't have to have the api change.

pkg/binding/example_implementing_test.go
```diff
@@ -64,18 +62,15 @@ func NewExSender(w io.Writer, factories ...binding.TransformerFactory) binding.S
 }
 
 func (s *ExSender) Send(ctx context.Context, m binding.Message) error {
-	// Translate tries the various encodings, starting with provided root encoder factories.
+	// RunEncoders tries the various encodings, starting with provided root encoder factories.
```
n3wscott: This comment is off, I think RunEncoders -> Encode

pkg/binding/buffering/binary_buffer_message.go
slinkydeveloper: Apart the `Init()` and `End()`, I don't see the reason to pass a `Context` in `BinaryEncoder` and `StructuredEncoder` interface. But I see the reason to do that in `Encode`, `RunDirectEncoders`, `CopyMessage` and `BufferMessage`

pkg/binding/interfaces.go
```diff
@@ -34,6 +49,14 @@ import (
 // try each method of interest and fall back to Event(EventEncoder) if none are supported.
 //
 type Message interface {
+	// TODO(slinkydeveloper) add docs
+	GetParent() Message
```
n3wscott: This feels two things:

1 - I think GetParent is a little confusing, and I would expect this to be just GetMessage or GetWrappedMessage

2 - Consider breaking this out of the Message interface into a new WrappedMessage interface you can use to implement accessor methods and do a cast to WrappedMessage when you need to do the internal grab.

pkg/binding/interfaces.go
```diff
@@ -54,19 +77,6 @@ type Message interface {
 	// intermediate Event.
 	Binary(BinaryEncoder) error
```
n3wscott: I am not sure what `Binary` does. It might need a verb in there, like `ToBinary`

pkg/binding/interfaces.go
```diff
@@ -98,6 +108,10 @@ type StructuredEncoder interface {
 // Protocols that supports binary encoding should implement this interface to implement direct
 // binary -> binary transfer.
 type BinaryEncoder interface {
+	//TODO explain init and end with relative propagation!!!
+
+	Init() error
```
n3wscott: `Init` implies only once per `BinaryEncoder` lifecycle, and guessing from usage, this happens once per message encoding? Please explain.

n3wscott: If once per BinaryEncoder, then init is correct. But if once per message, then something like `Start` or `Begin` might make more sense

slinkydeveloper: So `BinaryEncoder` interface is usually tied to one single "output" message, it's not reused (https://github.com/cloudevents/sdk-go/pull/302/files#diff-c5d6e121af538561803b092335cebcefR45)

`Init()` is run at the beginning of the encoding, so yeah maybe `Start()` is better

pkg/binding/interfaces.go
slinkydeveloper: `Binary` and `Structured` are the two methods to access to the message structure. I can argue that `To[X]` is not suitable, because it hints that returns a value of type X, while here it doesn't return nothing (except the `error`) because it mutates the state of `BinaryEncoder` you pass

pkg/binding/interfaces.go
slinkydeveloper: Does it looks better now? https://github.com/cloudevents/sdk-go/pull/302/commits/259f6f4857e2136bfb9370459fb4f9bfb295e663

pkg/binding/buffering/binary_buffer_message.go
slinkydeveloper: wdyt https://github.com/cloudevents/sdk-go/pull/302/commits/5e760fa2f6981a222cc1f58983e95df18b00ad54 ?

pkg/binding/interfaces.go
n3wscott: +1 nice

n3wscott:
Looking good!!

slinkydeveloper:
For me we can start reviewing this one and look for merging

slinkydeveloper:
@n3wscott any ideas on how to change the name "transformer" or "transcoder"?

pkg/binding/buffering/copy_message.go
```diff
@@ -1,15 +1,15 @@
...
 	"github.com/cloudevents/sdk-go/pkg/binding"
-	"github.com/cloudevents/sdk-go/pkg/binding/event"
-	"github.com/cloudevents/sdk-go/pkg/binding/spec"
 )
 
 // BufferMessage does the same than CopyMessage and it also bounds the original Message
 // lifecycle to the newly created message: calling Finish() on the returned message calls m.Finish()
-func BufferMessage(m binding.Message) (binding.Message, error) {
-	result, err := CopyMessage(m)
+func BufferMessage(ctx context.Context, m binding.Message, transformers ...binding.TransformerFactory) (binding.Message, error) {
```
n3wscott: We can migrate `transformer` in a followup...

n3wscott:
I have looked at this a few times now, and nice work on a lot of tests.

Let's merge and work in master.

LGTM

PR #304, opened 2020-02-14
Add SetContextHeaders setter.

ian-mi:
It is occasionally useful to set new headers while discarding headers
that were set for a previous request. See for example
https://github.com/knative/eventing/issues/1953.


n3wscott:
LGTM

PR #305, opened 2020-02-15
Replace the HTTP transport Client field with unexported transport field.

ian-mi:
Add a new HTTP option which allows setting the HTTP transport while
keeping the field unexported for improved safety. This change will allow
the internal HTTP client to be constructed with additional middleware
if needed, for example, for trace propagation.

### DIFF
--- a/pkg/cloudevents/transport/http/transport.go
+++ b/pkg/cloudevents/transport/http/transport.go
@@ -44,11 +44,6 @@ type Transport struct {
 	// If nil, DefaultShutdownTimeout is used.
 	ShutdownTimeout *time.Duration
 
-	// Sending
-
-	// Client is the http client that will be used to send requests.
-	// If nil, the Transport will create a one.
-	Client *http.Client
 	// Req is the base http request that is used for http.Do.
 	// Only .Method, .URL, .Close, and .Header is considered.
 	// If not set, Req.Method defaults to POST.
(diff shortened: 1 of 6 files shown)


pkg/cloudevents/transport/http/transport.go
```diff
@@ -44,11 +44,6 @@ type Transport struct {
 	// If nil, DefaultShutdownTimeout is used.
 	ShutdownTimeout *time.Duration
 
-	// Sending
-
-	// Client is the http client that will be used to send requests.
-	// If nil, the Transport will create a one.
-	Client *http.Client
```
n3wscott: This would be a breaking semver change, can you rework the PR to not change the api?

n3wscott:
LGTM, but I don't want to bump the api to 2.0 yet, can you leave httptransport.Client in place and either mark it deprecated and/or use it with the new inner transport from context?

ian-mi:
I've marked the Client field deprecated with the behavior that setting Client will cause WithHTTPTransport to be ignored.

pkg/cloudevents/transport/http/transport.go
ian-mi: Done.

pkg/cloudevents/transport/http/transport.go
```diff
@@ -46,9 +46,10 @@ type Transport struct {
 
 	// Sending
 
-	// Client is the http client that will be used to send requests.
-	// If nil, the Transport will create a one.
+	// Deprecated - setting http client will override use of the
```
n3wscott: THANKS!!!

n3wscott:
Thanks for the non-breaking change to the API.

LGTM.

PR #307, opened 2020-02-18
Fix copy paste docs error

johanbrandhorst:
The WithMethod function was using the same docstring as the WithTarget function


n3wscott:
Thanks for the catch!
/lgtm

PR #309, opened 2020-02-20
Add Distributed Trace Propagation

ian-mi:
Enables trace propagation using the distributed tracing extension by default in the eventing client with a new option to disable trace propagation. Adds a HasTracePropagation transport property which transports can use to indicate that the transport's incoming trace context should be preferred to the context defined by the distributed tracing extension.


slinkydeveloper:
I'm not sure it's a good idea to add a method about extensions to the `Transport` interface, nor add extension logic to the `Client` code directly... Should we reason about creating some abstractions to decouple this code?

ian-mi:
Could you elaborate on what type of abstraction you feel is lacking? The alternative to handling this in the client seemed to be adding this capability into each transport which would involve duplicating code across each transport. Client seemed like the appropriate place to put cross-cutting functionality but the client needs some information about the capabilities of the transport.

n3wscott:
@slinkydeveloper  This is adding the official CloudEvents trace extension to the SDK. We are suppose to support official extensions from https://github.com/cloudevents/spec/blob/master/SDK.md#extensions.

pkg/cloudevents/transport/transport.go
```diff
@@ -19,6 +19,8 @@ type Transport interface {
 	SetConverter(Converter)
 	// HasConverter is true when a non-nil converter has been set.
 	HasConverter() bool
+
+	HasTracePropagation() bool
```
n3wscott: Can you add godoc to this?

pkg/cloudevents/transport/http/transport.go
```diff
@@ -353,6 +353,12 @@ func (t *Transport) StartReceiver(ctx context.Context) error {
 	}
 }
 
+// HasTracePropagation implements Transport.HasTracePropagation
+// TODO: implement HTTP trace propagation.
```
n3wscott: this todo means there is still more to enable this feature?

pkg/cloudevents/client/client.go
```diff
@@ -107,6 +111,15 @@ func (c *ceClient) obsSend(ctx context.Context, event cloudevents.Event) (contex
 		}
 	}
 
+	// Set distributed tracing extension.
+	if !c.disableTracePropagation {
+		if span := trace.FromContext(ctx); span != nil {
```
n3wscott: So spans are only propagated if they exist on the context, correct?

n3wscott:
What is the plan with regards to `tracestate` ? https://github.com/cloudevents/spec/blob/master/extensions/distributed-tracing.md#tracestate

pkg/cloudevents/transport/http/transport.go
ian-mi: I've gone ahead and added HTTP propagation in this PR.

pkg/cloudevents/client/client.go
ian-mi: Correct, although if tracing is enabled in the SDK a new trace will be created immediately before this.

ian-mi:
This PR should correctly handle tracestate. I've made parsing fail-open however so an invalid tracestate will not fail a request and will not prevent a valid traceparent from being used.

n3wscott:
LGTM

PR #313, opened 2020-02-22
Remove receive/fn observable.

ian-mi:
Very little work is done outside of receive/fn in the receive
observable making separate observables unnecessary.


n3wscott:
LGTM

n3wscott:
I sprayed measurements all over not sure what would be useful yet. Thanks for the cleanup

Issue #317, opened 2020-02-27
Implement binding API for NATS


dan-j:
As per the title, we need support for different types of NATS subscriptions and NATS Streaming subscriptions. The first step in this journey is to implement the bindings and then the rest can come.

dan-j:
This is no longer relevant with the latest v2 updates

PR #318, opened 2020-02-27
NATS binding API and refactor transport

dan-j:
As per #317 and some discussion in the slack channel

I've also gone and refactored `nats.Transport`.

Everything should be implemented but I've not added tests yet... just wanted some feedback that I'm on the right track before spending time on tests

### DIFF
--- a/pkg/bindings/nats/encoder.go
+++ b/pkg/bindings/nats/encoder.go
@@ -1,0 +1,60 @@
+package nats
+
+import (
+	"context"
+	"github.com/cloudevents/sdk-go/pkg/binding"
+	"github.com/cloudevents/sdk-go/pkg/binding/format"
+	"github.com/cloudevents/sdk-go/pkg/binding/spec"
+	"github.com/nats-io/nats.go"
+	"io"
+	"io/ioutil"
+)
+
+type natsMsgEncoder nats.Msg
+
+func EncodeNatsMsg(ctx context.Context, m binding.Message, msg *nats.Msg, factories binding.TransformerFactories) error {
+	structuredEncoder := (*natsMsgEncoder)(msg)
+
+	_, err := binding.Encode(
+		ctx,
+		m,
+		structuredEncoder,
+		nil,
+		factories,
+	)
+	return err
+}
+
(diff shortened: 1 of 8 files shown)


n3wscott:
Ah sorry, stuff is moving around, can you rebase?

dan-j:
Sure thing.. still need to write tests though 😁

pkg/bindings/nats/encoder.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return err
+	}
+
+	e.Data = bytes
+
+	return nil
+}
+
+func (e *natsMsgEncoder) Start(ctx context.Context) error {
```
slinkydeveloper: You don't need to implement the `BinaryEncoder` methods at all

pkg/bindings/nats/errors.go
```diff
@@ -0,0 +1,5 @@
+package nats
+
+import "fmt"
+
+var BinaryEncodingNotSupported = fmt.Errorf("binary encoding for NATS protocol is not supported")
```
slinkydeveloper: You don't need it, if you need to signal that you don't support the binary encoding, you must use `binding.ErrNotBinary`

pkg/bindings/nats/message.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+var _ binding.Message = (*Message)(nil)
+
+// By the specification, all NATS message payloads must be in the JSON event format.
+// If this isn't the case then this function will return an error
+func NewMessage(msg *nats.Msg) (*Message, error) {
+	raw := make(map[string]json.RawMessage)
+
+	if err := json.Unmarshal(msg.Data, &raw); err != nil {
```
slinkydeveloper: Why do you perform this unmarshalling?

dan-j:
Sorry I've been quiet a couple of weeks, but I'm back on this (and STAN) now...

Is this still required? It looks like the latest v2 updates addressed this?

n3wscott:
Yes, sorry, we worked through nats because we needed to drop the codec and the transport interfaces.

I think this is now implemented.

slinkydeveloper:
@n3wscott is this one still relevant? can we close it?

PR #319, opened 2020-02-28
Separate trace handling from Observable.

ian-mi:
This allows trace spans to be created with a different granularity
than that of Observable metrics. In particular, this commit replaces
the nested send / receive spans into a single send and receive span
created in the client. In addition, this removes spans from the encode / decode reporters which are
created without access to context making them unable to attach tracing
data to the active trace. With the reduced granularity of emitted
spans, it should now be appropriate to always enable trace
handling. As such, the EnableTracing function has been marked as
deprecated and will have no effect.

The client send and receive spans have been renamed with the prefix
"cloudevents" in order to distinguish them from send and receive spans
created by other libraries. Additionally these spans are annotated
with cloud event attributes from handled events.


n3wscott:
LGTM

Issue #321, opened 2020-03-01
Implement binding API for NATS Streaming (STAN)


dan-j:
As discussed in the comments for #136, it's better to implement new transports backed by the bindings API. 

There isn't an issue for the mentioned PR, so creating it for tracking.

I've already 90% done with the actual work, just need to write tests.

n3wscott:
This is done?

PR #136, opened 2019-06-08
Added stan (nats-streaming)

xackery:
This implements https://github.com/nats-io/stan.go as a transport layer based on the existing nats code.

There are a few areas that need cleanup, but first wanted to see if any desire for me to harden and make this a supported transport type.

Thanks!

PR #322, opened 2020-03-01
feat: implement STAN binding API and transport

dan-j:
See issue #321 

This is an alternative PR to #136 but using the new bindings API. I haven't written tests yet, will probably be a couple of days but I'm fully committed to seeing this through this week.

It supports fully configuring:

- `stan.Conn` via the `stan.Option` API
- `stan.Subscription` via the `stan.SubscriptionOption` API
-  Standard or Queue Subscriptions using the `Subscriber` interface in the transport package

It also supports ACK'ing when the subscription is in "manual ack mode" by calling `Message.Finish()`


pkg/bindings/stan/message.go
```diff
@@ -0,0 +1,98 @@
...
+func (m *Message) Encoding() binding.Encoding {
+	return binding.EncodingStructured
+}
+
+func (m *Message) Structured(ctx context.Context, encoder binding.StructuredEncoder) error {
+	return encoder.SetStructuredEvent(ctx, format.JSON, bytes.NewReader(m.Data))
+}
+
+func (m *Message) Binary(context.Context, binding.BinaryEncoder) error {
+	return BinaryEncodingNotSupported
```
slinkydeveloper: By contract you should return `binding.ErrNotBinary`, not a custom error message

pkg/bindings/stan/message.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil
+	}
+}
+
+// By the specification, all NATS message payloads must be in the JSON event format.
+// If this isn't the case then this function will return an error
+func NewMessage(msg *stan.Msg, opts ...MessageOption) (*Message, error) {
+	raw := make(map[string]json.RawMessage)
+
+	if err := json.Unmarshal(msg.Data, &raw); err != nil {
```
slinkydeveloper: Why do you parse the whole message? There is no need: you just need to implement the `Structured()` visitor method (as you did below), then if you need to convert the message to event just use `binding.ToEvent()`

pkg/bindings/stan/message.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil, err
+	}
+
+	if _, err := spec.VS.FindVersion(versionExtractor(raw)); err != nil {
```
slinkydeveloper: Same here, any reason why you do that?

pkg/cloudevents/transport/stan/options.go
```diff
@@ -0,0 +1,34 @@
...
+
+type Option func(*Transport) error
+
+func applyDefaultOptions(t *Transport) error {
+	t.Subscriber = &StandardSubscriber{}
+	return nil
+}
+
+// WithEncoding sets the encoding for STAN transport.
+func WithEncoding(encoding Encoding) Option {
```
slinkydeveloper: You don't need this one too: if you want to have specific version encoding/translation, use `binding/transformers/WithVersion`

slinkydeveloper: If you need it for implementing the `Transport`, look at how it works in `amqp`

slinkydeveloper:
Is this pr related to #318 ?

dan-j:
This PR should only contain NATS Streaming (i.e. STAN) whereas #318 is for plain NATS, they’re two different things although under the hood they both use the NATS protocol.

dan-j:
Good spot on parsing the message in `NewMessage`. I wrote that before I learnt that messages are validated/converted later in the receiving phase by the SDK. Will amend this afternon

pkg/bindings/stan/message.go
dan-j: So before making this change I just went over the amqp implementation to try and remember why I did this in the first place, and it's because amqp pre-parses to determine whether the `binding.Encoding` is `EncodingStructured`, `EncodingBinary` or `EncodingUnknown`. 

Because by the cloudevents spec, the NATS protocol can only be structured, then we're validating this is the case in the `NewMessage` constructor. Notice that the `nats.Message.Encoding()` is hardcoded to return `EncodingStructured`. 

Either way, I see two options, but you understand the SDK better than I do so will wait for your decision or alternative:

1. Keep as is, encoding is always `EncodingStructured` and we validate this in the constructor by parsing the message.
2. Calculate the encoding in the constructor to either `EncodingStructured` or `EncodingUnknown`, we still must parse the message in `NewMessage`, but it doesn't return an error.

pkg/bindings/stan/message.go
dan-j: For the reasons above, let's keep a single thread and resolve this comment

pkg/bindings/stan/encoder.go
```diff
@@ -0,0 +1,56 @@
+package stan
```
n3wscott: Note that we have changed a lot of the world, this will end up in `pkg/protocol/stan` as a writer

pkg/protocol/stan/write_message.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (w *stanMessageWriter) SetStructuredEvent(ctx context.Context, f format.Format, event io.Reader) error {
+	if _, err := w.ReadFrom(event); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (w *stanMessageWriter) Start(ctx context.Context) error {
```
slinkydeveloper: You don't need this to implement `StructuredWriter`

dan-j: I copied this from nats, should I update both protocols?

slinkydeveloper: yes thank you! `Start` and `End` belongs only to `BinaryWriter`

dan-j:
@slinkydeveloper in response to the kafka suggestions, is there any reasoning behind what's exported and what's private? I'm struggling to see the pattern.

slinkydeveloper:
In `Protocol` the exported bits are the various `Sender` & `Receiver` implementations. Anyway, It still needs some work around :)

dan-j:
Ok, I've finished refactoring the STAN protocol into separate consumer/sender parts. 

It still needs tests, and I'm probably not going to get to that until the weekend. Although I'm going to play around using my fork in a real application maybe on Friday, but got a lot of other work things on my plate.

pkg/protocol/stan/protocol.go
```diff
@@ -0,0 +1,106 @@
...
+// New creates a new STAN protocol including managing the lifecycle of the connection
+func NewProtocol(clusterID, clientID, subject string, stanOpts []stan.Option, opts ...ProtocolOption) (*Protocol, error) {
+	conn, err := stan.Connect(clusterID, clientID, stanOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	p, err := NewProtocolFromConn(conn, subject, opts...)
+	if err != nil {
+		return nil, err
```
dan-j: The connection will need closing if this error occurs

pkg/protocol/stan/receiver.go
```diff
@@ -0,0 +1,207 @@
...
+	appliedSubOpts      *stan.SubscriptionOptions // only used locally, stored to avoid recomputing
+}
+
+func NewConsumer(clusterID, clientID, subject string, stanOpts []stan.Option, opts ...ConsumerOption) (*Consumer, error) {
+	conn, err := stan.Connect(clusterID, clientID, stanOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewConsumerFromConn(conn, subject, opts...)
```
dan-j: same here, if an error occurs we've lost reference to an open `conn`

pkg/protocol/stan/receiver.go
```diff
@@ -0,0 +1,207 @@
...
+	receiverOps, err := c.createReceiverOptions()
+	if err != nil {
+		return err
+	}
+
+	if c.Receiver, err = NewReceiver(receiverOps...); err != nil {
+		return err
+	}
+
+	c.sub, err = c.Subscriber.Subscribe(c.Conn, c.Subject, c.Receiver.MsgHandler, c.subscriptionOptions...)
```
dan-j: Should we wait for `ctx.Done()` and call `c.Close()`?

pkg/protocol/stan/sender.go
```diff
@@ -0,0 +1,77 @@
...
+// NewSender creates a new protocol.Sender responsible for opening and closing the STAN connection
+func NewSender(clusterID, clientID, subject string, stanOpts []stan.Option, opts ...SenderOption) (*Sender, error) {
+	conn, err := stan.Connect(clusterID, clientID, stanOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	s, err := NewSenderFromConn(conn, subject, opts...)
+	if err != nil {
+		return nil, err
```
dan-j: close `conn` if error occurs

pkg/protocol/stan/receiver.go
slinkydeveloper: In case of Kafka (and I think it's the same for http) `ctx.Done()` signals the stop of the receiver

pkg/protocol/stan/protocol.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// Send implements Sender.Send
+func (p *Protocol) Send(ctx context.Context, in binding.Message) error {
+	return p.Sender.Send(ctx, in)
+}
+
+// OpenInbound implements Opener.OpenInbound
+func (p *Protocol) OpenInbound(ctx context.Context) error {
+	return p.Consumer.OpenInbound(ctx)
```
dan-j: @n3wscott @slinkydeveloper regarding the slack conversation around how the `client` package handles closing (or lack of). If we relied on closing the context in `client.StartReceiver` here only p.Consumer would be closed, and the connection would remain open. 

Really the `client` package needs to call `protocol.Close()` which can handle all of the connection, sender and consumer.

dan-j:
OK. I think all comments except for tests and examples have been addressed. 

I need to move onto other things for a couple of days, hopefully I will be able to try my fork out in an application and will look at adding tests and examples once I've got that working next week

dan-j:
OK! @slinkydeveloper @n3wscott 

I think this is ready to be merged. There aren't many unit tests in the protocols (only on `options.go`), but the integration tests for the sender and consumer using both regular and queue subscriptions are done. As are sample cmds using the kafka ones as a baseline

.circleci/config.yml
```diff
@@ -9,6 +9,7 @@ jobs:
       - image: lensesio/fast-data-dev
         environment:
           - ADV_HOST=localhost
+      - image: nats-streaming:0.17.0
```
n3wscott: look at that <3

n3wscott:
LGTM!!!

happy to merge this. THANKS for pushing on STAN

n3wscott:
I did the merge if you want it, https://github.com/mercarto/sdk-go/pull/1

dan-j:
Hey @n3wscott, the build still failed on `go mod tidy`, I've just rebased and squashed everything so history it neat.

It's ready to go now 👍

v2/test/integration/stan/stan_test.go
```diff
@@ -0,0 +1,120 @@
+package stan
```
slinkydeveloper: This test checks the `protocol` bits, can you add another test that checks the integration with `client` (in a similar fashion to how we have them for kafka and amqp?)

dan-j: I can, but may I ask what's the benefit? 

My thought process here is that the protocol adheres to a set of interfaces which it compiles against and the client accepts. That should be the only contract between the protocol and the client, and if that broke, wouldn't this test also break?

slinkydeveloper: There is a difference and contracts are well distinguished, but it doesn't hurt to have such integration test. I particularly had hard times with this integration while testing kafka impl, with `StartReceiver` and passed context that should close the receiver.

n3wscott: We can in a followup https://github.com/cloudevents/sdk-go/issues/416

n3wscott:
LGTM

PR #331, opened 2020-03-03
Added AddOrUpdateExtension and AddOrUpdateAttribute

slinkydeveloper:


pkg/binding/transcoder/add_or_update_metadata.go
```diff
@@ -0,0 +1,20 @@
+package transcoder
+
+import (
+	"github.com/cloudevents/sdk-go/pkg/binding"
+	"github.com/cloudevents/sdk-go/pkg/binding/spec"
+)
+
+func AddOrUpdateAttribute(attribute spec.Kind, defaultValue interface{}, updater func(interface{}) (interface{}, error)) []binding.TransformerFactory {
```
n3wscott: `Add or Update` seems like it means `Set`

slinkydeveloper: Could it be confused with `Add`?

n3wscott: I would say drop Add and Update and only have Set, you can set nil or a none type to remove. No need for Add and Update overall

n3wscott:
LGTM

PR #334, opened 2020-03-03
Added time.Now() and UUID defaulter

slinkydeveloper:


n3wscott:
I am not sure this is something the protocol binding should do.

at the moment the defaults only work at the client level, not the transport level.

slinkydeveloper:
My idea of `transformers` module is to provide some common transformers that can be useful to everybody wants to use `Encoder`s related methods, like `Version`, `[Add|Delete|Update][Attribute|Extension]` and so on

n3wscott:
LGTM

Issue #315, opened 2020-02-27
BindingTransport does not allow transport.Converter to be used


dan-j:
Spoken to @slinkydeveloper on Slack about this. It turns out that migrating transports to use the new bindings API isn't compatible yet since there's no way to assign a `transport.Converter` to the `binding.BindingTransport` struct

Changes that need making:

- Add a `Converter transport.Converter` field to the `binding.BindingTransport` struct
- Update the `SetConverter` and `HasConverter` functions
- Inside `BindingTransport.handle()`, use the encoding returned from `ToEvent` and if it's "EncodingUnknown" then use the `Converter` (if it's set)

slinkydeveloper:
Is this one still valid @n3wscott ?

dan-j:
No this isn't valid anymore, I'll close

PR #342, opened 2020-03-04
[WIP] Basic first pass implementation of http cloudevents.transport using bindings

n3wscott:

### DIFF
--- a/pkg/cloudevents/transport/httpb/transport.go
+++ b/pkg/cloudevents/transport/httpb/transport.go
@@ -1,0 +1,327 @@
+package httpb
+
+import (
+	"context"
+	"fmt"
+	"github.com/cloudevents/sdk-go/pkg/binding"
+	"github.com/cloudevents/sdk-go/pkg/bindings/http"
+	cecontext "github.com/cloudevents/sdk-go/pkg/cloudevents/context"
+	"go.opencensus.io/plugin/ochttp"
+	"go.opencensus.io/plugin/ochttp/propagation/tracecontext"
+	"go.uber.org/zap"
+	"io"
+	"net"
+	nethttp "net/http"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport"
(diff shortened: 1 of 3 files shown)


pkg/cloudevents/transport/httpb/transport.go
```diff
@@ -0,0 +1,327 @@
...
+	return ctx, nil, nil
+}
+
+// SetReceiver implements Transport.SetReceiver
+func (t *Transport) SetReceiver(r transport.Receiver) {
+	t.consumer = r
+}
+
+// SetConverter implements Transport.SetConverter
+func (t *Transport) SetConverter(c transport.Converter) {
```
slinkydeveloper: The converter handling IMO should be implemented directly by `binding.BindingTransport`: https://github.com/cloudevents/sdk-go/issues/315

pkg/cloudevents/transport/httpb/transport.go
```diff
@@ -0,0 +1,327 @@
...
+
+// TODO: this is not what we want to use long term. For PoC.
+// Response is an http transport response.
+type Response struct {
+	StatusCode int
+	Message
+}
+
+// Message is an http transport message.
+type Message struct {
```
slinkydeveloper: Why do we need that? For the response?

slinkydeveloper:
Trace extension implementations both on sender and receiver side can be implemented using a transformer. I suppose the flow should be something like:

* `Sender` should put in the encoded message the span from the event, if any
* `Receiver` should put in the generated message the span info, if any

pkg/cloudevents/transport/httpb/transport.go
n3wscott: This is not that kind of converter, it allows you to implement a try again in the client side for custom converters

pkg/cloudevents/transport/httpb/transport.go
n3wscott: It is needed for the struct in Response but that section of the code is not used yet.

n3wscott:
This is suppose to be still WIP

PR #343, opened 2020-03-05
Rethink client with binding

n3wscott:
WIP

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -22,6 +22,11 @@ import (
 // Client interface defines the runtime contract the CloudEvents client supports.
 type Client interface {
 	// Send will transmit the given event over the client's configured transport.
+	// result is a function pointer. Valid result fn signatures are:
+	// * func()
+	// * func(context.Context)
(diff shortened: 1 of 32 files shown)


pkg/client/client.go
```diff
@@ -27,23 +37,20 @@ type Client interface {
...
 	// * func(event.Event)
-	// * func(event.Event) error
+	// * func(event.Event) event.Result
 	// * func(context.Context, event.Event)
-	// * func(context.Context, event.Event) error
-	// * func(event.Event, *event.EventResponse)
-	// * func(event.Event, *event.EventResponse) error
-	// * func(context.Context, event.Event, *event.EventResponse)
-	// * func(context.Context, event.Event, *event.EventResponse) error
+	// * func(context.Context, event.Event) event.Result
```
slinkydeveloper: I think this interface is good also to check if a `transport` implementation effectively supports the request/response model. When you set a function which can return a result, BUT the transport doesn't support writing "responses", this function fails

PR #344, opened 2020-03-05
adding the http bindings requester

n3wscott:


pkg/bindings/http/sender.go
```diff
@@ -45,10 +53,40 @@ func (s *Sender) Send(ctx context.Context, m binding.Message) (err error) {
 	return
 }
 
-func NewSender(client *http.Client, target *url.URL, options ...SenderOptionFunc) binding.Sender {
-	s := &Sender{Client: client, Target: target, transformerFactories: make(binding.TransformerFactories, 0)}
-	for _, o := range options {
-		o(s)
+func (r *Sender) Request(ctx context.Context, m binding.Message) (binding.Receiver, error) {
```
slinkydeveloper: name of the receiver should be `(s *Sender)`

pkg/bindings/http/sender.go
```diff
@@ -45,10 +53,40 @@ func (s *Sender) Send(ctx context.Context, m binding.Message) (err error) {
...
+	msg, err := NewMessage(resp.Header, resp.Body)
+	return &cachedReceiver{Message: msg, Error: err}, err
+}
+
+type cachedReceiver struct {
+	Message binding.Message
+	Error   error
+}
+
+func (r *cachedReceiver) Receive(ctx context.Context) (binding.Message, error) {
```
slinkydeveloper: Now that I see it I frankly don't like it. The contract of `Receiver` gives the idea that every time you invoke `Receive`, or you receive an `io.EOF`, or an error, or a new `Message`, but that's not what happens here...

Maybe should we rework the `Requester` interface to just let `Request` return a `Message`?

n3wscott: The only downside is that if you have return Message, what do you do when there is a batch or stream of events?

slinkydeveloper: Do we plan somehow to support batch? Because I have no clue on how to bring it inside binding APIs now :fearful:

pkg/bindings/http/sender.go
```diff
@@ -46,11 +58,31 @@ func (s *Sender) Send(ctx context.Context, m binding.Message) (err error) {
...
+	}
+	resp, err := r.Client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode/100 != 2 {
+		return nil, fmt.Errorf("%d %s", resp.StatusCode, nethttp.StatusText(resp.StatusCode))
+	}
+
+	return NewMessage(resp.Header, resp.Body), nil
```
slinkydeveloper: Small nit, you can use `NewMessageFromHttpResponse` now

PR #346, opened 2020-03-05
Code documentation for binding & bindings modules

### LINKED ISSUE #340: Break binding/interface.go apart in different files

slinkydeveloper:
Also fixed some interfaces and made some public symbols private

Fixes #340


pkg/binding/binary_encoder.go
```diff
@@ -0,0 +1,39 @@
...
+// Start() and End() methods are invoked every time this BinaryEncoder implementation is used to visit a Message
+type BinaryEncoder interface {
+	// Method invoked at the beginning of the visit. Useful to perform initial memory allocations
+	Start(ctx context.Context) error
+
+	// Set a standard attribute.
+	//
+	// The value can either be the correct golang type for the attribute, or a canonical
+	// string encoding. See package types to perform the needed conversions
+	SetAttribute(attribute spec.Attribute, value interface{}) error
```
n3wscott: for godoc to work the first word in the comment has to be the function name: `SetAttribute`

pkg/binding/doc.go
```diff
@@ -13,31 +15,45 @@ most "endpoint" applications that emit or consume events.
...
+through a specific Encode function (e.g. kafka.EncodeProducerMessage), in order to simplify the invocation of the
+encoding message.
+
+The encoding process can be customized in order to mutate the final result with binding.TransformerFactory. A bunch of these are provided directly by the binding/transcoder module.
+
+Usually binding.Message implementations can be encoded only one time, because the encoding process drain the message itself.
+In order to consume a message several times, the binding/buffering module provides several APIs to buffer the Message.
 
-Receiver: receives protocol messages and wraps them to implement the Message interface.
+A message can be converted to an event.Event using ToEvent method. An event.Event can be used as Message casting it to binding.EventMessage.
```
n3wscott: My feeling is if you intend to use events, bindings are the wrong level to interact with this SDK.

pkg/binding/event_message.go
```diff
@@ -63,7 +63,7 @@ func (m *EventMessage) SetEvent(e event.Event) error {
 
 var _ Message = (*EventMessage)(nil) // Test it conforms to the interface
 
-func EventContextToBinaryEncoder(c event.EventContext, b BinaryEncoder) (err error) {
+func eventContextToBinaryEncoder(c event.EventContext, b BinaryEncoder) (err error) {
```
n3wscott: what does this method do?

pkg/binding/format/doc.go
```diff
@@ -0,0 +1,8 @@
+package format
+
+/*
+Package format formats structured events.
```
n3wscott: only structured?

pkg/binding/format/doc.go
```diff
@@ -0,0 +1,8 @@
...
+
+The "application/cloudevents+json" format is built-in and always
```
n3wscott: this is only how http talks about this format, not the rest of the protocols.

pkg/binding/message.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Structured transfers a structured-mode event to a StructuredEncoder.
+	// It must return ErrNotStructured if message is not in structured mode.
+	//
+	// Returns a different err if something wrong happened while trying to read the structured event.
+	// In this case, the caller must Finish the message with appropriate error.
+	//
+	// This allows Senders to avoid re-encoding messages that are
+	// already in suitable structured form.
+	Structured(context.Context, StructuredEncoder) error
```
n3wscott: I have tried but I have no idea how to use this method or the binary method because I do not know how to make a StructuredEncoder.

pkg/binding/structured_encoder.go
```diff
@@ -0,0 +1,17 @@
...
+	"io"
+
+	"github.com/cloudevents/sdk-go/pkg/binding/format"
+)
+
+// StructuredEncoder is used to visit a structured Message and generate a new representation.
+//
+// Protocols that supports structured encoding should implement this interface to implement direct
+// structured to structured encoding and event to structured encoding.
+type StructuredEncoder interface {
```
n3wscott: This would not be an encoder. It seems like this is reading an already encoded event from the io.Reader and applying it to a message.

I feel like encoders and decoders should only be for translating between events and messages, not protocol and message

pkg/binding/doc.go
slinkydeveloper: Agree, but this should be documented somewhere, eg to explain to a binding implementer how to implement a `Transport` starting from a binding implementation

pkg/binding/event_message.go
slinkydeveloper: It's an helper to set attributes/extensions of the event context to the `BinaryEncoder`

pkg/binding/format/doc.go
slinkydeveloper: yes, this package is intended to provide the primitives to decode/encode messages in structured mode

pkg/binding/format/doc.go
slinkydeveloper: As far as I see from the spec, all protocol bindings which has both the structured and binary mode, uses `application/cloudevents+json` as content-type to define a cloudevent json encoded:

* https://github.com/cloudevents/spec/blob/master/amqp-protocol-binding.md#321-amqp-content-type
* https://github.com/cloudevents/spec/blob/master/amqp-protocol-binding.md#321-amqp-content-type
* https://github.com/cloudevents/spec/blob/master/kafka-protocol-binding.md#331-kafka-content-type
* https://github.com/cloudevents/spec/blob/master/mqtt-protocol-binding.md#321-mqtt-content-type

pkg/binding/message.go
slinkydeveloper: Both `StructuredEncoder` and `BinaryEncoder` are provided by the specific binding implementation. Their functionality is usually "served" through an `Encode` method, like:

* https://github.com/cloudevents/sdk-go/blob/master/pkg/bindings/kafka_sarama/producer_message_encoder.go#L23
* https://github.com/cloudevents/sdk-go/blob/master/pkg/bindings/amqp/encoder.go#L18
* https://github.com/cloudevents/sdk-go/blob/master/pkg/bindings/http/response_encoder.go#L17
* https://github.com/cloudevents/sdk-go/blob/master/pkg/bindings/http/request_encoder.go#L17

So the bindings user don't need to invoke them directly

pkg/binding/structured_encoder.go
slinkydeveloper: Ok that's another proof that the `Encoder` name doesn't sound suitable for `StructuredEncoder` and `BinaryEncoder`. Those are two visitors which are given to messages to visit them, and the visit mutates the encoder itself

n3wscott:
LGTM

PR #348, opened 2020-03-05
Removed specs.FindVersion()

### LINKED ISSUE #330: Prune v 0.1 & v 0.2
Prune Cloudevents versions 0.1 & 0.2

## Bindings specific

This includes removing the "hack" to extract the event version in https://github.com/cloudevents/sdk-go/blob/master/pkg/binding/spec/spec.go#L61 & https://github.com/cloudevents/sdk-go/blob/master/pkg/binding/spec/spec.go#L68, which now could be simplified to:

```
func (vs *Versions) SpecVersionName() string
```

And the caller just does:

```
if v, err := versions.Version(messageHeaders[versions.SpecVersionName()]); err == nil {
  // I have a spec version
}
```

slinkydeveloper:
Closes #330


n3wscott:
LGTM

PR #349, opened 2020-03-06
Implement Httpb

n3wscott:
follow up to #344 

This uses the new requester impl in http and uses the BindingTransport to implement a first cut http binding transport.


pkg/binding/transport.go
```diff
@@ -11,58 +13,100 @@ import (
...
+		return ctx, nil, t.Sender.Send(ctx, EventMessage(e))
+	}
+	// If provided a requester, use it to do request/response.
+	var resp *event.Event
+	msg, err := t.Requester.Request(ctx, EventMessage(e))
+	if err == nil {
+		if rs, _, err := ToEvent(ctx, msg, nil); err != nil {
+			cecontext.LoggerFrom(ctx).Warnw("failed calling ToEvent", zap.Error(err))
+		} else {
+			defer func() {
```
slinkydeveloper: `message.Finish()` should be called anyway, even if `ToEvent` failed, you should move this defer under `Request` statement

pkg/binding/transport.go
```diff
@@ -11,58 +13,100 @@ import (
...
 	e, _, err := ToEvent(ctx, m, nil)
 	if err != nil {
 		return err
 	}
-	return t.handler.Receive(ctx, e, nil)
+	eventResp := event.EventResponse{}
+	if err := t.handler.Receive(ctx, e, &eventResp); err != nil {
+		return err
+	}
+	// TODO: do something with eventResp
```
slinkydeveloper: And that's where we have a problem: this `BindingAdapter` is fine as soon as the protocol doesn't have a request/response model.
IMO the `StartReceiver` should be overriden by the `httpb` transport to support the full request/response handling, so you still reuse this adapter, but you support the full request/response model

pkg/binding/transport.go
```diff
@@ -11,58 +13,100 @@ import (
...
+			}()
+			resp = &rs
+		}
+	}
+	return ctx, resp, err
 }
 
-func (t *BindingTransport) SetReceiver(r transport.Receiver) { t.handler = r }
+func (t *BindingTransport) SetReceiver(r transport.Receiver) {
+	t.handler = r
```
slinkydeveloper: Look this comment: https://github.com/cloudevents/sdk-go/pull/343/files#r388793129

pkg/cloudevents/transport/httpb/options.go
```diff
@@ -0,0 +1,281 @@
...
+			return nil
+		}
+		return fmt.Errorf("http target option was empty string")
+	}
+}
+
+//
+//// WithMethod sets the HTTP verb (GET, POST, PUT, etc.) to use
+//// when using an HTTP request.
+//func WithMethod(method string) Option {
```
slinkydeveloper: I think that `WithMethod` and `WithHeader` should be supported directly by `http.Sender`, so here you should just configure the `http.Sender`

pkg/cloudevents/transport/httpb/transport.go
```diff
@@ -0,0 +1,238 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// Send implements Transport.Send
+func (t *Transport) Send(ctx context.Context, e event.Event) (context.Context, *event.Event, error) {
+	switch t.Encoding {
+	case Default, Binary:
```
slinkydeveloper: What if, for default, you don't decorate at all the context and let the binding try the best encoding?

pkg/binding/transport.go
n3wscott: ok I was not sure about that, thanks!

pkg/binding/transport.go
n3wscott: This is all temp dead code anyway.

n3wscott: like... this whole file

pkg/cloudevents/transport/httpb/transport.go
n3wscott: the transport has been set to a mode as a supported feature in cloudevents.Transport. This plumbs that feature through

slinkydeveloper:
LGTM

Issue #350, opened 2020-03-06
Client Mutates Events When Modifying Extensions


ian-mi:
Instead of modifying the caller's copy the client should clone the event context before setting event extensions.

PR #333, opened 2020-03-03
Fixed #332

slinkydeveloper:
Fixed #332, now the `Clone()` method copies the extensions map

Signed-off-by: Francesco Guardiani <francescoguard@gmail.com>

PR #351, opened 2020-03-06
Clone event context before setting the distributed tracing extension

ian-mi:
This prevents the client from mutating the caller's copy of the event
allowing for sharing across multiple clients. Also includes a cherry pick of PR #333 to fix event context Clone()

Backports fix for #350.


n3wscott:
LGTM

PR #352, opened 2020-03-06
Clone event context before setting the distributed tracing extension

### LINKED ISSUE #350: Client Mutates Events When Modifying Extensions

ian-mi:
This prevents the client from mutating the caller's copy of the event
allowing for sharing across multiple clients.

Fixes #350


n3wscott:
LGTM

PR #353, opened 2020-03-06
Now EventMessage can be plugged to use different marshalling formats

### LINKED ISSUE #345: binding.EventMessage support for different formats in Structured method
`binding.EventMessage` now defaults to `JSON` to convert the `EventMessage` back to an `io.Reader`. We should add support for all formats registered in `binding.format` package through the provided context.

Ref: https://github.com/cloudevents/sdk-go/blob/master/pkg/binding/event_message.go#L26

slinkydeveloper:
Fix #345 


n3wscott:
LGTM

PR #355, opened 2020-03-06
Add Client.Request, rework Client.Send

n3wscott:
As part of the work for v2, this reworks the client interface to be more inline with how Bindings work and how real integrations with previous transports work. 

For almost all protocol integrations besides http, it does not make sense to support a request. This changes the client api to make it clear what the user is attempting to do with Client:

- `Send` --> sends an event
- `Request` --> sends an event and returns an event in response.


pkg/binding/transport.go
```diff
@@ -27,11 +28,15 @@ func NewTransportAdapter(s Sender, r Receiver, senderContextDecorators []func(co
...
+func (t *BindingTransport) Send(ctx context.Context, e event.Event) error {
 	for _, f := range t.SenderContextDecorators {
 		ctx = f(ctx)
 	}
-	return ctx, nil, t.Sender.Send(ctx, EventMessage(e))
+	return t.Sender.Send(ctx, EventMessage(e))
+}
+
+func (t *BindingTransport) Request(ctx context.Context, e event.Event) (*event.Event, error) {
+	return nil, errors.New("Transport.Request is not yet supported by binding transport.")
```
slinkydeveloper: If you meant to use this func as default impl for this method, maybe rename it to `Transport.Request is not supported by this transport.`

n3wscott: In a followup I am implementing this. The error is the detail we want for now to find where it is not implemented. All the transport parts will be removed anyway.

jackedelic:
Hi Scott, this commit helps me tremendously because I have a service that sends a cloudevent but does not need a cloudevent as the response. Hope you don't mind I ask a small question here:
Is the change a part of the CloudEvent specification? I browse the doc [cloudevent spec](https://github.com/cloudevents/spec/blob/v1.0/primer.md#eventing) but I'm unable to gather information that explicitly mentions the "correct" format/type for CloudEvent "response". My guess is that the CloudEvent defines unidirectional behaviour of data transmission, rather than bidirectional. 

I'm not sure if you're the correct person to ask about the official behaviour of CloudEvent. But thanks for your time, and your contribution to the SDK.

Best,
Jack

btw, i believe your email [snichols@vmware.com](mailto:snichols@vmware.com) might not be in use, not sure what's the best/appropriate way to reach out to you regarding small question like this.

jackedelic:
Hi Scott @n3wscott , this commit helps me tremendously because I have a service that sends a cloudevent but does not need a cloudevent as the response. Hope you don't mind I ask a small question here:
Is the change a part of the CloudEvent specification? I browse the doc [cloudevent spec](https://github.com/cloudevents/spec/blob/v1.0/primer.md#eventing) but I'm unable to gather information that explicitly mentions the "correct" format/type for CloudEvent "response". My guess is that the CloudEvent defines unidirectional behaviour of data transmission, rather than bidirectional.

I'm not sure if you're the correct person to ask about the official behaviour of CloudEvent. But thanks for your time, and your contribution to the SDK.

Best,
Jack

btw, i believe your email [snichols@vmware.com](mailto:snichols@vmware.com) might not be in use, not sure what's the best/appropriate way to reach out to you regarding small question like this.

n3wscott:
hello @jackedelic! 

The client that is defined in this sdk is not part of the cloudevents spec as it does not detail the runtime characteristics. But I felt it was too hard to use without an opinion, so the client is an opinion on the spec that uses the underlying bindings which are defined by the spec.

If you need to break glass, you can use the transport interfaces and control what the transport is doing in much the same way the client is doing. There are some examples of integrations like this: https://github.com/knative/eventing and some of the other parts in that community like https://github.com/knative-sandbox/eventing-kafka-broker

I hope that helps, my new email address is n3wscott at chainguard.dev

-Scott

PR #361, opened 2020-03-09
Drop pkg/cloudevents and refactor Transport.

n3wscott:
This starts merging the transport and bindings packages. 

- ToEvent returns a pointer. Move Sender/Receiver out of Binding.
- Drop support for Pub/Sub for lack of public CloudEvents spec.
- Migrate to Binding from Codec.
- Drop transport event version overwriting. Pass the event version you wish to be sent.
- Drop WithMethod http transport control, use the Request Template directly.
- Drop HttpEncoding Selector function support.
- Drop Long Polling support.
- Drop support for http context based encoding.
- Drop transport context from client.Receive, use the transport directly.
- Drop ContextWithHeader support. Use transport directly.
- Drop http metrics for now, we will add back.
- Drop Converter function support, use and control the binding or transport directly.
- Migrate NATS to Bindings


pkg/binding/write.go
```diff
@@ -86,13 +86,13 @@ func Write(
...
 
-	var e event.Event
+	var e *event.Event
 	e, err = ToEvent(ctx, message, transformers)
 	if err != nil {
 		return enc, err
 	}
 
-	message = EventMessage(e)
+	message = EventMessage(*e)
```
slinkydeveloper: copy

pkg/transport/test/transformer.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/stretchr/testify/require"
```
slinkydeveloper: This file should be moved back to binding/test

pkg/transport/transport.go
```diff
@@ -0,0 +1,40 @@
...
+	// in-band trace propagation. When false, the client receiver
+	// will propagate trace context from distributed tracing
+	// extension attributes when available.
+	HasTracePropagation() bool
+}
+
+// REMOVE:
+// Receiver is an interface to define how a transport will invoke a listener
+// of incoming events.
+// Deprecated: use bindings.
```
slinkydeveloper: Use bindings?!

alias.go
```diff
@@ -57,19 +58,8 @@ const (
...
-	HTTPBatchedV1     = http.BatchedV1
-	HTTPBinaryV03     = http.BinaryV03
-	HTTPStructuredV03 = http.StructuredV03
-	HTTPBatchedV03    = http.BatchedV03
-
-	// Context HTTP Transport Encodings
-
-	Binary     = http.Binary
-	Structured = http.Structured
+	HTTPBinaryEncoding     = httpb.Binary
```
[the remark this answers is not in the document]
n3wscott: I think it should be not the responsibility of the binding to change the version of the event.

pkg/transport/transport.go
n3wscott: 🤣 there is a little cleanup to do

pkg/transport/test/transformer.go
n3wscott: moved back.

nachocano:
Really?
With such a big change, how come you didn't add more interested parties in the review? 
@n3wscott you basically removed all PubSub support in the SDK. I'm still a bit shocked. How come you didn't add at least myself to the review. I've done some contributions to this code...
We agreed in slack *yesterday* to give us a few days so that we could get the PubSub binding spec in, and then work on the necessary code refactor here. 
Even if you were planning on doing this for a long time, how come you just told me *yesterday*. From a person that used to sit next to you, I'd have expected some _camaraderie_.

n3wscott:
@nachocano I removed Pub/Sub in v2 of the sdk as I did not have the time to migrate it to the new Bindings. Also there is no published spec for Pub/Sub. I understand your side will work on publishing a spec, and at that time we will be happy to accept a PR to re-introduce Pub/Sub as a supported protocol.

We are trying to move fast on v2, aiming a release at the end of March. We have decided the layers introduced in v1 of the SDK were slightly off, and consistently piping features between the client and the protocol has proven to be not desirable for low level integrators and too many knobs for high level integrators. Separating the protocol binding from a reference protocol impl is the goal here.

We have created a branch for current integrators to fix bugs or continue to add features in the v1 sdk style: https://github.com/cloudevents/sdk-go/tree/release-1.y.z and have been doing point releases off that for 1.x.x

I did not ask for your review because you have not been involved and using the rubric of no public spec and no mention of Pub/Sub in CloudEvents, it was decided that Pub/Sub could be dropped until a spec and a new implementation are produced.

We are aiming for a v2.0.0-rc1 preview release laster this week.

PR #364, opened 2020-03-10
Update HTTP to combine Sender and Receiver into a single impl.

n3wscott:
This combines HTTP Sender and Requester, Receiver into a new thing called Protocol that leverages Bindings.

This makes a new Transport object that handles all the http server details. This is to enable Transport for legacy integration that will be removed before the real 2.0 cut.


n3wscott:
> Ok so this transport/protocol difference will disappear in future PRs correct? 

I do not know what they will be called exactly, Protocol does not handle the setting up and running of additional components required to allow for you to just use the protocol ServeHTTP method, which lets you wire this up to any other http lib (like gorilla).

So there is a component from us that shows how to use the protocol, and it is required that something run that infra for the client, I think we can continue to tease them apart.

PR #368, opened 2020-03-11
Event clone

### LINKED ISSUE #366: event.Event should have a clone function
`event.Event` doesn't have a function to perform a deep cloning of the full event.

@n3wscott any thoughts on this?

slinkydeveloper:
Fix #366

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -143,7 +143,9 @@ func (e Event) Clone() Event {
 }
 
 func (e Event) cloneData() interface{} {
-	if bytes, ok := e.Data.([]byte); ok {
+	if e.Data == nil {
+		return nil
+	} else if bytes, ok := e.Data.([]byte); ok {
 		new := make([]byte, len(bytes))
 		copy(new, bytes)
 		return new
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -9,7 +9,8 @@ import (
 
 // Event represents the canonical representation of a CloudEvent.
 type Event struct {
-	Context     EventContext
+	Context EventContext
+	// Data can be a string or a []byte
 	Data        interface{}
 	DataEncoded bool
 	DataBinary  bool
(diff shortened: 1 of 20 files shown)


pkg/event/event.go
```diff
@@ -130,3 +131,27 @@ func (e Event) String() string {
...
 	return b.String()
 }
+
+func (e Event) Clone() Event {
+	new := Event{}
+	new.Context = e.Context.Clone()
+	new.DataBinary = e.DataBinary
+	new.DataEncoded = e.DataEncoded
+	new.Data = e.cloneData()
+	new.FieldErrors = e.FieldErrors
```
n3wscott: The map is a pointer too, you will have to deep copy it too

https://play.golang.org/p/ptXhUuQrdXl

pkg/event/event.go
```diff
@@ -9,7 +9,8 @@ import (
 
 // Event represents the canonical representation of a CloudEvent.
 type Event struct {
-	Context     EventContext
+	Context EventContext
+	// Data can be a string or a []byte
```
n3wscott: Data can also be a struct, map, or array. Really, it can be anything that we can call json.Marshal on

pkg/event/event.go
```diff
@@ -130,3 +131,27 @@ func (e Event) String() string {
...
+	if e.Data == nil {
+		return nil
+	} else if bytes, ok := e.Data.([]byte); ok {
+		new := make([]byte, len(bytes))
+		copy(new, bytes)
+		return new
+	} else if s, ok := e.Data.(string); ok {
+		return s // Strings are immutable!
+	} else {
+		panic("Invalid value in Event.Data field")
```
n3wscott: this is only true for `e.DataEncoded = true`

pkg/event/event.go
slinkydeveloper: Damn I forgot it

pkg/event/event.go
slinkydeveloper: Ah that was not clear to me, how should i implement a deep copy for them?

n3wscott: you will hate this... but I think the surest way in go is to round trip it though json.Mashal/Unmarshal, but that only works if the objects are intended as json.

n3wscott:
Lets let this in with known issues and do what we talked about with DataEncoding, etc. I will make an issue and get that done today

n3wscott:
LGTM with a follow up task on #370

slinkydeveloper:
let's wait for it, is not urgent, i just want it landed before the milestone release

n3wscott:
LGTM

PR #370, opened 2020-03-11
Adding Responder interface, removing ResponseMessage

n3wscott:


pkg/binding/message.go
```diff
@@ -90,14 +90,6 @@ type ExactlyOnceMessage interface {
 	Received(settle func(error))
 }
 
-// ResponseMessage
```
slinkydeveloper: :birthday: :gift:

PR #372, opened 2020-03-11
Rework how responses are supported in Client

n3wscott:
This is a big breaking change to the Client interface, highlights:

- Removal of event.EventResponse
- Overloading of the returned error, it is now event.Result which can signal for protocols to do something special to Ack or Nack a request of response.
    - error and event.Result are equivalent in a receiver method.

The full receive method signature is changing from:

```
func(context.Context, event.Event, *event.EventResponse) error
```

to:

```
func(context.Context, event.Event) (*event.Event, event.Result)
```

This change brings Client much more inline with what the protocol interfaces are. The next step will be to remove the transport.Transport interface in favor of something more direct.


pkg/transport/http/message_write_responsewriter.go
```diff
@@ -15,15 +16,24 @@ import (
 
 // Write out to the the provided httpResponseWriter with the message m.
 // Using context you can tweak the encoding processing (more details on binding.Write documentation).
-func WriteResponseWriter(ctx context.Context, m binding.Message, rw http.ResponseWriter, transformers binding.TransformerFactories) error {
-	structuredWriter := &httpResponseWriterEncoder{rw: rw}
-	binaryWriter := &httpResponseWriterEncoder{rw: rw}
+func WriteResponseWriter(ctx context.Context, m binding.Message, er event.Result, rw http.ResponseWriter, transformers binding.TransformerFactories) error {
+	status := http.StatusOK
+	if er != nil {
```
slinkydeveloper: Are you sure this logic should live here? I see this logic more "client related": IMO this function should take `statusCode` as parameter, and the caller performs the transformation `event.Result` to status code

n3wscott: the way the `ResponseWriter` works is if you call `rw.Write(body)` and no headers have been set, it finalizes the headers with the default `http.StatusOK`. So unlike the other messages in other protocols, the ResponseWriter is an active write out the wire. There is no chance to call a `.Send` or `.Do` style method.

n3wscott:
> I think you should also implement the check when the `Receiver` is set up, to return an error when the used protocol doesn't support request/response semantic

that was todays task to work out. It is hard at the moment because the transport interface is too general for the new interfaces.

slinkydeveloper:
>that was todays task to work out. It is hard at the moment because the transport interface is too general for the new interfaces.

Ok we figure it out today

PR #376, opened 2020-03-12
Rework Event.SetData, require data content encoding

### LINKED ISSUE #371: it never makes sense to set event data without datacontent type
we currently have a race condition with setting the data inside the event where converting the data to the encoded form requires that datacontenttype is set. 

Proposed solution:

Change `event.SetData(interface{})` to `event.SetData(interface{}, contenttype`
Drop `event.SetContentType()`

This forces data to always be set with the content type.
This disallows changing content type on the fly, independent of setting the data.

As a side effect, we will drop event.Data. Event will have the following form from:

```go
type Event struct {
	Context     EventContext
	Data        interface{}
	DataEncoded bool
	DataBinary  bool
	FieldErrors map[string]error
}
```

To:

```go
type Event struct {
	Context     EventContext
	DataEncoded []byte
	DataBinary  []byte
	FieldErrors map[string]error
}
```

n3wscott:
fixes #371 

- Changed `event.SetData(interface{})` to `event.SetData(interface{}, contenttype)`
- Kept ~~Drop~~ `event.SetContentType()`

This forces data to always be set with the content type.

I did not disallow changing content type on the fly, because if the user wants to corrupt the outbound request, then ... I guess that is what they wanted to do.

As a side effect, we will drop event.Data. Event will have the following form FROM:

```go
type Event struct {
	Context     EventContext
	Data        interface{}
	DataEncoded bool
	DataBinary  bool
	FieldErrors map[string]error
}
```

TO:

```go
type Event struct {
	Context     EventContext
	DataEncoded []byte
	DataBinary  bool
	FieldErrors map[string]error
}
```


pkg/binding/to_event.go
```diff
@@ -87,7 +87,7 @@ func (b *messageToEventBuilder) SetData(data io.Reader) error {
 		return err
 	}
 	if w != 0 {
-		return b.event.SetData(buf.Bytes())
+		b.event.DataEncoded = buf.Bytes() // TODO: this might be wrong for base64.
```
slinkydeveloper: I think you can access the `ContentType` directly and use `SetData`, like:

```
ct := b.event.GetDataContentType()
// Maybe check if ct == "" and apply a default value
b.event.SetData(ct, buf.Bytes())
```

slinkydeveloper:
Also, if you prefer, I like more `SetData(contentType, data)`

slinkydeveloper:
Also fixes #366

pkg/binding/to_event.go
n3wscott: that will not do what you want because  `buf.Bytes()` is a `[]byte` and SetData will think you want base64 encoded data out

slinkydeveloper: Ok i'm fine leaving as is

PR #394, opened 2020-03-17
[WIP] Change Import Version to github.com/cloudevents/sdk-go/v2

ian-mi:
This is necessary to satisfy the go module import compability rule: https://blog.golang.org/v2-go-modules.

> If an old package and a new package have the same import path, the new package must be backwards compatible with the old package.


ian-mi:
Fixed by https://github.com/cloudevents/sdk-go/pull/395

PR #409, opened 2020-03-20
Adding unit tests for http.

n3wscott:
More to add, but this was able to uncover a few corner cases where we could get stuck.


n3wscott:
Not done, added a bunch but still working on more coverage..

PR #412, opened 2020-03-20
Fixed various Message.Finish() invocations

slinkydeveloper:
The integration tests that uses `protocol/test.SendReceive` now correctly test this behaviour

### DIFF
--- a/v2/protocol/amqp/sender.go
+++ b/v2/protocol/amqp/sender.go
@@ -19,7 +19,8 @@ func (s *sender) Send(ctx context.Context, in binding.Message) error {
 	var err error
 	defer func() { _ = in.Finish(err) }()
 	if m, ok := in.(*Message); ok { // Already an AMQP message.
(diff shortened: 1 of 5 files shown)


v2/protocol/amqp/sender.go
```diff
@@ -28,7 +28,8 @@ func (s *sender) Send(ctx context.Context, in binding.Message) error {
 		return err
 	}
 
-	return s.amqp.Send(ctx, &amqpMessage)
+	err = s.amqp.Send(ctx, &amqpMessage)
```
n3wscott: why this change?

slinkydeveloper: if error happened during `Send`, the `err` is used by the defer call

n3wscott: oooh to access inside the Finish.

n3wscott: line 22 needs the same change.

Issue #316, opened 2020-02-27
NATS transport, unable to reuse connections for multiple subscriptions


dan-j:
If an application wishes to subscribe to multiple NATS subjects it must create separate `Transport`s which means creating new connections. 

This could be mitigated in two ways:

1. Change the signature to `nats.New(conn *nats.Conn, queue string, opts ...Option) ...`. This would be a breaking change so would have to be in the v2 release
2. Add a `WithConn(conn *nats.Conn) Option` option which if set doesn't call `t.connect()` in the constructor, this isn't a breaking change, but makes the API a little less clear.

Alternative options that I am not in favour of:
- Create the struct manually - there's a fair bit of setup in the constructor
- Use the constructor but then do `t.Conn = anotherConn` - this adds overhead of starting multiple connections unnecessarily.

n3wscott:
I did not know much about nats when I wrote the original version... So if you have insight please do come help.

dan-j:
This should follow the similar pattern to kafka (and STAN when it's merged) protocol where we we have `New` and `NewFromConn` constructors

slinkydeveloper:
@dan-j Are you willing to contribute for it?

dan-j:
Yeah, will be after STAN gets tests/examples and is merged though.

dan-j:
Fixed by #415

Issue #258, opened 2019-11-22
Support for NATS Queue Subscriptions


dan-j:
Not properly thought about a solution for this yet, but is it something you'd accept a PR for?

The appropriate docs can be found here: https://docs.nats.io/developing-with-nats/receiving/queues

In essence, a queue subscription tells NATS to load balance messages to multiple subscribers. Whereas the current NATS Transport only allows traditional subscribers, if two instances connect with the same configuration, messages will be sent to both clients.

Like I said, I haven't really thought through how this would be implemented, but willing to have a discussion and contribute a PR 😄

n3wscott:
I wrote the current impl not being a current NATS user. PLEASE help me make it better! 👍

dan-j:
I'll need this functionality in a few weeks, so I will either get round to it then, or might try and whip something up sooner if I get some free time.

rob0t7:
Funny that this is exactly the same functionality that I also need to ensure that the message is not processed by more than 1 worker in the same group. 

@dan-j any progress or PR on this issue?

dan-j:
I have implemented the same thing for NATS Streaming (aka STAN) in PR #322, I need to add tests and examples before it can get merged, then I'll be back to focus on NATS. I'll more than likely address this at the same time as #316. 

I probably won't get to any NATS stuff until w/c 30th though. If you need it more urgently, you can see how I've implemented it in #322 and follow the same pattern?

ripienaar:
As an aside we (NATS) will in the next few months release our next gen streaming platform called JetStream, I'll revist here then to add support for that when ready

dan-j:
Fixed by #415

zhaojizhuang:
> As an aside we (NATS) will in the next few months release our next gen streaming platform called JetStream, I'll revist here then to add support for that when ready

@ripienaar Is there any progress in support for JetStream

ripienaar:
Jetstream is out and released but have not had a time to revisit integrations

Issue #257, opened 2019-11-22
Unable to tell when NATs Transport is ready


dan-j:
Similar to #131 but for NATS rather than HTTP.

I only started learning Go last week, so I apologise if I'm missing anything obvious.

Is there any specific reason why `StartListener` is a blocking call? It's documented within NATS `transport.go` that it's a blocking call, but the documentation on the `Transport` interface doesn't specify.

https://github.com/cloudevents/sdk-go/blob/73126cd029da7aed7b2ca33d5f4b135d8e66d051/pkg/cloudevents/transport/nats/transport.go#L116-L118

https://github.com/cloudevents/sdk-go/blob/73126cd029da7aed7b2ca33d5f4b135d8e66d051/pkg/cloudevents/transport/transport.go#L11-L22

I'd suggest changing the implementation to return after the subscription is created, something like this?

```go
t.sub, err = t.Conn.Subscribe(t.Subject, func(m *nats.Msg) {
    ...
})

go func() {
	<-ctx.Done()
	if sub != nil {
		err := sub.Unsubscribe()
		if err != nil {
			// we're in a goroutine, just log the error
		}
	}
}()

return err
```

This would be a breaking change, but I notice you haven't actually made a tagged release for this feature yet?

n3wscott:
In general it is a blocking call so that you can do lifecycle management using the go context and not have to pass around stop channels. I think what you have suggested could be how the transport manages its self or adding some options to help with these lifecycle events.

dan-j:
> and not have to pass around stop channels

^^ this is my point in the opposite argument, you'd have to provide a start channel which emits once the subscriber is ready.

I will look at making a separate PR as the same time as I tackle #258, as it's not much work to implement, and we can have a proper discussion once we have some actual code to review if that works for you?

dan-j:
Just to clarify as well. I have only just started learning Go, so I'm open to being told my way isn't "the Go way", just so long as some alternative solution to the issue is agreed ☺️

dan-j:
Just about to pick this back up and notice that v1.0.0 has been released which now makes the proposed solution a breaking change.

What's the thoughts on this? An alternative solution is to provide an additional transport `Option` to invoke a callback when the transport has established the subscription, but I'd prefer the initial solution.

slinkydeveloper:
Is this one still valid with the latest refactoring of nats from transport to the new protocol?

dan-j:
[...]

PR #415, opened 2020-03-22
NATS protocol improvements

dan-j:
Addresses:

- #258: You can create Consumers with the `WithQueueSubscriber("queue-name")` option
- #257: With the v2 refactor, you no longer need to wait for subscriptions to be created because calls to `Receive(ctx)` will block until either messages are received or `ctx` is closed
- #316: Follows the same pattern as kafka and STAN, with two constructors for one with an existing connection and one without.

Added integration tests for NATS, and updated the sample cmds

### DIFF
--- a/v2/cmd/samples/httptonats/main.go
+++ b/v2/cmd/samples/httptonats/main.go
@@ -31,7 +31,7 @@ func main() {
 	}
 	ctx := context.Background()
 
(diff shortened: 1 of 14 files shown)


dan-j:
Oops, I've branched off my STAN branch.. hold with me while I rebase

n3wscott:
Awesome!

LGTM

Issue #417, opened 2020-03-23
Make a channel based protocol binding.


n3wscott:
It would be nice if we could send events between objects via channels in go as a client interaction.

slinkydeveloper:
Somehow this already exists :smile: https://github.com/cloudevents/sdk-go/blob/master/v2/protocol/test/chan_receiver.go

slinkydeveloper:
We just need to make it less "internal" probably

n3wscott:
oh yeah...fml I just found it.

```
// ChanSender implements Sender by sending on a channel.
type ChanSender chan<- Message

// ChanReceiver implements Receiver by receiving from a channel.
type ChanReceiver <-chan Message
```

PR #418, opened 2020-03-23
Implement gochan protocol.

n3wscott:
fixes: https://github.com/cloudevents/sdk-go/issues/417


n3wscott:
> The code of sender and receiver already lives in pkg/protocol/test. can you remove the bits in test?

Done.

PR #419, opened 2020-03-23
wip: play around with stan lifecycle

dan-j:
do not merge, here just to share a diff

### DIFF
--- a/v2/protocol/stan/receiver.go
+++ b/v2/protocol/stan/receiver.go
@@ -6,6 +6,7 @@ import (
 	"github.com/cloudevents/sdk-go/v2/binding"
 	"github.com/nats-io/stan.go"
 	"io"
+	"sync"
 )
 
 type msgErr struct {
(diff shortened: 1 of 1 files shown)


dan-j:
Oops, was meant to create a PR local to my fork, will leave it here for discussion anyway 🤔

dan-j:
@slinkydeveloper I was thinking about our discussion on slack with closing subscriptions vs connections etc and got this prototype running. I need to add tests (and fix the ones I've broken) before it's anywhere near mergeable (hence meaning to create it on my fork), but thought it would be worth your input.

Essentially:

1. `Consumer.OpenInbound()` sets up the subscription and creates an internal channel `subCloser`, and will pause until either `subCloser` or `ctx` is closed.
1. `Consumer.Close()` checks whether or not `subCloser != nil` (i.e. the consumer is running) and if it is, calls `close(subCloser)` which will cause `OpenInbound` to stop.
1. `Consumer.Close()` only closes the underlying connection if it created it in the constructor. (Just spotted a race condition here, but the high-level details are the only important thing for now)

slinkydeveloper:
Is this still valid?

dan-j:
No, I don't think it is valid anymore. I think we answered the questions around who's responsible for closing connections/protocols/client. Will close

PR #420, opened 2020-03-26
Fix a race condition in the distributed tracing unit tests

ian-mi:
Also fix the following bugs which were uncovered as a result:
1. The sampled flag would not be correctly set when creating
SpanContext
2. Tracestate parsing would fail and be omitted from SpanContext


n3wscott:
LGTM

PR #421, opened 2020-03-26
fix: missing coma in the last example

owulveryck:
This is preventing a copy/paste from working out-of-the-box


n3wscott:
whoops! can you sign the commit and then we can merge, thanks for the fix!!

n3wscott:
Fixed, thanks!

PR #424, opened 2020-03-27
Fixed ResponseWriter Content-Length and transformers

slinkydeveloper:
Before this fix, the `ResponseWriter` was writing headers *after* `WriteHeader` which, in unit tests works because `ResponseRecorder` is bugged in golang <= 1.13 (https://go-review.googlesource.com/c/go/+/178058/) and we were using it wrongly (we should perform asserts on `ResponseRecorder.Result()`).

Now the content-length is best effort: if the `io.Reader` is a reader to buffered data, content-length is written, otherwise it's not

### DIFF
--- a/v2/protocol/http/write_responsewriter.go
+++ b/v2/protocol/http/write_responsewriter.go
@@ -1,10 +1,12 @@
 package http
 
 import (
+	"bytes"
 	"context"
 	"io"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/cloudevents/sdk-go/v2/binding"
 	"github.com/cloudevents/sdk-go/v2/binding/format"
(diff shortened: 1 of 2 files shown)


v2/protocol/http/write_responsewriter.go
```diff
@@ -86,5 +73,46 @@ func (b *httpResponseWriter) SetExtension(name string, value interface{}) error
 	return nil
 }
 
+func (b *httpResponseWriter) SetData(reader io.Reader) error {
+	b.body = reader
+	return nil
+}
+
+func (b *httpResponseWriter) finalizeWriter() error {
```
n3wscott: hm, similar code to the request it seems

n3wscott:
LGTM

PR #428, opened 2020-03-30
Reworking datacodec package

slinkydeveloper:
With this PR `datacodec` package doesn't handle the base64 encoding, this is left to the event marshaller/unmarshaller which injects into the `Data` field directly the decoded byte array.

Also improves documentation of `SetData` and `Data` explaining the various behaviours.

### DIFF
--- a/v2/event/event_marshal_test.go
+++ b/v2/event/event_marshal_test.go
@@ -66,7 +66,7 @@ func TestMarshal(t *testing.T) {
 				"exurl":    source,
 				"extime":   &now,
 			},
-			want: toBytes(map[string]interface{}{
+			want: jsonToBytes(map[string]interface{}{
 				"specversion":     "0.3",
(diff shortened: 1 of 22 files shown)


v2/event/event_marshal_test.go
```diff
@@ -66,7 +66,7 @@ func TestMarshal(t *testing.T) {
 				"exurl":    source,
 				"extime":   &now,
 			},
-			want: toBytes(map[string]interface{}{
```
n3wscott: In this file we need to add a test for each of the other encodings the sdk supports

n3wscott:
LGTM

We need to do a pass at xml testing, I am seeing a regression in the sdk (unrelated to this PR)

n3wscott:
nice. 

LGTM, thanks for cleaning up.

Issue #429, opened 2020-03-31
Pubsub protocol needs to implement responder so that messages can be properly acked/nacked


yolocs:
Currently the code directly acks a message once received (https://github.com/cloudevents/sdk-go/blob/master/v2/protocol/pubsub/protocol.go#L171). This is wrong because if the receiver fails to handle the message, the message will be lost.

yolocs:
@nlopezgi

PR #430, opened 2020-03-31
Fix Pubsub protocol to properly ack/nack received messages

yolocs:
Fix: https://github.com/cloudevents/sdk-go/issues/429

### DIFF
--- a/v2/protocol/pubsub/protocol.go
+++ b/v2/protocol/pubsub/protocol.go
@@ -167,9 +167,7 @@ func (t *Protocol) Receive(ctx context.Context) (binding.Message, error) {
 		return nil, io.EOF
 	}
 
-	msg := NewMessage(m.Data, m.Attributes)
-	m.Ack()
-	// TODO: when to do m.Nack()?
+	msg := NewMessage(&m)
 	return msg, nil
 }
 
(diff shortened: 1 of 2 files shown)


yolocs:
@nlopezgi

nlopezgi:
I think you need to sign your commit (e.g.,  `git commit -a -s --amend`) to get DCO check to pass

v2/protocol/pubsub/protocol.go
```diff
@@ -160,17 +160,27 @@ func (t *Protocol) getOrCreateConnection(ctx context.Context, topic, subscriptio
 	return conn
 }
 
-// Receive implements Receiver.Receive
-func (t *Protocol) Receive(ctx context.Context) (binding.Message, error) {
+// Respond implements Responder.Respond. This response function doesn't support reply.
```
n3wscott: Rather than do this, I would revert back to what you had, but change the channel:

```incoming chan pubsub.Message```

to be a Message and an Error, and there you can see if it should be a ACK or NACK from that error.

yolocs: Thanks for the pointer!

n3wscott:
LGTM!

PR #432, opened 2020-04-01
Fix Package Documentation and Enable Documentation Check in CircleCI

ian-mi:
Many packages were not correctly formatting package documentation either because the documentation did not precede the package statement as required or the first line did not correctly reference the package name. A few packages were missing documentation. To prevent regressions this PR enables staticcheck ST1000 to check that every v2 package has valid package docs.


slinkydeveloper:
thanks!

PR #435, opened 2020-04-02
Fix and Enable All Staticcheck Checks besides ST1003

ian-mi:
Adds a staticcheck.conf under v2/ to skip ST1003 since many identifiers violate the recommended practices. These include package names with underscores, variables names written in all caps, and uncapitalized initialisms such as Http.


slinkydeveloper:
Thanks for that!

PR #436, opened 2020-04-02
 add support for HTTP retries

lionelvillard:
Add support for retrying HTTP requests when a certain class of error occurs. 

There is no tests yet. Opening the PR to get early feedback

@n3wscott

### DIFF
--- a/v2/context/backoff.go
+++ b/v2/context/backoff.go
@@ -1,4 +1,4 @@
-package types
+package context
 
 import "time"
 
--- a/v2/context/context.go
+++ b/v2/context/context.go
@@ -3,6 +3,7 @@ package context
 import (
 	"context"
 	"net/url"
+	"time"
 
 	"github.com/cloudevents/sdk-go/v2/types"
 )
@@ -53,23 +54,28 @@ func TopicFrom(ctx context.Context) string {
 	return ""
 }
 
-// Opaque key type used to store backoff parameters
-type retryKeyType struct{}
+// Opaque key type used to store linear backoff parameters
+type linearBackoffKeyType struct{}
 
-var retryKey = retryKeyType{}
+var linearBackoffKey = linearBackoffKeyType{}
 
-// WithRetry returns back a new context with the retry backoff parameters.
-func WithRetry(ctx context.Context, backoff types.Backoff) context.Context {
-	return context.WithValue(ctx, retryKey, backoff)
+// WithLinearBackoff returns back a new context with the linear backoff parameters.
+// MaxTries is the maximum number for retries and delay is the time interval between retries
+func WithLinearBackoff(ctx context.Context, delay time.Duration, maxTries int) context.Context {
+	return context.WithValue(ctx, linearBackoffKey, types.Backoff{
+		Strategy: types.BackoffStrategyLinear,
+		Period:   delay,
+		MaxTries: maxTries,
+	})
 }
 
-// RetryFrom looks in the given context and returns the backoff parameters if found, otherwise nil
-func RetryFrom(ctx context.Context) *types.Backoff {
-	c := ctx.Value(retryKey)
+// RetryFrom looks in the given context and returns the linear backoff parameters if found, otherwise 0,0
+func LinearBackoffFrom(ctx context.Context) (time.Duration, int) {
+	c := ctx.Value(linearBackoffKey)
 	if c != nil {
 		if s, ok := c.(types.Backoff); ok {
-			return &s
+			return s.Period, s.MaxTries
 		}
 	}
-	return nil
+	return time.Duration(0), 0
 }
--- a/v2/context/context.go
+++ b/v2/context/context.go
@@ -3,6 +3,8 @@ package context
 import (
 	"context"
 	"net/url"
+
+	"github.com/cloudevents/sdk-go/v2/types"
 )
 
 // Opaque key type used to store target
@@ -49,4 +51,25 @@ func TopicFrom(ctx context.Context) string {
 		}
 	}
 	return ""
+}
+
+// Opaque key type used to store backoff parameters
+type retryKeyType struct{}
+
+var retryKey = retryKeyType{}
+
+// WithRetry returns back a new context with the retry backoff parameters.
+func WithRetry(ctx context.Context, backoff types.Backoff) context.Context {
+	return context.WithValue(ctx, retryKey, backoff)
+}
+
+// RetryFrom looks in the given context and returns the backoff parameters if found, otherwise nil
+func RetryFrom(ctx context.Context) *types.Backoff {
+	c := ctx.Value(retryKey)
+	if c != nil {
+		if s, ok := c.(types.Backoff); ok {
+			return &s
+		}
+	}
+	return nil
 }
--- a/v2/context/context.go
+++ b/v2/context/context.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"net/url"
 	"time"
-
-	"github.com/cloudevents/sdk-go/v2/types"
 )
 
 // Opaque key type used to store target
@@ -62,8 +60,8 @@ var linearBackoffKey = linearBackoffKeyType{}
 // WithLinearBackoff returns back a new context with the linear backoff parameters.
 // MaxTries is the maximum number for retries and delay is the time interval between retries
 func WithLinearBackoff(ctx context.Context, delay time.Duration, maxTries int) context.Context {
-	return context.WithValue(ctx, linearBackoffKey, types.Backoff{
-		Strategy: types.BackoffStrategyLinear,
+	return context.WithValue(ctx, linearBackoffKey, Backoff{
+		Strategy: BackoffStrategyLinear,
 		Period:   delay,
 		MaxTries: maxTries,
 	})
@@ -73,7 +71,7 @@ func WithLinearBackoff(ctx context.Context, delay time.Duration, maxTries int) c
 func LinearBackoffFrom(ctx context.Context) (time.Duration, int) {
 	c := ctx.Value(linearBackoffKey)
 	if c != nil {
-		if s, ok := c.(types.Backoff); ok {
+		if s, ok := c.(Backoff); ok {
 			return s.Period, s.MaxTries
 		}
 	}
--- a/v2/context/context.go
+++ b/v2/context/context.go
@@ -52,28 +52,29 @@ func TopicFrom(ctx context.Context) string {
 	return ""
 }
 
-// Opaque key type used to store linear backoff parameters
-type linearBackoffKeyType struct{}
+// Opaque key type used to store retry parameters
+type retriesKeyType struct{}
 
-var linearBackoffKey = linearBackoffKeyType{}
+var retriesKey = retriesKeyType{}
 
-// WithLinearBackoff returns back a new context with the linear backoff parameters.
+// WithRetriesLinearBackoff returns back a new context with retries parameters using linear backoff strategy.
 // MaxTries is the maximum number for retries and delay is the time interval between retries
-func WithLinearBackoff(ctx context.Context, delay time.Duration, maxTries int) context.Context {
-	return context.WithValue(ctx, linearBackoffKey, Backoff{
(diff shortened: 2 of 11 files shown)


v2/protocol/http/protocol.go
```diff
@@ -118,6 +120,16 @@ func (p *Protocol) Request(ctx context.Context, m binding.Message) (binding.Mess
 	if err = WriteRequest(ctx, m, req, p.transformers); err != nil {
 		return nil, err
 	}
+
+	do := p.doOnce
```
n3wscott, line 124: Oh nice. I like.

n3wscott, line 124: Oh nice. I like.

v2/protocol/http/retries_result.go
```diff
@@ -0,0 +1,46 @@
...
+	return protocol.ResultIs(e.Result, target)
+}
+
+// Error returns the string that is formed by using the format string with the
+// provided args.
+func (e *RetriesResult) Error() string {
+	if e.Retries == 0 {
+		return e.Result.Error()
+	}
+	return fmt.Sprintf("%s (%dx)", e.Result.Error(), e.Retries)
```
n3wscott, line 45: This might do funky things with the ack wrapping so a test for this is needed

n3wscott, line 45: This might do funky things with the ack wrapping so a test for this is needed

v2/protocol/http/retries_result.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+// Result wraps the fields required to make adjustments for http Responses.
+type RetriesResult struct {
+	// The last result
+	protocol.Result
+
+	// Retries is the number of times the request was tried
+	Retries int
```
n3wscott, line 25: I think we want to return back the total time spent trying too if we do exp backoff, what do you think?

v2/types/backoff.go
```diff
@@ -0,0 +1,14 @@
+package types
+
+import "time"
+
+// Backoff holds parameters applied to retries
+type Backoff struct {
```
n3wscott, line 6: Rather than make a custom type, I would hide that detail from the user with a change to the WithBackoff method, I will comment there.

v2/context/context.go
```diff
@@ -50,3 +52,24 @@ func TopicFrom(ctx context.Context) string {
...
+
+var retryKey = retryKeyType{}
+
+// WithRetry returns back a new context with the retry backoff parameters.
+func WithRetry(ctx context.Context, backoff types.Backoff) context.Context {
+	return context.WithValue(ctx, retryKey, backoff)
+}
+
+// RetryFrom looks in the given context and returns the backoff parameters if found, otherwise nil
+func RetryFrom(ctx context.Context) *types.Backoff {
```
n3wscott, line 67: I know slinky recommended moving this to http but there are settings in other protocols that have similar settings, I feel like it is ok to keep it in the general spot here.

v2/context/context.go
```diff
@@ -50,3 +52,24 @@ func TopicFrom(ctx context.Context) string {
...
 	return ""
 }
+
+// Opaque key type used to store backoff parameters
+type retryKeyType struct{}
+
+var retryKey = retryKeyType{}
+
+// WithRetry returns back a new context with the retry backoff parameters.
+func WithRetry(ctx context.Context, backoff types.Backoff) context.Context {
```
n3wscott, line 62: We might want more than one strategy for backoff, this is linear and we might also want exponential backoff.

Also, rather than a custom type (or an internal custom type is ok) I would not expose that to the user. I recommend being explicit about the backoff and retries:

```
func WithLinearBackoff(ctx context.Context, period time.Duration, maxTries uint) context.Context {
```

is one way. I think internally we should make a type, it can be defined here in this file:

```go
type BackoffConfig struct {
    Strategy string
    Period time.Duration
    MaxRetries int
}
```

And WithLinearBackoff makes a `BackoffConfig{Strategy:"linear", ...`. 

I suggest this to let the api self documenting and not require the caller to understand another object to make.

v2/protocol/http/protocol.go
```diff
@@ -133,6 +145,64 @@ func (p *Protocol) Request(ctx context.Context, m binding.Message) (binding.Mess
...
+
+			if !retry {
+				return NewMessage(resp.Header, resp.Body), NewRetriesResult(result, retries, results)
+			}
+
+			results = append(results, result)
+			retries++
+
+			// Linear backoff.
+			time.Sleep(backoff.Delay)
```
n3wscott, line 201: I would make this method more of a wrapper for doOnce with a some kind of pattern of:


```go
for { 
  select {
    case <-ctx.Done():
    // canceled.
    case <-timer.Tick():
      strongTypedErr: doOnce(ctx, request)
      ... custom logic to see if we should stop retry
  }
}
```

That way the logic of producing the `protocol.NewReceipt` is only one place, and you are still free to collect them in the outer retry method.

v2/protocol/http/protocol.go
```diff
@@ -133,6 +145,64 @@ func (p *Protocol) Request(ctx context.Context, m binding.Message) (binding.Mess
 	return NewMessage(resp.Header, resp.Body), NewResult(resp.StatusCode, "%w", result)
 }
 
+func (p *Protocol) doWithRetry(backoff *types.Backoff) func(*http.Request) (binding.Message, error) {
```
n3wscott, line 148: I think this needs context.Context passed in to allow for the app to cancel in the case of shutdown or term signals.

v2/protocol/http/retries_result.go
lionelvillard: to make sure we are on the same page, this is the time spent trying *excluding* the succesful (if any) request, right. I'm pretty sure the answer is yes...

v2/context/context.go
lionelvillard: sound good. Fixing....

v2/context/context.go
slinkydeveloper: Ok i'm fine with that

v2/protocol/http/protocol.go
lionelvillard: good point! thanks!

v2/protocol/http/protocol.go
lionelvillard: I tried but the reason I didn't call `doOnce` is because of `if retry && !err.(*url.Error).Timeout() {`. When `doOnce` returns an error, there is no way to unwrap the original error.

lionelvillard: which can be easily solved by changing `doOnce` signature.

lionelvillard: ended up trying go 1.13 unwrap API. Not tested yet.

lionelvillard:
this is ready for a second round of review. I added some tests under `v2/protocol/http`. I didn't see `v2/test`. I'm happy to move the tests over there if needed.

@n3wscott @slinkydeveloper

lionelvillard:
@slinkydeveloper there is no `content` _package_. Do you mean `context`?

slinkydeveloper:
yep, `context` package

v2/context/context.go
```diff
@@ -50,3 +51,29 @@ func TopicFrom(ctx context.Context) string {
...
+func WithLinearBackoff(ctx context.Context, delay time.Duration, maxTries int) context.Context {
+	return context.WithValue(ctx, linearBackoffKey, Backoff{
+		Strategy: BackoffStrategyLinear,
+		Period:   delay,
+		MaxTries: maxTries,
+	})
+}
+
+// RetryFrom looks in the given context and returns the linear backoff parameters if found, otherwise 0,0
+func LinearBackoffFrom(ctx context.Context) (time.Duration, int) {
```
n3wscott, line 71: I would have this return a `Backoff` struct.

v2/context/context.go
```diff
@@ -50,3 +51,29 @@ func TopicFrom(ctx context.Context) string {
 	}
 	return ""
 }
+
+// Opaque key type used to store linear backoff parameters
+type linearBackoffKeyType struct{}
+
+var linearBackoffKey = linearBackoffKeyType{}
```
n3wscott, line 58: this should just be a `backoffKey`

v2/context/context.go
```diff
@@ -50,3 +51,29 @@ func TopicFrom(ctx context.Context) string {
...
+
+// WithLinearBackoff returns back a new context with the linear backoff parameters.
+// MaxTries is the maximum number for retries and delay is the time interval between retries
+func WithLinearBackoff(ctx context.Context, delay time.Duration, maxTries int) context.Context {
+	return context.WithValue(ctx, linearBackoffKey, Backoff{
+		Strategy: BackoffStrategyLinear,
```
n3wscott, line 64: Exactly. The goal here is if we add a new strategy, it will override any previously set strategies. 

So down the road when we make ExpBackoff we can make a method called `WithExponentialBackoff` and all the plumbing is done.

The aim with `WithLinearBackoff` is to keep the consumer api simple and they do not have to be aware of the details of the Backoff Struct

v2/protocol/http/protocol.go
```diff
@@ -118,6 +118,16 @@ func (p *Protocol) Request(ctx context.Context, m binding.Message) (binding.Mess
 	if err = WriteRequest(ctx, m, req, p.transformers); err != nil {
 		return nil, err
 	}
+
+	do := p.doOnce
+	if delay, tries := cecontext.LinearBackoffFrom(ctx); tries > 0 {
```
n3wscott, line 123: I would change this method to `BackoffFrom` and return the `Backoff` struct, then have a switch for `BackoffStrategyLinear` or `BackoffStrategyNone`

v2/context/context.go
n3wscott: Also renamed to `BackoffFrom`

v2/context/context.go
```diff
@@ -50,3 +51,29 @@ func TopicFrom(ctx context.Context) string {
...
+
+// RetryFrom looks in the given context and returns the linear backoff parameters if found, otherwise 0,0
+func LinearBackoffFrom(ctx context.Context) (time.Duration, int) {
+	c := ctx.Value(linearBackoffKey)
+	if c != nil {
+		if s, ok := c.(Backoff); ok {
+			return s.Period, s.MaxTries
+		}
+	}
+	return time.Duration(0), 0
```
n3wscott, line 78: This could return a Backoff with strategy: `BackoffStrategyNone`. And then your test in the http impl becomes a simple flow that always returns a `Backoff`

v2/context/context.go
n3wscott: `Backoff` seems to be confusing. I would rename to `Retry` or `Retires`

v2/context/context.go
n3wscott: would `WithRetiesLinearBackoff` make more sense?

v2/context/backoff.go
```diff
@@ -0,0 +1,24 @@
...
+
+type BackoffStrategy string
+
+const (
+	BackoffStrategyLinear      = "linear"
+	BackoffStrategyExponential = "exponential"
+)
+
+// Backoff holds parameters applied to retries
+type Backoff struct {
```
n3wscott, line 13: Perhaps this should be called `Retries` or `RetryParams`?

v2/context/context.go
lionelvillard: I updated the comment but not the code. Distracted me. I'll go with Retries

lionelvillard: fixed.

v2/context/context.go
lionelvillard: using `retriesKey`.

v2/context/context.go
lionelvillard: done.

v2/context/context.go
lionelvillard: Added and modified impl.

v2/context/backoff.go
lionelvillard: yes make sense. Done.

n3wscott:
LGTM!!

Thanks, this is gonna be awesome!!

PR #437, opened 2020-04-06
Kafka Key Mapping implementation

slinkydeveloper:
Also splitted a part of `binding.Write` to `binding.WriteEvent`

I have implemented the mapping from `partitionkey` to Kafka Message key, but I didn't **intentionally** added any custom key mapper function, because I think this adds unnecessary complexity to the method `WriteProducerMessage`.
If the user wants a custom mapping, he can do it by itself by taking the message, converting to event, doing whatever he wants to calculate the key from the event, and then pass the event message to the `WriteProducerMessage`. Maybe there is room for the custom key mapper in the kafka sender?

Let's keep this on hold until https://github.com/cloudevents/spec/pull/599 is merged.

### DIFF
--- a/v2/protocol/kafka_sarama/write_producer_message.go
+++ b/v2/protocol/kafka_sarama/write_producer_message.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/Shopify/sarama"
 
-	ce "github.com/cloudevents/sdk-go/v2"
 	"github.com/cloudevents/sdk-go/v2/binding"
 	"github.com/cloudevents/sdk-go/v2/binding/format"
 	"github.com/cloudevents/sdk-go/v2/binding/spec"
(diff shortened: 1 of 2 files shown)


v2/binding/write.go
```diff
@@ -59,10 +59,45 @@ func DirectWrite(
...
+func WriteEvent(
+	ctx context.Context,
+	e *event.Event,
+	structuredWriter StructuredWriter,
+	binaryWriter BinaryWriter,
+) (Encoding, error) {
+	message := (*EventMessage)(e)
+
+	if GetOrDefaultFromCtx(ctx, preferredEventEncoding, EncodingBinary).(Encoding) == EncodingStructured {
+		if structuredWriter != nil {
```
n3wscott, line 79: it bothers me that the order of the binary and structured are different.

slinkydeveloper: WDYM? it's just choosing between which of the two is preferred

n3wscott:
LGTM

v2/binding/write.go
n3wscott: ooh I see

slinkydeveloper:
Going forward merging this one, since https://github.com/cloudevents/spec/pull/599 it's merged

PR #442, opened 2020-04-06
Adding Exp and Const backoff for good measure.

n3wscott:

### DIFF
--- a/v2/alias.go
+++ b/v2/alias.go
@@ -96,10 +96,15 @@ var (
 	WithTimeNow          = client.WithTimeNow
 	WithTracePropagation = client.WithTracePropagation()
 
+	// Event Creation
+
+	NewEvent = event.New
+
 	// Results
 
-	ResultIs = protocol.ResultIs
-	ResultAs = protocol.ResultAs
+	NewResult = protocol.NewResult
+	ResultIs  = protocol.ResultIs
+	ResultAs  = protocol.ResultAs
 
 	// Receipt helpers
 
@@ -111,13 +116,8 @@ var (
 	IsACK  = protocol.IsACK
 	IsNACK = protocol.IsNACK
 
-	// Event Creation
+	// HTTP Results
 
-	NewEvent = event.New
-
-	// Results
-
-	NewResult            = protocol.NewResult
 	NewHTTPResult        = http.NewResult
 	NewHTTPRetriesResult = http.NewRetriesResult
 
@@ -135,11 +135,14 @@ var (
 
 	// Context
 
-	ContextWithTarget               = context.WithTarget
-	TargetFromContext               = context.TargetFrom
-	ContextWithRetriesLinearBackoff = context.WithRetriesLinearBackoff
-	WithEncodingBinary              = binding.WithForceBinary
-	WithEncodingStructured          = binding.WithForceStructured
+	ContextWithTarget                    = context.WithTarget
+	TargetFromContext                    = context.TargetFrom
+	ContextWithRetriesConstantBackoff    = context.WithRetriesConstantBackoff
+	ContextWithRetriesLinearBackoff      = context.WithRetriesLinearBackoff
+	ContextWithRetriesExponentialBackoff = context.WithRetriesExponentialBackoff
+
+	WithEncodingBinary     = binding.WithForceBinary
+	WithEncodingStructured = binding.WithForceStructured
 
 	// Custom Types
 
--- a/v2/protocol/http/protocol_retry.go
+++ b/v2/protocol/http/protocol_retry.go
@@ -1,0 +1,123 @@
+package http
+
+import (
+	"context"
+	"errors"
+	"go.uber.org/zap"
+	"net/http"
+	"net/url"
+	"time"
+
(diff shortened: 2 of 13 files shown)


v2/alias.go
```diff
@@ -135,11 +135,14 @@ var (
...
-	ContextWithTarget               = context.WithTarget
-	TargetFromContext               = context.TargetFrom
-	ContextWithRetriesLinearBackoff = context.WithRetriesLinearBackoff
-	WithEncodingBinary              = binding.WithForceBinary
-	WithEncodingStructured          = binding.WithForceStructured
+	ContextWithTarget                    = context.WithTarget
+	TargetFromContext                    = context.TargetFrom
+	ContextWithRetriesConstantBackoff    = context.WithRetriesConstantBackoff
+	ContextWithRetriesLinearBackoff      = context.WithRetriesLinearBackoff
+	ContextWithRetriesExponentialBackoff = context.WithRetriesExponentialBackoff
```
slinkydeveloper, line 142: growing and growing :smile:

v2/protocol/http/protocol_retry.go
```diff
@@ -0,0 +1,123 @@
...
+			//		// system level error, retry.
+			//		goto DO_BACKOFF
+			//	}
+			//}
+
+			{
+				var uErr *url.Error
+				if errors.As(result, &uErr) {
+					// Only retry when the error is a timeout
+					//if uErr.Timeout() || uErr.Temporary() {
```
slinkydeveloper, line 76: This seems reasonable to me. eg: If the hostname is invalid, there's no point on retrying

v2/protocol/http/protocol_retry.go
```diff
@@ -0,0 +1,123 @@
...
+				return msg, NewRetriesResult(result, retry, then, results)
+			}
+
+			// Try again?
+			//
+			// Make sure the error was something we should retry.
+
+			//{
+			//	var sErr *syscall.Errno
+			//	if errors.As(result, &sErr) {
```
slinkydeveloper, line 66: I think we should not care of the underlying syscall error, because we already check the `*url.Error`

v2/protocol/http/protocol_retry.go
```diff
@@ -0,0 +1,123 @@
...
+				}
+			}
+
+		DO_BACKOFF:
+			// Wait for the correct amount of backoff time.
+
+			// total tries = retry + 1
+			if err := params.Backoff(ctx, retry+1); err != nil {
+				// do not try again.
+				cecontext.LoggerFrom(ctx).Debugw("backoff error, will not try again", zap.Error(err))
```
slinkydeveloper, line 115: Can we get the logger once at the beginning of this function?

v2/protocol/http/protocol_retry.go
```diff
@@ -0,0 +1,123 @@
...
+		result = protocol.ResultACK
+	} else {
+		result = protocol.ResultNACK
+	}
+
+	return NewMessage(resp.Header, resp.Body), NewResult(resp.StatusCode, "%w", result)
+}
+
+func (p *Protocol) doWithRetry(ctx context.Context, params *cecontext.RetryParams) func(*http.Request) (binding.Message, error) {
+	return func(req *http.Request) (binding.Message, error) {
```
slinkydeveloper, line 46: Ok I think i'm missing something, but why this method `doWithRetry` returns a function with the logic more than just executing the logic directly?

v2/protocol/http/protocol_retry.go
n3wscott: it might change. There is no reason not to try. Like, take k8s, if the service is not up but will be soon... keep trying.

n3wscott: We should not get to this point with an invalid hostname because we use url.URL parsed objects

v2/protocol/http/protocol_retry.go
n3wscott: no need, just pull if from context if you need it

v2/protocol/http/protocol_retry.go
n3wscott: hm, right. :D fixed.

v2/protocol/http/protocol_retry.go
n3wscott: removed.

v2/alias.go
n3wscott: it is better than all the source in a single dir

slinkydeveloper: yeah but before the v2 i wish to review it to check what we need and what not

v2/protocol/http/protocol_retry.go
slinkydeveloper: if the service is not up i assume that `Error.Temporary() == true` right? like unable to connect and stuff like that is temporary

slinkydeveloper:
LGTM

v2/protocol/http/protocol_retry.go
n3wscott: no, if the service is not up it is fatal. that is only temporary in k8s :D

PR #443, opened 2020-04-06
Add ExtractExtensions TransformerFactory

ian-mi:
ExtractExtensions is a TransformerFactory which extracts a set of
extensions from transformed mesages. An instance of this transformer
should only be used to process a single message, after which the
extracted extensions can be read from the map. All extension which are
present as keys in the map will be extracted and other extensions will
be ignored.

Signed-off-by: Ian Milligan <ianmllgn@gmail.com>

### DIFF
--- a/v2/binding/transformer/extract_metadata_test.go
+++ b/v2/binding/transformer/extract_metadata_test.go
@@ -1,0 +1,124 @@
+package transformer
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/cloudevents/sdk-go/v2/binding"
+	"github.com/cloudevents/sdk-go/v2/binding/test"
+	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/stretchr/testify/assert"
+)
+
+func ExampleExtractExtensions() {
+	e := event.New("1.0")
+	e.Context = &event.EventContextV1{}
+	e.SetID("id")
+	e.SetSource("source")
+	e.SetType("type")
+	e.SetExtension("someextension", "somevalue")
+	e.SetExtension("otherextension", "othervalue") // extension will be ignored
+	extractor := ExtractExtensions{"someextension": nil}
+	_, err := binding.ToEvent(context.TODO(), binding.ToMessage(&e), extractor)
+	if err != nil {
+		fmt.Printf("Error running extractor: %v", err)
+	}
+	fmt.Println("extractor:", extractor)
+	//Output:
+	//extractor: map[someextension:somevalue]
+}
+
+func TestExtractExtensions(t *testing.T) {
+	t.Parallel()
+
+	e := test.MinEvent()
+	e.Context = e.Context.AsV1()
+	e.SetExtension("extension1", "extension1-val")
+	e.SetExtension("extension2", false)
+
+	tcs := []struct {
+		name      string
+		event     *event.Event
+		extractor ExtractExtensions
+		want      ExtractExtensions
+	}{
+		{
+			name:  "no extensions extracted",
+			event: &e,
+			extractor: ExtractExtensions{
+				"extension3": nil,
+			},
+			want: ExtractExtensions{
+				"extension3": nil,
+			},
+		},
+		{
+			name:  "one extension extracted",
+			event: &e,
+			extractor: ExtractExtensions{
+				"extension3": nil,
+				"extension1": nil,
+			},
+			want: ExtractExtensions{
+				"extension1": "extension1-val",
+				"extension3": nil,
+			},
+		},
+		{
+			name:  "two extensions extracted",
+			event: &e,
+			extractor: ExtractExtensions{
+				"extension1": nil,
+				"extension2": nil,
+				"extension3": nil,
+			},
+			want: ExtractExtensions{
+				"extension1": "extension1-val",
+				"extension2": false,
+				"extension3": nil,
+			},
+		},
+	}
+	for _, tc := range tcs {
+		eventExtractor := make(ExtractExtensions)
+		binaryExtractor := make(ExtractExtensions)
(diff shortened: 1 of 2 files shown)


v2/binding/transformer/extract_metadata_test.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		fmt.Printf("Error running extractor: %v", err)
+	}
+	fmt.Println("extractor:", extractor)
+	//Output:
+	//extractor: map[someextension:somevalue]
+}
+
+func TestExtractExtensions(t *testing.T) {
+	t.Parallel()
```
slinkydeveloper, line 33: You don't need this

v2/binding/transformer/extract_metadata_test.go
```diff
@@ -0,0 +1,124 @@
...
+			},
+			{
+				Name:         "structured",
+				InputMessage: test.MustCreateMockStructuredMessage(*tc.event),
+				AssertFunc:   assertExtractExtensions(tc.event, structuredExtractor, tc.want),
+				Transformers: []binding.TransformerFactory{structuredExtractor},
+			},
+		}
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
```
slinkydeveloper, line 113: You don't need this

v2/binding/transformer/extract_metadata_test.go
```diff
@@ -0,0 +1,124 @@
...
+			test.RunTransformerTests(t, context.TODO(), testArgs)
```
slinkydeveloper, line 114: Just a small comment about test organization: can you put this `RunTransformerTests` outside the loop? prepare all testArgs, then just invoke the run test one time without wrapping `t.Run` again. This makes clearer the view of the test cases in IDEs

ian-mi:
This should be superior in the case that you need to gather multiple extensions (which I will need to for tracing). Otherwise you would need to use multiple factories adding additional nested calls to every writer call as well as performing multiple string comparisons on every SetExtension.

slinkydeveloper:
> Otherwise you would need to use multiple factories adding additional nested calls to every writer call as well as performing multiple string comparisons on every SetExtension.

Yeah my guess is that string comparisons are cheaper, but i wanna try it before reaching a conclusion

slinkydeveloper:
So I've tried to create a `ExtractExtension` of my own with single extension extraction. old is this patch and new is my patch:

```
benchmark                                   old ns/op     new ns/op     delta
BenchmarkHttpWithToEvent-8                  21728         20512         -5.60%
BenchmarkNoDataHttpWithToEvent-8            20200         20828         +3.11%
BenchmarkHttpWithBuffering-8                11823         11926         +0.87%
BenchmarkHttpWithDirect-8                   9911          9692          -2.21%
BenchmarkNoDataHttpWithBuffering-8          11737         11645         -0.78%
BenchmarkNoDataHttpWithDirect-8             9728          9668          -0.62%

benchmark                                   old allocs     new allocs     delta
BenchmarkHttpWithToEvent-8                  151            151            +0.00%
BenchmarkNoDataHttpWithToEvent-8            145            145            +0.00%
BenchmarkHttpWithBuffering-8                121            122            +0.83%
BenchmarkHttpWithDirect-8                   114            115            +0.88%
BenchmarkNoDataHttpWithBuffering-8          119            120            +0.84%
BenchmarkNoDataHttpWithDirect-8             112            113            +0.89%

benchmark                                   old bytes     new bytes     delta
BenchmarkHttpWithToEvent-8                  10500         10496         -0.04%
BenchmarkNoDataHttpWithToEvent-8            9864          9864          +0.00%
BenchmarkHttpWithBuffering-8                6086          6109          +0.38%
BenchmarkHttpWithDirect-8                   4760          4791          +0.65%
BenchmarkNoDataHttpWithBuffering-8          6038          6068          +0.50%
BenchmarkNoDataHttpWithDirect-8             4712          4744          +0.68%

```

Interesting enough, having this map allocates less, but in some cases it consumes a little bit more in terms of cpu time (but, honestly to me that delta looks like measurement error).

So yeah, let's go with the map. Please apply the changes to the tests and then we merge it

v2/binding/transformer/extract_metadata_test.go
ian-mi: Why not enable parallel execution?

v2/binding/transformer/extract_metadata_test.go
ian-mi: I don't use an IDE which is bothered by this so I'm not sure what the difference is, but I like having the different variations grouped as subtests of the same test case.

v2/binding/transformer/extract_metadata_test.go
slinkydeveloper: Yes but why forcing this from the code? It's enabled by the flag `-parallel`

v2/binding/transformer/extract_metadata_test.go
slinkydeveloper: ok fine

v2/binding/transformer/extract_metadata_test.go
ian-mi: from `go help testflag`:
> -parallel n
>       Allow parallel execution of test functions that call t.Parallel.
>       ...

T.Parallel marks a test as being safe for parallel execution. If it's not called the test will run sequentially with other tests in the same package or, in the case of subtests, with other subtests in the same test. The parallel flag only sets the maximum number of parallel test processes.

slinkydeveloper:
I think this one is invalid now. Because now it's more easy to implement the extraction of extensions with transformers, i close this one in favour of https://github.com/cloudevents/sdk-go/pull/447 which is more targeted to our need

PR #446, opened 2020-04-08
Transformers rework

slinkydeveloper:
## Problem

Transformers are based on a wrong assumption: that you can build a `StructuredTransformer`. This assumption drove to designing transformers as they are now on master. Problem is: having a `StructuredTransformer` doesn't make sense without a buffering step, which makes them useless by definition.

## Solution

This PR proposes to redesign `Transformers` around the assumption that the message I'm transforming is already in some "semi-parsed" state, so I can consequently read cloudevents attributes/extensions from it. To do that, I introduce an interface called `MessageMetadataReader` that every `Message` that supports `Binary`/`Event` mode must implement. This adds the burden of implementing these two methods to the binding implementer, which IMO is negligible.

## Breaking changes

There are also a couple of important changes:

* If a `Message` can be in a binary/event encoding, it **must** implement `MessageMetadataReader`.
* Implementations of `Message.ReadBinary` **must not** control the writer lifecycle (`Start` and `End`)
* `MessageMetadataWriter` extracts two methods from `BinaryWriter`. Implementations of `SetAttribute` and `SetExtension` **must** treat `value == nil` as deletion
* Removed `UpdateAttribute`/`UpdateExtension` (now `SetAttribute`/`SetExtension` covers that use case)

## Performance

Because we removed an allocation on the hot path (transformer factories creating transformers), memory allocations are less:

```
benchmark                                   old ns/op     new ns/op     delta
BenchmarkHttpWithToEvent-8                  16310         15828         -2.96%
BenchmarkNoDataHttpWithToEvent-8            14379         15306         +6.45%
BenchmarkHttpWithBuffering-8                9164          8649          -5.62%
BenchmarkHttpWithDirect-8                   8203          7046          -14.10%
BenchmarkNoDataHttpWithBuffering-8          8691          8523          -1.93%
BenchmarkNoDataHttpWithDirect-8             6675          7113          +6.56%

benchmark                                   old allocs     new allocs     delta
BenchmarkHttpWithToEvent-8                  108            105            -2.78%
BenchmarkNoDataHttpWithToEvent-8            102            99             -2.94%
BenchmarkHttpWithBuffering-8                80             78             -2.50%
BenchmarkHttpWithDirect-8                   73             71             -2.74%
BenchmarkNoDataHttpWithBuffering-8          78             76             -2.56%
BenchmarkNoDataHttpWithDirect-8             71             69             -2.82%

benchmark                                   old bytes     new bytes     delta
BenchmarkHttpWithToEvent-8                  8060          7992          -0.84%
BenchmarkNoDataHttpWithToEvent-8            7435          7358          -1.04%
BenchmarkHttpWithBuffering-8                5499          5326          -3.15%
BenchmarkHttpWithDirect-8                   4177          3981          -4.69%
BenchmarkNoDataHttpWithBuffering-8          5466          5285          -3.31%
BenchmarkNoDataHttpWithDirect-8             4132          3951          -4.38%

```


n3wscott:
LGTM

PR #456, opened 2020-04-14
Fix error message spelling error: unnknown => unknown

grayside:
There are spots in both v1 and v2 marshaling code where `unknown` is spelled with an extra `n`. This PR fixes it in v2 on the assumption that v1 error messages should not be changed.

Skipping filing an issue to save time on overhead, since I do not see a CONTRIBUTING that states a firm rule to always file first, hopefully this works for the community.


slinkydeveloper:
Yes v1 will be removed soon, thank you for this PR!

PR #457, opened 2020-04-15
Now Send/Request/ResponseFn accept transformers

### LINKED ISSUE #451: Accept transformers is Send/Request method signatures
Because transformers are usually per-request, it doesn't make sense to assign them to a Send/Request implementation like here: https://github.com/cloudevents/sdk-go/blob/master/v2/protocol/http/protocol.go#L27

slinkydeveloper:
Fixes #451

Also removes `transformer.SetUUID`

### DIFF
--- a/v2/protocol/http/protocol.go
+++ b/v2/protocol/http/protocol.go
@@ -24,7 +24,6 @@ const (
 type Protocol struct {
 	Target          *url.URL
 	RequestTemplate *http.Request
-	transformers    binding.Transformers
 	Client          *http.Client
 	incoming        chan msgErr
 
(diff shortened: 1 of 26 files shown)


v2/protocol/http/protocol.go
```diff
@@ -24,7 +24,6 @@ const (
 type Protocol struct {
 	Target          *url.URL
 	RequestTemplate *http.Request
-	transformers    binding.Transformers
```
n3wscott, line 27: don't you want some static ones set? or they should always come from the outside?

slinkydeveloper:
@n3wscott they are not meant to be usable from client. If the user wants to perform a trasformation at client level, why bother using transformers? He can directly operate on the `Event`

n3wscott:
Well look how time or Id defaulting works.

slinkydeveloper:
> Well look how time or Id defaulting works.

The client uses the defaulters in `client/defaulters.go`.

I think the UUID defaulter as transformer should be removed frankly

slinkydeveloper:
I mean, the client impl could always use the transformers, not the user directly

PR #459, opened 2020-04-15
 ack manually by the the fn result

ti:
In the most MQ case, we just ack when there is no errors, but the  sdk v2 code does not transfer the ReceiverFn result context to MQ protocol `(m *Message) Finish(err error) error` method, so i add this pull request.

in my case:

```go
// The ReceiverFn
fn := func(ctx context.Context, event cloudevents.Event) protocol.Result {
	// ack or nack by the event handler result
	err := doSomething(event)
	if err != nil {
		return protocol.ResultNACK
	}
	if err == IgnoreError {
		return protocol.ResultACK
	}
	return err
}
c.StartReceiver(ctx, fn)
```

in the message Finish function

```go
// Finish marks if there is no error
	if protocol.IsACK(err) {
		m.internal.Ack()
	} else {
		m.internal.Nack()
	}
	return nil
}

```
the full version of kafka implement: https://github.com/cloudevents/sdk-go/pull/462

### DIFF
--- a/v2/client/invoker.go
+++ b/v2/client/invoker.go
@@ -67,7 +67,7 @@ func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn p
 
 		// protocol can manual ack by the result
 		if respFn == nil {
-			if result != nil {
+			if !protocol.IsACK(result) {
 				err = m.Finish(result)
 				isFinished = true
 			}
--- a/v2/client/invoker.go
+++ b/v2/client/invoker.go
@@ -37,9 +37,12 @@ type receiveInvoker struct {
 }
 
 func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn protocol.ResponseFn) (err error) {
+	var isFinished bool
 	defer func() {
-		if err2 := m.Finish(err); err2 == nil {
-			err = err2
+		if !isFinished {
+			if err2 := m.Finish(err); err2 == nil {
+				err = err2
+			}
 		}
 	}()
 
@@ -61,14 +64,22 @@ func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn p
 				cecontext.LoggerFrom(ctx).Error(fmt.Errorf("cloudevent validation failed on response event: %v, %w", verr, err))
 			}
 		}
-		if respFn != nil {
-			var rm binding.Message
-			if resp != nil {
-				rm = (*binding.EventMessage)(resp)
+		
+		// protocol can manual ack by the result
+		if respFn == nil {
+			if result != nil  {
+				err = m.Finish(result)
+				isFinished = true
 			}
-
-			return respFn(ctx, rm, result) // TODO: there is a chance this never gets called. Is that ok?
+			return
 		}
+
+		var rm binding.Message
+		if resp != nil {
+			rm = (*binding.EventMessage)(resp)
+		}
+
+		return respFn(ctx, rm, result)
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


ti:
@slinkydeveloper is this ok?  

Is there any other reason why it cannot be ack when it is received?

slinkydeveloper:
Can you be more specific on why we need this patch? I don't get the problem honestly :smile:

ti:
> Can you be more specific on why we need this patch? I don't get the problem honestly 😄

In most case, we can not lost message when the event handler have panic or some other errors.

In my case, i want ack by the the ReceiverFn result, if ReceiverFn has some error or panic do not ack.

but i review the code, find the code does not transfer the ReceiverFn result context to MQ protocol, so i add the code.

//more info:
https://github.com/cloudevents/sdk-go/issues/460

slinkydeveloper:
What puzzles me is: there is another protocol/MQ you're trying to implement which has Ack/Nack feature and request/response model? Because in case of HTTP, if the receiver function returns an error, this error is then propagated back (at least, in theory it should as far as i see from the code https://github.com/cloudevents/sdk-go/blob/master/v2/protocol/http/protocol.go#L242)

ti:
> What puzzles me is: Is there another protocol/MQ you're trying to implement which has Ack/Nack feature and request/response model? Because in case of HTTP, if the receiver function returns an error, this error is then propagated back (at least, in theory it should as far as i see from the code https://github.com/cloudevents/sdk-go/blob/master/v2/protocol/http/protocol.go#L242)

I use kafka for cloudevents backend, the problems is also happened on google PubSub, it can not nack the event when Receiver handler have some error.  

The http protocol does not have cache/retry middleware, so it may not suitable for some cases

ti:
in the google PubSub or kafka, it can not get the receive event result, because it does not have responder implement. and also, the responder strategy  is not suitable for "Only Sub Client" cases.

```go
// Finish marks if there is no error
	if protocol.IsACK(err) {
		m.internal.Ack()
	} else {
		m.internal.Nack()
	}
	return nil
}
```

yolocs:
FWIW, I fixed the pubsub nack problem in https://github.com/cloudevents/sdk-go/pull/430

v2/client/invoker.go
```diff
@@ -61,14 +64,22 @@ func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn p
...
 		}
-		if respFn != nil {
-			var rm binding.Message
-			if resp != nil {
-				rm = (*binding.EventMessage)(resp)
+
+		// protocol can manual ack by the result
+		if respFn == nil {
+			if result != nil {
+				err = m.Finish(result)
```
n3wscott, line 71: result could be non-nil and an ACK, and respFn could return back a NACK and will have done the wrong thing here.

v2/client/invoker.go
```diff
@@ -61,14 +64,22 @@ func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn p
...
 			}
 		}
-		if respFn != nil {
-			var rm binding.Message
-			if resp != nil {
-				rm = (*binding.EventMessage)(resp)
+
+		// protocol can manual ack by the result
+		if respFn == nil {
+			if result != nil {
```
n3wscott, line 70: != nil is not quite enough, result is a protocol.Result, and it has to be tested for a NACK directly

ti:
> FWIW, I fixed the pubsub nack problem in #430

No, it is not enough,  the recevier fn result is not returned to your message Finish function

v2/client/invoker.go
ti: ok, this can be  `protocol.IsACK(result)`, it should more safe

yolocs:
> No, it is not enough, the recevier fn result is not returned to your message Finish function

I see. This is fixing client.

ti:
@n3wscott @slinkydeveloper 
refer to the https://github.com/cloudevents/sdk-go/pull/461

This can only be implemented if the execution ReceiverFn result be fed back.

slinkydeveloper:
Double checked now, LGTM

n3wscott:
LGTM

Thanks for fixing this.

PR #462, opened 2020-04-16
kafka receiver ack/nack support

ti:
refer to the https://github.com/cloudevents/sdk-go/pull/459, in many cases, we want Receiver ack/nack by the event handler result, like this:

```go
// The ReceiverFn
fn := func(ctx context.Context, event cloudevents.Event) protocol.Result {
	// ack or nack by the event handler result
	err := doSomething(event)
	if err != nil {
		return protocol.ResultNACK
	}
	if err == IgnoreError {
		return protocol.ResultACK
	}
         // if the err !=nil, just do not ack
	return err
}
c.StartReceiver(ctx, fn)
```

This merge request is the implement for kafka.


slinkydeveloper:
Ok so let's wait first for #459 to merge and then we'll look into this one

slinkydeveloper:
I think you should not create a new struct for kafka message, but you should rely on `WithFinish()` to handle the ack/nack

ti:
The WithFinish method has two disadvantages

1. It repeated with message.Finish
2. Not clear for nack and ack logic than meaage.Finish

The implement will make kafka message just like the PubSub message.

slinkydeveloper:
`WithFinish()` function takes the same error of `Finish`, so you can easily plug the logic there.

I'm worried that this PR adds quite a lot of code.

ti:
Ok, i'll change it

ian-mi:
Please add tests to cover the updated client behavior.

v2/protocol/kafka_sarama/receiver.go
```diff
@@ -48,7 +50,31 @@ func (r *Receiver) ConsumeClaim(session sarama.ConsumerGroupSession, claim saram
 		m := NewMessageFromConsumerMessage(message)
 
 		r.incoming <- msgErr{
-			msg: binding.WithFinish(m, func(err error) { session.MarkMessage(message, "") }),
+			msg: binding.WithFinish(m, func(err error) {
+				if protocol.IsACK(err) {
```
ti, line 54: as @slinkydeveloper say, i move the ack/nack logic to binding.WithFinish. then there just 2 files changes from master

ti:
@slinkydeveloper I have moved the ack/nack logic to `binding.WithFinish`. then there just 2 files changes from master 

https://github.com/cloudevents/sdk-go/pull/462/files

v2/protocol/kafka_sarama/receiver.go
ti: you mean this? https://github.com/cloudevents/sdk-go/pull/464/files

ti:
@slinkydeveloper 
I simplified it， only implemented the ack method is ok。
https://github.com/cloudevents/sdk-go/pull/464/files

slinkydeveloper:
I think we can close it now as it's supersed by #464

PR #463, opened 2020-04-16
Backport: Fix a race condition in the distributed tracing unit tests (#420)

ian-mi:
Also fix the following bugs which were uncovered as a result:
1. The sampled flag would not be correctly set when creating
SpanContext
2. Tracestate parsing would fail and be omitted from SpanContext

Signed-off-by: Ian Milligan <ianmllgn@gmail.com>


n3wscott:
LGTM

PR #466, opened 2020-04-20
Fix v1 gomodule path

chizhg:
According to https://github.com/golang/go/wiki/Modules#semantic-import-versioning:
```
If the module is version v0 or v1, do not include the major version in either the module path or the import path.
```
More details can be found in https://github.com/golang/go/issues/24301#issuecomment-371228664.

If the module path is `github.com/cloudevents/sdk-go/v1`, we won't be able to switch knative/eventing to use Go modules as there will be an error:
```
github.com/cloudevents/sdk-go/v1/cloudevents: module github.com/cloudevents/sdk-go@latest found (v1.1.2), but does not contain package github.com/cloudevents/sdk-go/v1/cloudevents
```
And it seems there is no workaround.

This PR removes `v1` from the module path, and update all import paths.

### DIFF
--- a/v1/alias.go
+++ b/v1/alias.go
@@ -4,12 +4,12 @@ package v1
 // the number of imports for simple HTTP clients.
 
 import (
-	"github.com/cloudevents/sdk-go/v1/cloudevents"
-	"github.com/cloudevents/sdk-go/v1/cloudevents/client"
-	"github.com/cloudevents/sdk-go/v1/cloudevents/context"
-	"github.com/cloudevents/sdk-go/v1/cloudevents/observability"
-	"github.com/cloudevents/sdk-go/v1/cloudevents/transport/http"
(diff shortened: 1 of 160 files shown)


n3wscott:
This fixes it for go mod but breaks it for dep.

ian-mi:
go module code can still use the v1 releases with the correct import path, but we should make sure to tag a new v1 release before switching knative to make sure it has the latest v2 backport.

n3wscott:
> go module code can still use the v1 releases with the correct import path, but we should make sure to tag a new v1 release before switching knative to make sure it has the latest v2 backport.

We should use v1 off the release-1.y.z branch, not master.

ian-mi:
Right, that is what I'm suggesting but I think a new release should be tagged from that branch.

Issue #470, opened 2020-04-23
Pubsub protocol returns empty error when being cancelled


yolocs:
## What's happening?
When a pubsub protocol is stopped by cancelling its context, it still returns an empty error.

## Expected behavior
When intentionally stopped, the pubsub protocol shouldn't return error at all.

## Fix

https://github.com/cloudevents/sdk-go/blob/6ec31e9557787a676442efb2416ebce153f6f8b3/v2/protocol/pubsub/protocol.go#L236
If `errs` is nil, if should return nil rather than `errors.New`.

slinkydeveloper:
@yolocs can you also check out this PR please? https://github.com/cloudevents/sdk-go/pull/458
I reworked all `Close` lifecycles to be consistent between them except pubsub :smile:

yolocs:
@slinkydeveloper ack.  Could you help create an issue and assign it me?

PR #471, opened 2020-04-23
V2 pubsub protocol: don't return error if there is no error

yolocs:
Fixes: https://github.com/cloudevents/sdk-go/issues/470


n3wscott:
LGTM

PR #480, opened 2020-04-28
Return http BadRequest instead of terminating in case of conversion error

### LINKED ISSUE #478: Receiver freezes upon reception of non-CE payloads
(SDK v2)

The receiver seemingly "freezes" when receiving a non-CE JSON payload.

Given this client
```go
func main() {
    cli, _ := cloudevents.NewDefaultClient()
    cli.StartReceiver(ctx, func(cloudevents.Event) {})
}
```

```
$ curl -XPOST -H "Content-Type: application/cloudevents+json" localhost:8080/ -d'{"foo":"bar"}'

// never returns
```

antoineco:
An error occurring during conversion from Message to Event is simply returned http BadRequest instead of causing a termination of the receiver.

```console
$ curl -D- -H "Content-Type: application/cloudevents+json" localhost:8080/ -d'{"foo": "bar"}'
HTTP/1.1 400 Bad Request
Date: Wed, 29 Apr 2020 17:14:52 GMT
Content-Length: 0
```

Closes #478

### DIFF
--- a/v2/client/invoker.go
+++ b/v2/client/invoker.go
@@ -2,10 +2,10 @@ package client
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/cloudevents/sdk-go/v2/binding"
 	cecontext "github.com/cloudevents/sdk-go/v2/context"
+	"github.com/cloudevents/sdk-go/v2/event"
 	"github.com/cloudevents/sdk-go/v2/protocol"
 )
 
@@ -37,22 +37,26 @@ type receiveInvoker struct {
 }
 
 func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn protocol.ResponseFn) (err error) {
-	var isFinished bool
 	defer func() {
-		if !isFinished {
-			if err2 := m.Finish(err); err2 == nil {
-				err = err2
-			}
-		}
+		err = m.Finish(err)
 	}()
 
-	e, err := binding.ToEvent(ctx, m)
-	if err != nil {
-		return err
-	}
+	var respMsg binding.Message
+	var result protocol.Result
 
-	if e != nil && r.fn != nil {
-		resp, result := r.fn.invoke(ctx, *e)
+	e, eventErr := binding.ToEvent(ctx, m)
+	switch {
+	case eventErr != nil:
+		result = protocol.NewReceipt(false, "failed to convert Message to Event: %v", eventErr)
+
+	case e != nil && r.fn != nil:
+		var resp *event.Event
+
+		resp, result = r.fn.invoke(ctx, *e)
+
+		if respFn == nil {
+			break
+		}
 
 		// Apply the defaulter chain to the outgoing event.
 		if resp != nil && len(r.eventDefaulterFns) > 0 {
@@ -60,29 +64,24 @@ func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn p
 				*resp = fn(ctx, *resp)
 			}
 			// Validate the event conforms to the CloudEvents Spec.
-			if verr := resp.Validate(); verr != nil {
-				cecontext.LoggerFrom(ctx).Error(fmt.Errorf("cloudevent validation failed on response event: %v, %w", verr, err))
+			if vErr := resp.Validate(); vErr != nil {
+				cecontext.LoggerFrom(ctx).Errorf("cloudevent validation failed on response event: %w", vErr)
 			}
 		}
 
-		// protocol can manual ack by the result
-		if respFn == nil {
-			if !protocol.IsACK(result) {
-				err = m.Finish(result)
-				isFinished = true
-			}
-			return
-		}
-
-		var rm binding.Message
+		// because binding.Message is an interface, casting a nil resp
+		// here would make future comparisons to nil false
 		if resp != nil {
-			rm = (*binding.EventMessage)(resp)
+			respMsg = (*binding.EventMessage)(resp)
 		}
-
-		return respFn(ctx, rm, result)
 	}
 
-	return nil
+	if respFn == nil {
+		// let the protocol ACK based on the result
+		return result
+	}
+
+	return respFn(ctx, respMsg, result)
 }
 
 func (r *receiveInvoker) IsReceiver() bool {
--- a/v2/protocol/http/protocol.go
+++ b/v2/protocol/http/protocol.go
@@ -20,6 +20,12 @@ const (
 	DefaultShutdownTimeout = time.Minute * 1
 )
 
+type msgErr struct {
+	msg    *Message
+	respFn protocol.ResponseFn
+	err    error
+}
+
 // Protocol acts as both a http client and a http handler.
 type Protocol struct {
 	Target          *url.URL
@@ -212,12 +218,6 @@ func (p *Protocol) Respond(ctx context.Context) (binding.Message, protocol.Respo
 	}
 }
 
-type msgErr struct {
-	msg    *Message
-	respFn protocol.ResponseFn
-	err    error
-}
-
 // ServeHTTP implements http.Handler.
 // Blocks until ResponseFn is invoked.
 func (p *Protocol) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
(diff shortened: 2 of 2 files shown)


v2/client/invoker.go
```diff
@@ -40,18 +42,21 @@ func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn p
...
-	if err != nil {
-		return err
-	}
+	var respMsg binding.Message
+	var result protocol.Result
+
+	e, eventErr := binding.ToEvent(ctx, m)
+	switch {
+	case eventErr != nil:
+		result = httpproto.NewResult(http.StatusBadRequest, "Failed to convert Message to Event: %v", eventErr)
```
n3wscott, line 57: the client can't be coupled to http here, we can only test if it is an ack, nack, nor not nil.

v2/client/invoker.go
```diff
@@ -65,24 +70,20 @@ func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn p
...
 
-		// protocol can manual ack by the result
+		// protocol can ack itself based on the result
 		if respFn == nil {
 			if !protocol.IsACK(result) {
-				err = m.Finish(result)
 				isFinished = true
 			}
-			return
+			return m.Finish(result)
```
n3wscott, line 78: this will cause finish to be called twice if !Ack

v2/client/invoker.go
```diff
@@ -65,24 +70,20 @@ func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn p
...
 		}
 
-		var rm binding.Message
 		if resp != nil {
-			rm = (*binding.EventMessage)(resp)
+			respMsg = (*binding.EventMessage)(resp)
```
n3wscott, line 82: while we are here, the cast should check that it was ok here with `respMsg, ok = ...` and then handle `!ok`

v2/client/invoker.go
```diff
@@ -65,24 +70,20 @@ func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn p
...
 		}
-
-		return respFn(ctx, rm, result)
 	}
 
-	return nil
+	return respFn(ctx, respMsg, result)
```
n3wscott, line 86: respFn could be nil

v2/client/invoker.go
antoineco: Are you suggesting that we can only return an error code without any context?

v2/client/invoker.go
antoineco: Where would be the second time?

v2/client/invoker.go
antoineco: Oh no, I think I got it now. I'll change that.

v2/client/invoker.go
antoineco: `resp` is not an interface type so a type assertion won't work, and a cast returns only 1 value.

v2/client/invoker.go
antoineco: Addressed with the condition above.

v2/client/invoker.go
antoineco: Got it, in the defer

v2/protocol/http/protocol.go
```diff
@@ -223,34 +224,45 @@ func (p *Protocol) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
-	var fn protocol.ResponseFn = func(ctx context.Context, resp binding.Message, er protocol.Result, transformers ...binding.Transformer) error {
+	var fn protocol.ResponseFn = func(ctx context.Context, respMsg binding.Message, res protocol.Result, transformers ...binding.Transformer) error {
 		// Unblock the ServeHTTP after the reply is written
 		defer func() {
 			done <- struct{}{}
 		}()
-		status := http.StatusOK
+
 		if finishErr != nil {
-			http.Error(rw, fmt.Sprintf("cannot forward CloudEvent: %v", finishErr), http.StatusInternalServerError)
```
antoineco, line 240: @n3wscott I've got a feeling we want to return here, can you confirm?

v2/client/invoker.go
```diff
@@ -39,50 +40,55 @@ type receiveInvoker struct {
...
-	if err != nil {
-		return err
-	}
+	var respMsg binding.Message
+	var result protocol.Result
+
+	e, eventErr := binding.ToEvent(ctx, m)
+	switch {
+	case eventErr != nil:
+		result = protocol.NewReceipt(false, "failed to convert Message to Event: %v", eventErr)
```
slinkydeveloper, line 57: Look the comment here: https://github.com/cloudevents/sdk-go/issues/478#issuecomment-621123504

antoineco: The function is invoked with a nil message and this `result` few lines below. Did I misunderstand?

slinkydeveloper: What I meant in the comment is that, before invoking `ToEvent`, you should check the encoding of the message

antoineco: The encoding has already been checked and marked as valid when we reach here (maybe that's also an issue).

The problem occurs when fields such as `version` are missing.

slinkydeveloper: I'm sorry you're right it's checked here :smile: https://github.com/cloudevents/sdk-go/pull/480/files/ec24024f23df968a6752e321ea05324804408487#diff-255451943fd83a8be86080df4ef96fb9R221

v2/client/invoker.go
```diff
@@ -39,50 +40,55 @@ type receiveInvoker struct {
 func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn protocol.ResponseFn) (err error) {
 	var isFinished bool
 	defer func() {
+		// skip finishing if we did it explicitly already
 		if !isFinished {
-			if err2 := m.Finish(err); err2 == nil {
-				err = err2
+			if fErr := m.Finish(err); fErr == nil {
```
n3wscott, line 45: ah well there is a logic bug in the original code... if `err2 == nil` should be `err2 != nil`

This will prevent any error from m.Finish to be returned. I like the idea that m.Finish can eat the error. 

Can you return fErr if it is not nil?

maybe it is as simple as changing this whole block to:

```err = m.Finish(err)```

File: v2/client/invoker.go [hunk 8]
```diff
@@ -39,50 +40,55 @@ type receiveInvoker struct {
...
 		// Apply the defaulter chain to the outgoing event.
 		if resp != nil && len(r.eventDefaulterFns) > 0 {
 			for _, fn := range r.eventDefaulterFns {
 				*resp = fn(ctx, *resp)
 			}
 			// Validate the event conforms to the CloudEvents Spec.
-			if verr := resp.Validate(); verr != nil {
-				cecontext.LoggerFrom(ctx).Error(fmt.Errorf("cloudevent validation failed on response event: %v, %w", verr, err))
+			if vErr := resp.Validate(); vErr != nil {
+				cecontext.LoggerFrom(ctx).Error(fmt.Errorf("cloudevent validation failed on response event: %w", vErr))
```
n3wscott, line 75: I think zap has `Errorf`?

v2/client/invoker.go
```diff
@@ -39,50 +40,55 @@ type receiveInvoker struct {
...
-		}
+		respMsg = (*binding.EventMessage)(resp)
+	}
 
-		var rm binding.Message
-		if resp != nil {
-			rm = (*binding.EventMessage)(resp)
+	if respFn == nil {
+		// let the protocol ACK based on the result
+		if !protocol.IsACK(result) {
```
n3wscott, line 84: coud we not just return here and set `err = result`? let the deferred function call finish? because if IsAck, we will call finish anyway.

v2/protocol/http/protocol.go
n3wscott: if finishErr is set it means that the message is closed and responding will do nothing, so I think that logic makes sense. But also, because the message is closed, the http Error will likely be ignored by the caller.

v2/client/invoker.go
antoineco: That's what I wanted to do originally but I was not certain what was going on with that defer.

Done

v2/client/invoker.go
antoineco: Done

v2/client/invoker.go
antoineco: Done

antoineco:
@n3wscott @slinkydeveloper I removed the "Could not read CloudEvent ..." message from the response.

n3wscott:
LGTM

Thanks for the effort!!

PR #482, opened 2020-04-29
Fix comparison to nil interface

### LINKED ISSUE #477: GET requests or bad payloads make the SDK panic
The receiver panics when receiving a GET request of empty payload.

Given this client
```go
func main() {
    cli, _ := cloudevents.NewDefaultClient()
    cli.StartReceiver(ctx, func(cloudevents.Event) {})
}
```

just `curl localhost:8080`.

<details><summary>📄 error trace</summary>

```
panic: runtime error: invalid memory address or nil pointer dereference
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x7b2946]

goroutine 19 [running]:
github.com/cloudevents/sdk-go/v2/protocol/http.(*Message).Finish(0x0, 0x0, 0x0, 0x0, 0x0)
        /go/pkg/mod/github.com/cloudevents/sdk-go/v2@v2.0.0-!r!c1/protocol/http/message.go:151 +0x26
github.com/cloudevents/sdk-go/v2/client.(*receiveInvoker).Invoke.func1(0xd91c20, 0x0, 0xc00006de78)
        /go/pkg/mod/github.com/cloudevents/sdk-go/v2@v2.0.0-!r!c1/client/invoker.go:41 +0x47
panic(0xba6860, 0x128d010)
        /usr/local/go/src/runtime/panic.go:969 +0x166
github.com/cloudevents/sdk-go/v2/protocol/http.(*Message).ReadEncoding(0x0, 0xc000020000)
        /go/pkg/mod/github.com/cloudevents/sdk-go/v2@v2.0.0-!r!c1/protocol/http/message.go:81 +0x5
github.com/cloudevents/sdk-go/v2/binding.ToEvent(0xd91c60, 0xc0000ae000, 0x7fce984681a0, 0x0, 0x0, 0x0, 0x0, 0xc00006ddbc, 0x2, 0x2)
        /go/pkg/mod/github.com/cloudevents/sdk-go/v2@v2.0.0-!r!c1/binding/to_event.go:28 +0x52
github.com/cloudevents/sdk-go/v2/client.(*receiveInvoker).Invoke(0xc00009e560, 0xd91c60, 0xc0000ae000, 0xd91c20, 0x0, 0x0, 0x0, 0x0)
        /go/pkg/mod/github.com/cloudevents/sdk-go/v2@v2.0.0-!r!c1/client/invoker.go:46 +0xfc
github.com/cloudevents/sdk-go/v2/client.(*ceClient).StartReceiver(0xc0000ce0a0, 0xd91c60, 0xc0000ae000, 0xb69e00, 0xcc3008, 0x0, 0x0)
        /go/pkg/mod/github.com/cloudevents/sdk-go/v2@v2.0.0-!r!c1/client/client.go:218 +0x287
github.com/cloudevents/sdk-go/v2/client.(*obsClient).StartReceiver(0xc00009e540, 0xd91c60, 0xc0000ae000, 0xb69e00, 0xcc3008, 0x0, 0x0)
        /go/pkg/mod/github.com/cloudevents/sdk-go/v2@v2.0.0-!r!c1/client/client_observed.go:93 +0xac
main.(*handler).Start.func1(0xc0000dc0c0, 0xc0000b03d0, 0xd91c60, 0xc0000ae000)
        /go/src/github.com/triggermesh/myproject/cmd/mycmdtarget/main.go:85 +0x63
created by main.(*handler).Start
        /go/src/github.com/triggermesh/myproject/cmd/mycmdtarget/main.go:84 +0x7a
exit status 2
```

</details>

The culprit is here:

https://github.com/cloudevents/sdk-go/blob/0fc16e13d87f0f2b1c99ec41183e35c0e68b6b5c/v2/binding/to_event.go#L23-L26

This comparison is `false` even when the request contains no payload.

---

edit: I observe the same result by sending valid payloads with a "bad" or missing `Content-Type` header:

```
$ curl -XPOST -H "Content-Type: application/json" localhost:8080/ -d'{
    "specversion" : "1.0",
    "type" : "com.example.someevent",
    "source" : "/mycontext",
    "id" : "A234-1234-1234",
    "time" : "2018-04-05T17:31:00Z",
    "comexampleextension1" : "value",
    "comexampleothervalue" : 5,
    "datacontenttype" : "text/xml",
    "data" : "<much wow=\"xml\"/>"
}'

// missing "+json" in the header -> boom
```

antoineco:
Fix an issue where comparing a nil Message interface to nil would always return `false`, causing the receiver to crash.

Also respond with an error instead of OK when the message encoding can't be determined.

```
HTTP/1.1 400 Bad Request
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Wed, 29 Apr 2020 10:18:26 GMT
Content-Length: 49

Cannot read CloudEvent: unknown Message encoding
```

Fixes #477

### DIFF
--- a/v2/protocol/http/protocol.go
+++ b/v2/protocol/http/protocol.go
@@ -201,7 +201,12 @@ func (p *Protocol) Respond(ctx context.Context) (binding.Message, protocol.Respo
 		if !ok {
 			return nil, nil, io.EOF
 		}
+
+		if in.msg == nil {
+			return nil, in.respFn, in.err
+		}
 		return in.msg, in.respFn, in.err
+
 	case <-ctx.Done():
 		return nil, nil, io.EOF
 	}
--- a/v2/client/client.go
+++ b/v2/client/client.go
@@ -216,7 +216,7 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 		}
 
 		if err != nil {
-			cecontext.LoggerFrom(ctx).Warn("Error while receiving a message: %s", err.Error())
+			cecontext.LoggerFrom(ctx).Warnf("Error while receiving a message: %s", err)
 			continue
 		}
 
--- a/v2/protocol/http/protocol_test.go
+++ b/v2/protocol/http/protocol_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 	"time"
 
@@ -11,6 +12,7 @@ import (
 	"github.com/cloudevents/sdk-go/v2/protocol"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
@@ -48,13 +50,15 @@ func TestNew(t *testing.T) {
 }
 
 func protocols(t *testing.T) []*Protocol {
-	ps := make([]*Protocol, 0, 1)
+	ps := make([]*Protocol, 1)
 
 	p, err := New()
 	if err != nil {
-		t.Fail()
+		t.Fatalf("Failed to create test Protocol: %s", err)
 	}
-	ps = append(ps, p)
+
+	ps[0] = p
+
 	return ps
 }
 
@@ -133,14 +137,14 @@ func TestReceive(t *testing.T) {
 			wantErr: "nil Context",
 		},
 		"timeout": {
-			ctx:     context.TODO(),
+			ctx:     newDoneContext(),
 			wantErr: "EOF",
 		},
 	}
 	for n, tc := range testCases {
 		for _, p := range protocols(t) {
 			t.Run(n, func(t *testing.T) {
-				ReceiveTest(t, p, tc.ctx, tc.want, tc.wantErr)
+				ReceiveTest(t, p, tc.ctx, nil, tc.want, tc.wantErr)
 			})
 		}
 	}
@@ -161,27 +165,31 @@ func TestRespond(t *testing.T) {
 		"nil context": {
 			wantErr: "nil Context",
 		},
-		"timeout": {
-			ctx:     context.TODO(),
+		"timeout context": {
+			ctx:     newDoneContext(),
 			wantErr: "EOF",
+		},
+		"non-expiring context": {
+			ctx: context.Background(),
 		},
 	}
 	for n, tc := range testCases {
 		for _, p := range protocols(t) {
 			t.Run(n, func(t *testing.T) {
-				if tc.ctx != nil {
-					var done context.CancelFunc
-					tc.ctx, done = context.WithDeadline(tc.ctx, time.Now().Add(time.Millisecond*10))
-					defer done()
-				}
+				go func() {
+					time.Sleep(time.Millisecond * 10)
+					p.incoming <- msgErr{}
+				}()
 
 				got, fn, err := p.Respond(tc.ctx)
 				if tc.wantErr != "" {
-					if err == nil || err.Error() != tc.wantErr {
-						t.Fatalf("Expected error '%s'. Actual '%v'", tc.wantErr, err)
-					}
-				} else if err != nil {
-					t.Fatalf("Unexpected error: %v", err)
+					assert.EqualError(t, err, tc.wantErr)
+				} else {
+					assert.NoError(t, err)
+				}
+
+				if got != nil {
+					assert.NotNil(t, got, "Nil interface compares to nil")
 				}
 
 				if diff := cmp.Diff(tc.want, got); diff != "" {
@@ -213,6 +221,7 @@ func TestServeHTTP_Receive(t *testing.T) {
 		wantErr string
 	}{
 		"nil": {
+			rw:      httptest.NewRecorder(),
 			wantErr: "unknown Message encoding",
 		},
 	}
@@ -220,26 +229,24 @@ func TestServeHTTP_Receive(t *testing.T) {
 		for _, p := range protocols(t) {
 			t.Run(n, func(t *testing.T) {
 				go p.ServeHTTP(tc.rw, tc.req)
-				ReceiveTest(t, p, context.Background(), tc.want, tc.wantErr)
+				rec := (tc.rw).(*httptest.ResponseRecorder)
+				ReceiveTest(t, p, context.Background(), rec, tc.want, tc.wantErr)
 			})
 		}
 	}
 }
 
-func ReceiveTest(t *testing.T, p *Protocol, ctx context.Context, want binding.Message, wantErr string) {
-	if ctx != nil {
-		var done context.CancelFunc
-		ctx, done = context.WithTimeout(ctx, time.Millisecond*10)
-		defer done()
-	}
-
+func ReceiveTest(t *testing.T, p *Protocol, ctx context.Context, rec *httptest.ResponseRecorder, want binding.Message, wantErr string) {
 	got, err := p.Receive(ctx)
 	if wantErr != "" {
-		if err == nil || err.Error() != wantErr {
-			t.Fatalf("Expected error '%s'. Actual '%v'", wantErr, err)
+		assert.EqualError(t, err, wantErr)
+
+		if rec != nil {
+			defer rec.Result().Body.Close()
+			// TODO perform assertions on result if necessary
 		}
-	} else if err != nil {
-		t.Fatalf("Unexpected error: %v", err)
+	} else {
+		assert.NoError(t, err)
 	}
 
 	if want == nil {
@@ -263,4 +270,10 @@ func (r *roundTripperTest) RoundTrip(req *http.Request) (*http.Response, error) 
 	}
 
 	return &http.Response{StatusCode: code}, nil
+}
+
+func newDoneContext() context.Context {
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	return ctx
 }


v2/protocol/http/protocol.go
```diff
@@ -201,7 +201,12 @@ func (p *Protocol) Respond(ctx context.Context) (binding.Message, protocol.Respo
 		if !ok {
 			return nil, nil, io.EOF
 		}
+
+		if in.msg == nil {
+			return nil, in.respFn, in.err
+		}
```
antoineco, line 207: An alternative would be to use reflection to compare `message` to `nil` here: https://github.com/cloudevents/sdk-go/blob/0fc16e13d87f0f2b1c99ec41183e35c0e68b6b5c/v2/binding/to_event.go#L23-L26

I believe we should avoid using reflection whenever possible.

slinkydeveloper: Fully agree!

v2/protocol/http/protocol.go
```diff
@@ -218,6 +223,7 @@ type msgErr struct {
 func (p *Protocol) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	m := NewMessageFromHttpRequest(req)
 	if m == nil || m.ReadEncoding() == binding.EncodingUnknown {
+		http.Error(rw, fmt.Sprintf("Cannot read CloudEvent: %s", binding.ErrUnknownEncoding), http.StatusBadRequest)
```
slinkydeveloper, line 226: Can we address this problem in the https://github.com/cloudevents/sdk-go/pull/480?

slinkydeveloper: Ah ok i see now, the problem is different from #480.

Anyway, I think this is a bad change because `EncodingUnknown` happens in several situations, even on a simple get request without any parameter/cloudevent: eg ff the receiver just wants to return events, without expecting to receive any, the encoding of the received event will be unknown. What i'm trying to say is: at the protocol level, i think we should not just fail the request, this should be up to the protocol user. If you experienced this problem using the client, then we need to address the problem in a different way on the client code.

antoineco: What I want to avoid here is external clients (not CE) sending random requests to the endpoint and crashing it. The HTTP response is to me a plus, just to play nice with the HTTP protocol. If that's not fine with the CE contract I'm happy to remove this.

Still, I'm a bit puzzled when I see a request fail with 200 and no payload in general.

antoineco: Removed in 49fae7d

antoineco:
@slinkydeveloper general question about the case where I send a valid event, but the `Content-Type` header is set to e.g. `Content-Type: application/foo`. Is it acceptable to just respond with `200` like we do?

antoineco:
> Can you try to provide a test to check if the problem you experienced doesn't happen again?

On it.

slinkydeveloper:
> Clearly the event was dropped, but from the perspective of the client we're just fine.

I suppose yes, how do i know that the body content is an event if the content-type is not `application/cloudevents+json`? Maybe `application/foo` uses the same stuff of a cloudevent :smile: .

Apart the jokes, i think this is a perfectly fine behaviour for the `Client`. The answer for these more "complex" use cases where you mix cloudevent and non-cloudevents is: use directly the `http.Protocol` and handle manually the encoding unknown, parsing the eventual non cloudevent request as you prefer

antoineco:
> Maybe `application/foo` uses the same stuff of a cloudevent

Sure, but our endpoint handles `application/cloudevents+json` exclusively, so it should notify the client (_any_ HTTP client) back if it doesn't accept the document type, otherwise the implementation doesn't satisfy [RFC2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.16).

---

I [added a test](https://github.com/cloudevents/sdk-go/compare/49fae7d70b0040ced37065ad78954a49f0a7b526..5341e719805bd33503ba5d9bf8cc3cb1fe766b4b). Here's the output with the change commented:

```
--- FAIL: TestRespond (0.01s)
    --- FAIL: TestRespond/TODO (0.01s)
        protocol_test.go:192:
                Error Trace:    protocol_test.go:192
                Error:          Expected value not to be nil.
                Test:           TestRespond/TODO
                Messages:       Nil interface compares to nil
        protocol_test.go:196: unexpected diff (-want, +got) =   interface{}(
            +   (*http.Message)(nil),
              )
FAIL
FAIL    github.com/cloudevents/sdk-go/v2/protocol/http  2.046s
FAIL
```

slinkydeveloper:
> Sure, but our endpoint handles application/cloudevents+json exclusively, so it should notify the client (any HTTP client) back if it doesn't accept the document type

That's a really interesting point here, i never thought about that... I feel this kind of problem should be brought at spec level, because it gives an hint on how to parse the event https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md#3-http-message-mapping but it also states _Otherwise it defaults to binary mode._ which is clearly what we don't do in the sdk, because in the end the user needs a distinction between event/non event. So the criteria on replying with a content type not supported is interesting, i think you should open an issue on the spec about that

antoineco:
@slinkydeveloper I opened #483 for discussions.

v2/protocol/http/protocol.go
n3wscott: we use reflection in the client invoker but that is only processed once at startup.

v2/protocol/http/protocol.go
```diff
@@ -201,7 +201,12 @@ func (p *Protocol) Respond(ctx context.Context) (binding.Message, protocol.Respo
 		if !ok {
 			return nil, nil, io.EOF
 		}
+
+		if in.msg == nil {
+			return nil, in.respFn, in.err
```
n3wscott, line 206: ah, we got caught by the typed nil in go... nice. 👍

n3wscott:
> $ curl -D- -H "Content-Type: application/foo" localhost:8080/ -d'{
>     "specversion" : "1.0",
>     "type" : "com.example.someevent",
>     "source" : "/mycontext",
>     "id" : "A234-1234-1234",
>     "time" : "2018-04-05T17:31:00Z",
>     "comexampleextension1" : "value",
>     "comexampleothervalue" : 5,
>     "datacontenttype" : "text/xml",
>     "data" : "<much wow=\"xml\"/>"
> }'
> HTTP/1.1 200 OK
> Date: Wed, 29 Apr 2020 10:55:59 GMT
> Content-Length: 0

this should be a 400 return I think.

v2/protocol/http/protocol.go
antoineco: Hehe looks like it, a very vicious one!

antoineco:
@n3wscott so are you fine with me reverting that second commit?

v2/protocol/http/protocol.go
```diff
@@ -223,7 +223,6 @@ type msgErr struct {
 func (p *Protocol) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	m := NewMessageFromHttpRequest(req)
 	if m == nil || m.ReadEncoding() == binding.EncodingUnknown {
-		http.Error(rw, fmt.Sprintf("Cannot read CloudEvent: %s", binding.ErrUnknownEncoding), http.StatusBadRequest)
```
n3wscott: if it returns anything it should be in the form of a cloudevent

v2/protocol/http/protocol.go
```diff
@@ -223,7 +223,6 @@ type msgErr struct {
...
 		p.incoming <- msgErr{msg: nil, err: binding.ErrUnknownEncoding}
 		return // if there was no message, return.
```
n3wscott: we should set the rw.Status to 415 as you are saying

v2/protocol/http/protocol.go
n3wscott: not crashing sounds like a good thing 😄

n3wscott:
> i think this is a perfectly fine behaviour for the `Client`.

Totally. But if the request ends up in the sdk handler and it can not process it, it should reject it and tell the caller that the request was bad. 

We should do this for GET methods too if we are not a pull based protocol, as a follow up.

v2/protocol/http/protocol.go
slinkydeveloper: I don't think so @n3wscott, what about an empty request?

slinkydeveloper: that's why imo this shouldn't fail, even better not here in the protocol where a protocol user could decide to handle the encoding unknown case

antoineco: done

v2/protocol/http/protocol.go
antoineco: @n3wscott I'd appreciate some pointer or example to set all the attributes and format correctly. Do you do that in other places?

v2/protocol/http/protocol.go
n3wscott: slinky has a good point, we use to have a converter feature that would allow us to try to convert or wrap a bad message into a cloudevent and then deliver that to a consumer, how could we allow that use case and still do this if no other option was available?

v2/protocol/http/protocol.go
```diff
@@ -218,6 +223,7 @@ type msgErr struct {
 func (p *Protocol) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	m := NewMessageFromHttpRequest(req)
 	if m == nil || m.ReadEncoding() == binding.EncodingUnknown {
+		http.Error(rw, "Cannot read CloudEvent: "+binding.ErrUnknownEncoding.Error(), http.StatusUnsupportedMediaType)
```
n3wscott, line 226: ok sorry, lets hold off on this change until we decide on a fix in #483 

There is value in letting the client level know that an event was attempted and it failed... the user or client could choose to fix the event for the consumer. So let's hold off on http.Error for now. Sorry for the flip flopping.... The reason is just because we are trying to not put too many runtime choices at the protocol level.

antoineco: Removed

slinkydeveloper:
Thanks for your contribution! @antoineco

Issue #483, opened 2020-04-29
Should the SDK respond with "415: Unsupported Media Type" when the Content-Type is unsupported?


antoineco:
I made the point in #482 that data sent over the HTTP protocol without a valid `Content-Type` header was acknowledged by a `200: OK` code although the request is in fact dropped by the SDK.

I believe this is wrong and conflicts with RFC 2616. I would suggest the SDK responds with
* 415: Unsupported Media Type, when the Content-Type header is not supported
* 400: Bad Request, when the data couldn't be recognized as a CloudEvent payload

@slinkydeveloper suggested in https://github.com/cloudevents/sdk-go/pull/482#issuecomment-621148893 that I open an issue to discuss this:

> That's a really interesting point here, i never thought about that... I feel this kind of problem should be brought at spec level, because it gives an hint on how to parse the event https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md#3-http-message-mapping but it also states _Otherwise it defaults to binary mode._ which is clearly what we don't do in the sdk, because in the end the user needs a distinction between event/non event. So the criteria on replying with a content type not supported is interesting, i think you should open an issue on the spec about that

slinkydeveloper:
@n3wscott any opinions on this one?

n3wscott:
I agree. A 4xx level error in response seems like the right thing.

415 seems to fit. returning 200 sounds like it was just bypassing all our code and returning and that is the default that the writers use. I think this is an easy fix.

slinkydeveloper:
My proposal to handle this issue is:

* if the receiver function expects an event, then **the client** fails if encoding is unknown and protocol correctly writes the error with the right status code (checking if error wraps binding.ErrEncodingUnknown should suffice)
* if the receiver function doesn't expect an event, message is just dropped
* if the user really wants to do crazy stuff like converting non events to event, then he should use the protocol directly

n3wscott:
I like the idea of giving the client a chance to fix (or report) an event that was invalid rather than do it at the protocol level. This gives us hooks to change the response code or behavior by replacing the client without changing the protocol impl. That seems nice.

duglin:
> if the user really wants to do crazy stuff like converting non events to event, then he should use the protocol directly

can you elaborate on what you mean by "use the protocol directly" ?  You mean not use the SDK?

slinkydeveloper:
[...]

PR #484, opened 2020-04-29
client: return result from receiver invoke

lopezator:
Return result from function receiver invoke, this way functions
returning an err such as:

func(event.Event) (*event.Event, protocol.Result)

Will return its err instead of masking it with an empty result

CC/ @n3wscott

### DIFF
--- a/v2/client/client.go
+++ b/v2/client/client.go
@@ -220,8 +220,8 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 			continue
 		}
 
-		if err := c.invoker.Invoke(ctx, msg, respFn); err != nil {
-			return err
-		}
+		err := c.invoker.Invoke(ctx, msg, respFn)
+		fmt.Println("this, however, is still nil", err)
+		return err
 	}
 }
--- a/v2/client/invoker.go
+++ b/v2/client/invoker.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/cloudevents/sdk-go/v2/binding"
 	cecontext "github.com/cloudevents/sdk-go/v2/context"
@@ -78,6 +79,7 @@ func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn p
 
 	if respFn == nil {
 		// let the protocol ACK based on the result
+		fmt.Println("this has the error returned by the func", result)
 		return result
 	}
 


antoineco:
Overlaps with #480

lopezator:
Unfortunately the bug is still there,  see the code and try it by yourself.

Maybe a race condition introduced in the latest commit? Couldn't find the root cause.

This was working ok with before the changes introduced to master just by adding the `return result` instead of just `return` in `Invoke()`.

Could you please take a look and see if you can reproduce? 

CC/ @antoineco @n3wscott

lopezator:
I think I finally understood this thing.

The result err is being masked here:

```
defer func() {
	err = m.Finish(err)
}()
```

That's why the error is empty after:

```
c.invoker.Invoke(ctx, msg, respFn)
```

Correct me if I'm wrong, I assume this is being done in purpose, because we don't want to the receiver to exit when there is an error on the receiverFunc.

What confused me is that now I don't have any way to know if a function has failed or not and why, maybe we should at least log the error here? 

https://github.com/cloudevents/sdk-go/pull/484/files#diff-6d4d1b3b42eccbde61fe76507433ac65R82

```
cecontext.LoggerFrom(ctx).Error(result)
```

I don't know exactly how, because the above would flood during repetion on a not recoverable error, but I definitevely think that is there is an error inside the `SendReceiver` function it should be printed/returned somehow intead of silently failing.

Thoughts?

v2/client/client.go
```diff
@@ -220,8 +220,8 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 			continue
 		}
 
-		if err := c.invoker.Invoke(ctx, msg, respFn); err != nil {
-			return err
-		}
+		err := c.invoker.Invoke(ctx, msg, respFn)
+		fmt.Println("this, however, is still nil", err)
```
antoineco, line 224: Because returning an error here is **fatal**, it means the receiver has to exit. You don't want a bad input to crash the receiver, that could be abused by any client.

antoineco:
@lopezator I'm also quite unhappy about the way the SDK obfuscates the errors server side. It doesn't play very nice with the HTTP protocol when it comes to reporting errors, especially for bad requests.

See #483.

lopezator:
Ok! I see.

Then would be the correct way to handle errors inside StartReceiver funcs with this signature then?

```
* func(context.Context, event.Event) protocol.Result
```

Returning an err silently fails, which is pretty confusing.

I guess they only way is just to log then return err in my side?

```go
func (s *Something) MyBusiness(ctx context.Context, event cloudevents.Event) error {
	if a == "bad" {
		err := errors.New("something went wrong")
		fmt.Println(err) // at least print it out
		return err // nack, but silent fail
	}
	return nil // ack
}
```

antoineco:
If you custom receive func returns a non-ACK result, CE clients shoukd pick that error, e.g. [like this](https://github.com/triggermesh/aws-event-sources/blob/99ce7fd746/pkg/adapter/awsdynamodbsource/adapter.go#L239).

I haven't tried to provoke a failure on my side using a Go event producer yet. If you have, can you confirm the clients sees a ACK even in case of error?

slinkydeveloper:
Is this PR still relevant?

antoineco:
I guess not, now that #498 was merged (I didn't know there was a PR open, thanks!)

slinkydeveloper:
Ok, @lopezator @antoineco reopen if you need it again!

Issue #487, opened 2020-05-03
Client.StartReceiver causes panic if Kafka connection is lost


timjonesdev:
https://github.com/cloudevents/sdk-go/blob/v2.0.0-RC2/v2/client/client.go#L192-L200

```golang
// Start the opener, if set.
if c.opener != nil {
	go func() {
		// TODO: handle error correctly here.
		if err := c.opener.OpenInbound(ctx); err != nil {
			panic(err)
		}
	}()
}
```

I'm just getting started trying to use the SDK with Kafka, and I'm trying to test how to recover from a lost connection. It seems that this panic causes my application to crash, and I'm wondering if there are plans to address the `TODO`.

If not, can you help with ideas for how to recover from the panic? Thank you for your time!

slinkydeveloper:
Hi @timjonesdev, thanks for spotting the issue! There is an hidden gem in the library we use for kafka i need to fix, I'll assign this issue to myself and i'll ping you when it's ready.

slinkydeveloper:
Check out the https://github.com/cloudevents/sdk-go/issues/488

timjonesdev:
Awesome! Thanks for the quick response!

timjonesdev:
I wanted to mention, this fixed my issue. I am now able to reconnect to Kafka once I restart it. Thanks again!

PR #488, opened 2020-05-04
Fixed consumer group usage

slinkydeveloper:
Fix https://github.com/cloudevents/sdk-go/pull/487

### DIFF
--- a/v2/protocol/kafka_sarama/receiver.go
+++ b/v2/protocol/kafka_sarama/receiver.go
@@ -148,7 +148,7 @@ func (c *Consumer) startConsumerGroupLoop(cg sarama.ConsumerGroup, ctx context.C
 			return
 		// Something else happened
 		default:
-			if err == nil || (err != nil && (err == sarama.ErrClosedClient || err == sarama.ErrClosedConsumerGroup)) {
+			if err == nil || err == sarama.ErrClosedClient || err == sarama.ErrClosedConsumerGroup {
 				// Consumer group closed correctly, we can close that loop
 				return
 			} else {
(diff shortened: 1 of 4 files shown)


v2/protocol/kafka_sarama/receiver.go
```diff
@@ -114,9 +114,7 @@ func (c *Consumer) OpenInbound(ctx context.Context) error {
 
 	errCh := make(chan error)
 
-	go func(errs chan error) {
-		errs <- cg.Consume(context.Background(), []string{c.topic}, c)
-	}(errCh)
+	go c.startConsumerGroupLoop(cg, ctx, errCh)
 
 	select {
 	case <-ctx.Done():
```
pierDipi, line 120: The select statement blocks until it gets an error from `errCh` or the context is done.
If `cg.Consume(...)` returns, for example, `ErrClosedClient`, we stop consuming and the select statement will block until it gets a signal from the context. 

Is that what we want?

v2/protocol/kafka_sarama/receiver.go
```diff
@@ -135,6 +133,33 @@ func (c *Consumer) OpenInbound(ctx context.Context) error {
...
+		select {
+		// If context is closed, then consumer group session was closed by the user
+		case <-ctx.Done():
+			if err != nil {
+				errs <- err
+			}
+			return
+		// Something else happened
+		default:
+			if err == nil || (err != nil && (err == sarama.ErrClosedClient || err == sarama.ErrClosedConsumerGroup)) {
```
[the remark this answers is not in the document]
slinkydeveloper: Good spot!

v2/protocol/kafka_sarama/receiver.go
slinkydeveloper: > Is that what we want?

yep, all `OpenInbound` implementations are intended to block: https://github.com/cloudevents/sdk-go/blob/master/v2/protocol/lifecycle.go#L9

PR #491, opened 2020-05-04
Abuse Protection for webhooks

n3wscott:


slinkydeveloper:
looks like you pulled in the changes to the website :smile:

n3wscott:
I bet I just have to rebase

v2/cmd/samples/http/raw/main.go
```diff
@@ -0,0 +1,34 @@
+package main
+
+import (
```
n3wscott, line 3: I added this back in, it is extremely useful for debugging.

slinkydeveloper: Can we have it somewhere else? it's not really an sdk sample...

v2/cmd/samples/http/receiver-protected/main.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	rate := 100
+	p.WebhookConfig = &http.WebhookConfig{
+		AllowedMethods:  []string{"POST", "OPTIONS"},
+		AllowedRate:     &rate,
+		AutoACKCallback: true,
+		AllowedOrigins:  []string{"http://localhost:8181"},
+	}
+	p.OptionsHandlerFn = p.OptionsHandler
```
slinkydeveloper, line 26: ?

v2/cmd/samples/http/sender/main.go
```diff
@@ -30,6 +30,7 @@ func main() {
 			"message": "Hello, World!",
 		})
 
+		ctx = cloudevents.WithEncodingStructured(ctx)
```
slinkydeveloper, line 33: Is it a leftover from a debugging session?

v2/protocol/http/protocol.go
```diff
@@ -221,6 +230,33 @@ func (p *Protocol) Respond(ctx context.Context) (binding.Message, protocol.Respo
...
+	case http.MethodOptions:
+		if p.OptionsHandlerFn == nil {
+			rw.WriteHeader(http.StatusMethodNotAllowed)
+			return
+		}
+		p.OptionsHandlerFn(rw, req)
+		return
+
+	case http.MethodGet:
+		if p.GetHandlerFn == nil {
```
slinkydeveloper, line 244: This looks wrong: if i have a receiver that doesn't accept events but just creates it, i think get should just work ootb without configuring `GetHandlerFn`

v2/cmd/samples/http/receiver-protected/main.go
slinkydeveloper: Can this be a default?

v2/cmd/samples/http/raw/main.go
n3wscott: maybe I move it to `cmd/tools`

v2/cmd/samples/http/receiver-protected/main.go
n3wscott: yeah a future PR can add this to the default client to respond to OPTIONS

v2/cmd/samples/http/sender/main.go
n3wscott: It kinda is, I should switch it to a flag. I used this to send events to azure event grid but it needed more...

v2/protocol/http/protocol.go
n3wscott: what does http.GET do on a push based protocol?

slinkydeveloper: Like: https://github.com/cloudevents/sdk-go/pull/498/files#diff-3666ce1687203ab4a81aec71bee67724R81

slinkydeveloper: I think it's a reasonable use case to support, given also that is cheap to mantain: eg you want to create an endpoint (like a REST API) that replies with some informations wrapped in a cloudevent

v2/cmd/samples/http/raw/main.go
slinkydeveloper: fine for me :+1:

v2/cmd/samples/http/receiver-protected/main.go
n3wscott: I think it should be opt-in, because you need to know the rate limit and such

n3wscott: plus the default allowed origin would have to be "*" which seems like a bad default

slinkydeveloper: Do you think it should be an opt-in or an opt-out feature? I don't think this interfere with the average usage of sdk-go

v2/cmd/samples/http/sender/main.go
slinkydeveloper: sure, and please update the documentation of the example here: https://github.com/cloudevents/sdk-go/blob/b3a2bcf1ef60d18bfb312acd5802bffdf46a82ad/v2/cmd/samples/README.md

v2/cmd/samples/http/receiver-protected/main.go
n3wscott: because OPTIONS is optional by the spec, it should be OPT-IN so that integrators have to make choices about rate limits and origin filters up front.

v2/protocol/http/protocol.go
n3wscott: oh yeah, added. nice.

v2/cmd/samples/http/sender/main.go
n3wscott: I will revert for now and add later.

v2/cmd/samples/http/receiver-protected/main.go
slinkydeveloper: Okok i feel we should document it and make a "better entrypoint", like  and option for the protocol `EnableWebHookSupport()` or something like that that does this assignment

v2/protocol/http/protocol_test.go
```diff
@@ -222,6 +222,7 @@ func TestServeHTTP_Receive(t *testing.T) {
 	}{
 		"nil": {
 			rw:      httptest.NewRecorder(),
+			req:     httptest.NewRequest("POST", "http://unittest", nil),
```
n3wscott, line 225: this has changed how https://github.com/cloudevents/sdk-go/pull/491/files#diff-255451943fd83a8be86080df4ef96fb9R263 is tested. Before the test would return m == nil because the test was testing an empty request. I made it non-nil and there is an issue.

v2/protocol/http/protocol.go
```diff
@@ -262,6 +262,7 @@ func (p *Protocol) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	m := NewMessageFromHttpRequest(req)
 	if m == nil || m.ReadEncoding() == binding.EncodingUnknown {
 		p.incoming <- msgErr{msg: nil, err: binding.ErrUnknownEncoding}
+		rw.WriteHeader(http.StatusUnsupportedMediaType)
```
slinkydeveloper, line 265: Nope :smile:

slinkydeveloper: This breaks this use case: https://github.com/cloudevents/sdk-go/blob/master/v2/cmd/samples/message-handle-non-cloudevents/main.go

http.UnsupportedMediaType is returned by that https://github.com/cloudevents/sdk-go/blob/master/v2/protocol/http/protocol.go#L259

n3wscott: I tried to find where the spec says this and was unable to find any data. It seems real wrong to send back a 200 if it is not a cloudevent. The only way to get to this code is if it is expecting to get a cloudevent.

slinkydeveloper: But that means that we are impeding to use the sdk to receive non events completely? If that's the case, then we need to proper document it and make this behaviour consistent between protocols

n3wscott: you can still map a path to a custom handler, there are a few examples of that. But for events that are directed to the sdk handler, and the request is not a CloudEvent, it should be an error.

slinkydeveloper: What I'm saying is that i don't see how "leaving this door open" could be a problem

n3wscott: in this codepath, there is no chance to get to the users code of the sdk, it is returning `HTTP - 200 OK` when the return hits. I think that is a bug because it means the caller can do things like post a blank message to the sdk handler and it will get back `200 OK`. We need to tell the caller they done screwed up, or are calling an end point that has no idea what you are talking about. In this codepath the sdk bails out. It is totally an error on the callers side and not `OK`.

n3wscott: Note that `msg` is `nil`. which means the request is not held open like in other cases.

n3wscott: ok removed.

v2/protocol/http/abuse_protection.go
```diff
@@ -0,0 +1,121 @@
...
+func (p *Protocol) validateOrigin(ro string) (string, bool) {
+	cecontext.LoggerFrom(context.TODO()).Infow("Validating origin.", zap.String("origin", ro))
+
+	for _, ao := range p.WebhookConfig.AllowedOrigins {
+		if ao == "*" {
+			return ao, true
+		}
+		// TODO: it is not clear what the rules for allowed hosts are.
+		// Need to find docs for this. For now, test for prefix.
+		if strings.HasPrefix(ro, ao) {
```
slinkydeveloper, line 115: Maybe we could use this https://github.com/rs/cors/blob/master/cors.go or something similar to validate the origins, let's do it in a follow up.

v2/test/integration/http/receiver_v1_test.go
```diff
@@ -90,6 +90,42 @@ func TestClientReceiver_Status_Codes(t *testing.T) {
...
 				ContentLength: 0,
 			},
+			receiverFuncFactory: func(cancelFunc context.CancelFunc) interface{} {
+				return func() *cloudevents.Event {
+					defer cancelFunc()
+					return nil // acts as a 200 OK
+				}
+			},
+		},
+		"405 if the receiver is not expecting a GET request": {
```
slinkydeveloper, line 100: Maybe in a followup, but we need a test for the use case:

> I want to set a receiver that doesn't accept events, accepts method `GET`, and the user replies with events

n3wscott:
Issues added.

PR #495, opened 2020-05-06
Fix kafka samples

xstevens:
This fixes the kafka sample applications so that you can run them. Mostly this involves setting a config version and required settings by sarama's synchronous producer.

For the sender and receiver example, I've changed it to use the same topic for receiver and sender and added a barrier to make sure the receiver has seen `count` messages before allowing the main function to exit.

### DIFF
--- a/v2/cmd/samples/kafka/sender-receiver/main.go
+++ b/v2/cmd/samples/kafka/sender-receiver/main.go
@@ -18,7 +18,6 @@ const (
 
 func main() {
 	saramaConfig := sarama.NewConfig()
-	saramaConfig.Producer.Return.Successes = true
 	saramaConfig.Version = sarama.V2_0_0_0
 
 	// With NewProtocol you can use the same client both to send and receive.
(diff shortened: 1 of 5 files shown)


n3wscott:
Awesome!! thanks for this. 

LGTM

cc: @slinkydeveloper as the kafka expert.

slinkydeveloper:
@xstevens thanks for your contribution! You need to fix the commit signoff, like explained here: https://github.com/cloudevents/sdk-go/pull/495/checks?check_run_id=650033696

v2/cmd/samples/kafka/sender-receiver/main.go
```diff
@@ -4,18 +4,25 @@ import (
...
 	"github.com/cloudevents/sdk-go/v2/protocol/kafka_sarama"
 )
 
+const (
+	count = 10
+)
+
 func main() {
 	saramaConfig := sarama.NewConfig()
+	saramaConfig.Producer.Return.Successes = true
```
slinkydeveloper, line 21: Mh now that i see that, I think it just makes sense to force this value directly in production code, here:

* https://github.com/cloudevents/sdk-go/blob/master/v2/protocol/kafka_sarama/sender.go#L19
* https://github.com/cloudevents/sdk-go/blob/master/v2/protocol/kafka_sarama/protocol.go#L42

PR #497, opened 2020-05-07
Implemented parallelism in client polling

### LINKED ISSUE #453: Implement parallelism in client for Receive/Responder
Now in `Client` there is a single poll loop for `Receive`/`Responder` https://github.com/cloudevents/sdk-go/blob/master/v2/client/client.go#L205

We should probably have N goroutines polling, where N presumably could be `GOMAXPROCS` (one poll loop per OS thread).

This also means that the protocols implementations should have a sized channel, presumably same of number of Goroutines: https://github.com/cloudevents/sdk-go/blob/master/v2/protocol/http/protocol.go#L57

slinkydeveloper:
Fix #453

### DIFF
--- a/v2/client/client.go
+++ b/v2/client/client.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"runtime"
 	"sync"
 
 	"go.uber.org/zap"
@@ -47,7 +48,10 @@ type Client interface {
 // New produces a new client with the provided transport object and applied
 // client options.
 func New(obj interface{}, opts ...Option) (Client, error) {
-	c := &ceClient{}
+	c := &ceClient{
+		// Running runtime.GOMAXPROCS(0) doesn't update the value, just returns the current one
+		pollGoroutines: runtime.GOMAXPROCS(0),
+	}
 
 	if p, ok := obj.(protocol.Sender); ok {
 		c.sender = p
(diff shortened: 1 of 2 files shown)


v2/client/client.go
```diff
@@ -200,27 +209,35 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
...
-		if err != nil {
-			cecontext.LoggerFrom(ctx).Warnf("Error while receiving a message: %s", err)
-			continue
-		}
 
-		if err := c.invoker.Invoke(ctx, msg, respFn); err != nil {
-			return err
-		}
+	// Start Polling.
+	wg := sync.WaitGroup{}
```
n3wscott, line 214: nice.

v2/client/client.go
```diff
@@ -200,27 +209,35 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
...
+	for i := 0; i < c.pollGoroutines; i++ {
```
n3wscott, line 215: I do not think you want to `i++` here, this looks like it will grow too large for long lived processes.

slinkydeveloper: ?!

n3wscott: nevermind!

n3wscott:
LGTM

PR #500, opened 2020-05-08
dropping v1 dir

n3wscott:
also adds a tag for the conformance tests. 


n3wscott:
v1 comes from go mod releases out of the `release-1.y.z` branch

PR #502, opened 2020-05-13
Data race fix in client

slinkydeveloper:
There were two data races in client:

* Using `GetPort` for getting the actual port was racy because another goroutine is starting the listening. `GetPort` now is renamed with `GetListeningPort` for clarity and the internal `net.Listener` is behind an atomic.Value
* Trivial data race introduced by this PR https://github.com/cloudevents/sdk-go/pull/497

Enabled data race check in CI

### DIFF
--- a/hack/integration-test.sh
+++ b/hack/integration-test.sh
@@ -16,7 +16,7 @@ fi
 COVERPKG=$(go list ./... | grep -v /vendor | tr "\n" ",")
 for gomodule in $(go list ./test/integration/... | grep -v /cmd | grep -v /vendor)
 do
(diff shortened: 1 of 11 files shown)


slinkydeveloper:
@n3wscott I disabled the pubsub test because it's preventing this pr to be merged.

Do you have any input on why the pubsub test fail? On my machine they run smoothly, but CI doesn't like them, and doesn't log why https://circleci.com/gh/cloudevents/sdk-go/1138

n3wscott:
that is very weird. File an issue and we can give it to the pubsub maintainers to look at.

LGTM

Issue #512, opened 2020-05-19
Why even attempt to convert resp to event when protocol.Request returns error?


yolocs:
Here:

https://github.com/cloudevents/sdk-go/blob/97abfeb3da0bed09e395bff2c5bcf35b6435cb5f/v2/client/client.go#L142-L166

When `requester.Request` returns an error (line 144), it still attempts to convert the resp to an event (line 154). The error message seems confusing in certain cases. Here is an example.

Call `client.Request` to an HTTP server that always returns 400. The error message I see is:

```
time="2020-05-19T20:53:57Z" level=error msg="Failed to seed event (id=eed6e5df-917e-42cf-9e30-40120f003895) to target \"http://actor.m0.svc.cluster.local\": 400: %!w(*protocol.Receipt=&{false  []}); failed to convert response into event: unknown Message encoding"
```

This is confusing because I'm replying 400 to indicate an error, in which case the sdk shouldn't even attempt to convert my resp.

slinkydeveloper:
I think you're right, it's probably missing an `if !protocol.IsACK(err)` statement

PR #516, opened 2020-05-20
Fix client.Request error handling

yolocs:
Signed-off-by: Chen Shou <cshou@google.com>

Fix: https://github.com/cloudevents/sdk-go/issues/512


v2/client/client.go
```diff
@@ -149,18 +149,17 @@ func (c *ceClient) Request(ctx context.Context, e event.Event) (*event.Event, pr
 			}
 		}()
 	}
+	if protocol.IsNACK(err) {
```
slinkydeveloper, line 152: I think what you should check here is `!protocol.IsACK(err)`. err could be either ack, nack or none of both, which means error even before sending

v2/client/client_test.go
```diff
@@ -78,6 +78,182 @@ func simpleStructuredClient(target string) client.Client {
...
+	structureLen := int64(len(structureData))
+
+	testCases := map[string]struct {
+		c       func(target string) client.Client
+		event   event.Event
+		resp    *http.Response
+		want    *requestValidation
+		wantRes string
+	}{
+		"binary simple v0.3": {
```
slinkydeveloper, line 99: Why you added such tests? These are already covered by tests in integration tests. Can you move the test you made specifically for the error response somewhere in integration tests and remove the tabular test here?

yolocs: Could you point out where is the integ tests?

v2/client/client.go
n3wscott: yeah, NACK means only NACK, not rando error.

v2/client/client_test.go
n3wscott: in `./v2/test`

n3wscott: Yeah these look  a lot like the tap tests we have

v2/client/client.go
slinkydeveloper: @n3wscott OT but should we have a more comprehensive method like `protocol.IsFailed(res)`?

v2/client/client_test.go
n3wscott: maybe this change means there is a hole in those tests?

slinkydeveloper: It could be, maybe we need to cover more the status codes like we did recently with the receiver

yolocs: FWIW, this is something pretty easy to catch with unit test here.

v2/client/client.go
n3wscott: Sure, we can add more helpers as we find them useful.

v2/client/client_test.go
yolocs: @slinkydeveloper I believe there is bug in the loopback test lib. It doesn't verify `wantResult` correctly. I don't get to further investigate this issue.

v2/test/integration/http/loopback_v03_test.go
```diff
@@ -153,6 +155,39 @@ func TestClientLoopback_binary_base64_v03tov03(t *testing.T) {
...
+				Context: cloudevents.EventContextV1{
+					ID:              "ABC-123",
+					Type:            "unit.test.client.sent",
+					Source:          *cloudevents.ParseURIRef("/unit/test/client"),
+					Subject:         strptr("resource"),
+					DataContentType: cloudevents.StringOfApplicationJSON(),
+				}.AsV1(),
+				DataEncoded: toBytes(map[string]interface{}{"hello": "unittest"}),
+			},
+			result: cloudevents.NewHTTPResult(http.StatusBadRequest, "unit test %s", http.StatusText(http.StatusBadRequest)),
```
slinkydeveloper, line 170: Because 400 could be fired directly by the sdk internals, can you use whatever other status code to check that the status code you wrote as result is exactly the one you receive on the other side?

n3wscott: We can follow up with more tests for more status codes. Let's let this test in and add more.

slinkydeveloper: I'm not saying to add more tests for status codes, i'm just saying to change this because this status code could be triggered by the sdk and not by the user, while the test here is testing that the status code triggered by the user is written on the wire

yolocs: Found out why result assertion doesn't work. We didn't set return status code in the `TapHandler`. Fixed that and a bunch of related tests. Trying to keep a minimum change in this PR. So please don't suggest any behavioral change in this PR.

n3wscott:
LGTM

PR #522, opened 2020-05-27
Created some nice assertions for events

slinkydeveloper:
That's an initial list of assertions, we can add more of them later


ericlem:
This looks nice.  I'd definitely like to see this go in.  There may be some more predicates we want later, but they are easy to add even by external callers by just implementing the interface, so I'm good with what's there.

slinkydeveloper:
> There may be some more predicates we want later, but they are easy to add even by external callers by just implementing the interface, so I'm good with what's there.

Sure, or we can add them here, so they can be beneficial for all users

PR #523, opened 2020-05-27
Creating automation to help create releases.

n3wscott:
Steps:

1. Run the ./hack/tag-release.sh locally, and push up the tags. 
2. github actions will generate releases from these tags.

Later we can automate the creation of the sub-modules from the parent release.


.github/workflows/draft-release.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+    - name: Create Draft Release
+      id: create_draft_release
+      uses: actions/create-release@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        tag_name: ${{ github.ref }}
+        release_name: Release ${{ github.ref }}
+        draft: true
```
slinkydeveloper, line 25: draft release? is it valid?

n3wscott: I wanted to have a run with this as a draft before trusting it, it is valid.

slinkydeveloper: Ok cool, maybe we should understand how to make it configurable somehow

slinkydeveloper: like some fancy regex depending on the tag

slinkydeveloper: but you can do it as followup after we enable it for real releases :)

PR #528, opened 2020-06-01
Fix pubsub message nil pointer due to missing format

liu-cong:
Fix a nil pointer bug, also add basic testing.

This bug caused panic in GCP broker: https://github.com/google/knative-gcp/issues/1178


n3wscott:
LGTM

PR #534, opened 2020-06-16
Fix the link to the examples folder in the README.md

afrittoli:
Examples are not available under ./v2/examples (anymore?), they
are in the ./examples directory, so fixing the link in the root
REAME.md.


slinkydeveloper:
Thanks for spotting it!

PR #539, opened 2020-06-24
fix broken tests

johejo:
## Description
On the CI it looks like the test is passing, but actually not.
I think this is an issue that needs to be resolved early as soon as possible for the project to proceed well.

## Why
`tee` hides the original return code.
https://github.com/cloudevents/sdk-go/blob/b276f25392c52ea8c93ef7713c817f4e5d551eb1/.circleci/config.yml#L59-L70

```
$ git checkout master
$ ./hack/unit-test.sh
...
...
PASS
coverage: 11.2%
ok  	github.com/cloudevents/sdk-go/v2/types	0.061s	coverage: 11.2%
FAIL
$ echo $?
1
```

## Need attention
And I fixed the test because it is normal to panic if a nil context is passed.
Follows the behavior of the standard library.
https://github.com/golang/go/blob/9f33108dfa22946622a8a78b5cd3f64cd3e455dd/src/context/context.go#L521

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,14 +59,17 @@ jobs:
       - run:
           name: Run unit tests
           command: |
+            set -o pipefail
             ./hack/unit-test.sh | tee ${TEST_RESULTS}/go-unit-test.out
       - run:
           name: Run integrations tests
           command: |
+            set -o pipefail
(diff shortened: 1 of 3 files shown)


johejo:
This seems to be the correct test for Go 1.15, but it seems to fail for 1.14.
How should I fix it...

slinkydeveloper:
Thank you for your contribution!

Issue #537, opened 2020-06-23
"http: superfluous response.WriteHeader call" error, unreceived events when serving via http.ListenAndServe 


maschmid:
Basing the receiver code on https://github.com/cloudevents/sdk-go/blob/master/samples/http/receiver-gorilla/main.go  , the receiver sometimes doesn't receive events, when events are sent in parallel, showing the following errors in logs:

```
http: superfluous response.WriteHeader call from github.com/cloudevents/sdk-go/v2/protocol/http.(*Protocol).ServeHTTP.func2 (protocol.go:321)
```

The same code based on https://github.com/cloudevents/sdk-go/blob/master/samples/http/receiver/main.go works fine.

See https://github.com/maschmid/cloudevents-superfluous-error for a reproducer, containg both a working and broken variants.

n3wscott:
Thanks for the example repo, this helps a lot...

ok findings so far, it looks like this happens when the receiver returns a 500, still digging in.

```going to write out status, 500
2020/06/25 09:07:33 http: superfluous response.WriteHeader call from github.com/cloudevents/sdk-go/v2/protocol/http.(*Protocol).ServeHTTP.func2 (protocol.go:323)```

**Comment by n3wscott** [comment]
The error seems to be related to the size of the payload.

```
going to write out status, 500, failed to convert Message to Event: http: invalid Read on closed Body
```

**Comment by n3wscott** [comment]
Perhaps this: https://github.com/golang/go/issues/4637#issuecomment-66073563

We might be doing something wrong... 😬

**Comment by n3wscott** [comment]
got it.

from net/http server.go:93

```
// A ResponseWriter may not be used after the Handler.ServeHTTP method
// has returned.
```

PR pending to fix.

PR #541, opened 2020-06-25
fix sync issues when using standalone http receivers

n3wscott:
fixes https://github.com/cloudevents/sdk-go/issues/537

Thanks for the repo-repo @maschmid it helped a lot to track this down. 

The root cause was the function that delegated to the ServeHTTP was exiting too soon and closing the http connection, so the bindings message to event function was attempting to read the large body out of a closed http connection, throwing an error, attempting to return a 500, but unable to do that because the ... connection was already closed. 

Adding a sync group to the standalone ServeHTTP method and also using sync.WaitGroup in the normal ServeHTTP to be consistent (I know, a little overkill, but I think it might be better than just a done channel. It gives options for batch mode in the near future.)

### DIFF
--- a/v2/client/http_receiver.go
+++ b/v2/client/http_receiver.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"net/http"
+	"sync"
 
 	thttp "github.com/cloudevents/sdk-go/v2/protocol/http"
 )
@@ -25,8 +26,11 @@ type EventReceiver struct {
 }
 
 func (r *EventReceiver) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	wg := sync.WaitGroup{}
+	wg.Add(1)
 	go func() {
 		r.p.ServeHTTP(rw, req)
+		wg.Done()
 	}()
 
 	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


v2/client/http_receiver.go
```diff
@@ -25,8 +26,11 @@ type EventReceiver struct {
 }
 
 func (r *EventReceiver) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	wg := sync.WaitGroup{}
+	wg.Add(1)
 	go func() {
 		r.p.ServeHTTP(rw, req)
+		wg.Done()
```
vaikas, line 33: I admittedly don't have the full context just saw this floating by so apologies for likely noise :) 
Can't we just call this synchronously instead of in a func? Am I missing something? Perhaps worthy of a comment so it's easier to reason about, but it's probably just me :)

n3wscott: There are internal details of the SDK that require these two parts to work in parallel based on how the bindings work is documented, so yeah the server needs to go and it will block until the receive call is finished which is the second block of blocking call, but the bug was the function can't exit until ServeHTTP has finished, which is blocked on the callback... 

yikes

Issue #526, opened 2020-06-01
Simpler result checking than !cloudevents.IsACK()


slinkydeveloper:
Now to check if a Send/Receive was completed succesfully (without caring about the error side), the user should do:

```go
if !cloudevents.IsACK(result) {
  // Something happened but i don't care what, i just log the error
  log.Printf("Something wrong: %s", result.Error())
} else {
    // It worked
}
```

The condition is not trivial to understand for somebody that doesn't have knowledge of the sdk, and in particular of the result type. I propose to create a simpler method that has the same semantic of `!cloudevents.IsACK` but it's simpler to use to cover such use cases.

slinkydeveloper:
An idea:

* `IsSuccess()` == `cloudevents.IsACK()`
* `IsFailure()` == `!cloudevents.IsACK()`

slinkydeveloper:
cc @n3wscott

n3wscott:
what about the case where the result is an ACK and also an error?

slinkydeveloper:
I suppose `IsSuccess()` should identify it as well? Can you do an example?

dan-j:
I agree that it's not trivial to understand how the `protocol.IsACK()` or `procotol.IsNACK()` works, after just spending over an hour studying the in's and out's of it, however I can't see how your `IsSuccess()`/`IsFailure()` idea would improve it?

Why not just keep `protocol.IsACK()` and remove `protocol.IsNACK()`? 

To help argue the point, I see four possible permutations of `error` which you can pass to either function (forgive me if my logic is wrong, in which case this whole argument falls down, although it implies these edge cases needs tests)

- if `err == nil` then `IsACK(err) == true` and `IsNACK(err) == false`
- if `err` wraps a `protocol.Receipt(true)` then `IsACK(err) == true` and `IsNACK(err) == false`
- if `err` wraps a `protocol.Receipt(false)` then `IsACK(err) == false` and `IsNACK(err) == true`

All the above 3 permutations can produce the same result by just doing `!protocol.IsACK(err)`. I can only see a logical difference between then when it comes to this fourth permutation:

- if `err != nil` and doesn't wrap a receipt, then `IsACK(err) == false` and `IsNACK(err) == false` (unless the underlying `error` also implements `Is()` and returns `true`).

[...]

PR #542, opened 2020-06-25
Adding IsUndelivered to help the readability of results from the protocol layers

n3wscott:
Fixes https://github.com/cloudevents/sdk-go/issues/526

No API change, just a new helper method to understand result a little better.

Old:

```go
		if res := c.Send(ctx, e); !cloudevents.IsACK(res) {
			log.Printf("[sender] failed to send: %v", res)
		} else {
			log.Printf("[sender] sent: %d", i)
		}
```

Turns out, this is not true always, `!IsACK` can mean it was send and not accepted on the target side.

New:

```go
		if res := c.Send(ctx, e); cloudevents.IsUndelivered(res) {
			log.Printf("[sender] failed to send: %v", res)
		} else {
			log.Printf("[sender] sent: %d, accepted: %t", i, cloudevents.IsACK(res))
		}
```

Now we can quickly understand if the event was seen by the target or not, and then test if it was accepted or not.

### DIFF
--- a/v2/protocol/result.go
+++ b/v2/protocol/result.go
@@ -53,11 +53,11 @@ func IsNACK(target Result) bool {
 	return ResultIs(target, ResultNACK)
 }
 
-// Undelivered true means the target result is not an ACK/NACK, but some other
+// IsUndelivered true means the target result is not an ACK/NACK, but some other
 // error unrelated to delivery not from the intended recipient. Likely target
 // is an error that represents some part of the protocol is misconfigured or
 // the event that was attempting to be sent was invalid.
-func Undelivered(target Result) bool {
+func IsUndelivered(target Result) bool {
 	return !ResultIs(target, ResultACK) && !ResultIs(target, ResultNACK)
 }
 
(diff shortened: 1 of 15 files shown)


v2/protocol/result.go
```diff
@@ -47,10 +48,19 @@ func IsACK(target Result) bool {
...
+// IsNACK true means the recipient did not acknowledge the event.
 func IsNACK(target Result) bool {
 	return ResultIs(target, ResultNACK)
 }
 
+// Undelivered true means the target result is not an ACK/NACK, but some other
+// error unrelated to delivery not from the intended recipient. Likely target
+// is an error that represents some part of the protocol is misconfigured or
+// the event that was attempting to be sent was invalid.
+func Undelivered(target Result) bool {
```
slinkydeveloper, line 60: The naming is not consistent with `IsACK` and `IsNACK`, perhaps it should be `IsUndelivered`?

n3wscott: Sure, Grouping these together seems fine.

PR #544, opened 2020-06-26
fix gh action from api change.

n3wscott:


n3wscott:
Super minor release process change needed for GH actions to work.

PR #550, opened 2020-06-29
Fixed wrong result propagation

slinkydeveloper:
Using `Request` was not propagating correctly the result. This fixes it


v2/client/client.go
```diff
@@ -159,7 +159,7 @@ func (c *ceClient) Request(ctx context.Context, e event.Event) (*event.Event, pr
 		// If the protocol returns no error, it is an ACK on the request, but we had
 		// issues turning the response into an event, so make an ACK Result and pass
 		// down the ToEvent error as well.
-		err = fmt.Errorf("%w; failed to convert response into event: %s", protocol.ResultACK, rserr)
+		err = protocol.NewReceipt(true, "failed to convert response into event: %s\n%w", rserr.Error(), err)
```
n3wscott, line 162: ah, I see, we are dropping the err type and we can't understand if it is a special error. got it. 

LGTM

PR #555, opened 2020-07-23
[Validation] fix event validation bug

hongjic:
fix a bug which can cause invalid event to pass validation


slinkydeveloper:
Thank you for your contribution!

PR #557, opened 2020-07-25
Fix error deleting attributes when writing to HTTP messages

ian-mi:
Transformers which delete attributes currently fail when writing a binary message to an HTTP request with the error "invalid CloudEvents value: <nil>".


slinkydeveloper:
Thanks!

Issue #558, opened 2020-07-27
Parsing CloudEvent from HTTP response for 4xx,5xx status codes.


matejvasek:
Spec states `Events can be also be transferred in HTTP responses and with all HTTP status codes that permit payload body transfers.` However when I do request that results in 500:
```go
receivedEvt, resp := cloudEventClient.Request(context.Background(), *evtToSend)
```
only error is set (correctly to "500:"), the `receivedEvt` is nil, despite the fact that response contained body with CloudEvent.

matejvasek:
@n3wscott wdyt?

n3wscott:
This should work if the headers are correct as well, can you make a little demo or paste the raw http body of the response?

matejvasek:
Response as seen in WireShark:
```
HTTP/1.1 500 Internal Server Error
Date: Mon, 27 Jul 2020 10:43:55 GMT
Content-Type: application/cloudevents+json; charset=utf-8
Server: Kestrel
Transfer-Encoding: chunked

{
  "specversion": "1.0",
  "type": "core.errorresponse",
  "source": "/of",
  "id": "de66d47f-6521-4b86-bc04-0b421e41367b",
  "time": "2020-07-27T10:43:55.9400808Z",
  "data": "{\"ErrorCode\":null,\"ErrorDetails\":\"Internal runtime failure. Please refer to runtime logs for details\"}",
  "datacontenttype": "application/json"
}
```

matejvasek:
@n3wscott I think that this may be deliberate:
somewhere in the code:

```go
	var result protocol.Result
	if resp.StatusCode/100 == 2 {
		result = protocol.ResultACK
	} else {
		result = protocol.ResultNACK
	}
```
then later

``` go
	if !protocol.IsACK(err) {
		return nil, err
	}
```

but doesn't that go against spec `Events can be also be transferred in HTTP responses and with all HTTP status codes that permit payload body transfers.` ?

n3wscott:
in the case of the above event and the caller code of:

```
receivedEvt, resp := cloudEventClient.Request(context.Background(), *evtToSend)
```

I would expect `resp` to be a NACK, but also `receivedEvt` to be the full event in response. Let me poke around...

n3wscott:
YES, it is a simple fix.. 😅 one sec

n3wscott:
[...]

PR #559, opened 2020-07-27
Fixing nil http.Result ack compare

n3wscott:
fixes https://github.com/cloudevents/sdk-go/issues/558

I found two issues in the codebase:

1. the client was using the old style `!IsACK` to understand if it should process the response into a CloudEvent. The thinking is if it was not an ACK, then there must not be any event to parse. This logic is incorrect, it should test for `IsUndelivered`. If the event never left, then it can't have a response.
2. Nil http protocol `Result`s are not treated as ACK.

Running the `samples/http/responder` and `samples/http/requester` now get the expected output:

```
2020/07/27 14:16:18 Event sent at 2020-07-27 14:16:18.236459 -0700 PDT m=+3.540167810
2020/07/27 14:16:18 Response status code 500
Response:
Validation: valid
Context Attributes,
  specversion: 1.0
  type: samples.http.mod7
  source: /mod7
  subject: https://github.com/cloudevents/sdk-go/v2/samples/requester#7
  id: 10a6563c-6400-4782-8654-39180bf8f080
  time: 2020-07-27T21:16:18.236026Z
  datacontenttype: application/json
Data,
  {
    "id": 7,
    "message": "mod 7 has issues!"
  }
```

### DIFF
--- a/samples/http/requester/main.go
+++ b/samples/http/requester/main.go
@@ -73,22 +73,19 @@ func main() {
 					ctx = cloudevents.WithEncodingStructured(ctx)
 				}
 
-				if resp, res := c.Request(ctx, event); cloudevents.IsUndelivered(res) {
-					log.Printf("Failed to request: %v", res)
-				} else if resp != nil {
+				resp, result := c.Request(ctx, event)
+				if cloudevents.IsUndelivered(result) {
+					log.Printf("Failed to deliver request: %v", result)
+				} else {
+					// Event was delivered, but possibly not accepted and without a response.
+					log.Printf("Event delivered at %s, Acknowledged==%t ", time.Now(), cloudevents.IsACK(result))
 					var httpResult *cehttp.Result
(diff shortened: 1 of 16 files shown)


samples/http/requester/main.go
```diff
@@ -76,22 +76,22 @@ func main() {
...
-					fmt.Printf("Got Event Response Context: %+v\n", resp.Context)
-					data := &Example{}
-					if err := resp.DataAs(data); err != nil {
-						fmt.Printf("Got Data Error: %s\n", err.Error())
-					}
-					fmt.Printf("Got Response Data: %+v\n", data)
-					fmt.Printf("----------------------------\n")
-				} else {
-					// Parse result
 					var httpResult *cehttp.Result
```
slinkydeveloper, line 79: Can you add a little comment here?

> If event is delivered then the result can be parsed as http result

And later

> response could contain or not an event

PR #561, opened 2020-07-29
Avoid Copying Event Data When Possible in binding.ToEvent

ian-mi:
Protocols which already buffer the event data can directly pass that buffer to the event created by binding.ToEvent. This change uses coercion from bytes.Buffer to skip copying the event data. Messages which do not wish to allow sharing of buffered event data, e.g. buffered messages which use a byte pool, should use bytes.Reader instead.


n3wscott:
LGTM

PR #565, opened 2020-07-31
Do not copy bytes.Buffer when writing a pubsub message

ian-mi:
Applies the optimization from #561 to pubsub messages, allowing them to avoid copying data when given a bytes.Buffer.


johejo:
It looks like bytes.Buffer is not needed.

```go
func (b *pubsubMessagePublisher) SetData(reader io.Reader) error {
	data, err := ioutil.ReadAll(reader)
	if err != nil {
		return err
	}
	b.Data = data
	return nil
}
```

ian-mi:
AFAICT, util.ReadAll has no special handling of bytes.Buffer. The purpose of this change is to share a buffer between the old and new messages when possible.

n3wscott:
LGTM

Issue #570, opened 2020-08-22
Reply validation formatting


odacremolbap:
There seems to be a validation error misformat at the response validation, this line:

https://github.com/cloudevents/sdk-go/blob/3db01d4122c22c0d825ecf0ec776e8520d24af2c/v2/client/invoker.go#L74

is logging this:
 
```log
{"level":"error","ts":1598035219.3899446,"logger":"fallback","caller":"client/invoker.go:74","msg":"cloudevent validation failed on response event: %!w(event.ValidationError=map[type:0xc000b36ab0])","stacktrace":"github.com/cloudevents/sdk-go/v2/client.(*receiveInvoker).Invoke\n\tgithub.com/cloudevents/sdk-go/v2@v2.0.0/client/invoker.go:74\ngithub.com/cloudevents/sdk-go/v2/client.(*ceClient).StartReceiver.func3\n\tgithub.com/cloudevents/sdk-go/v2@v2.0.0/client/client.go:235"}
```

Can be easily tested by starting a receiver using one of the signatures that allows replies:

```golang
func dispatch(ctx context.Context, event cloudevents.Event) (*cloudevents.Event, error) {
   	out := cloudevents.NewEvent()
        
        // do not inform mandatory fields for the out object
        
        return out, nil
}
```

PR #571, opened 2020-08-22
Log: fix response log formatting

odacremolbap:
Fixes https://github.com/cloudevents/sdk-go/issues/570

Logger should use `%v` instead of `%w`


n3wscott:
LGTM

Thanks ! wrong place to wrap 😅

PR #573, opened 2020-08-26
Add ochttp middleware for W3C tracecontext headers in default client

### LINKED ISSUE #572: Enable Trace Context Propagation in the Default HTTP Client
For v2 it was decided that trace context propagation should be opt-in rather than opt-out. Over HTTP, however, there should be little downside to enabling trace context propagation using the W3C tracecontext headers as this does not require modifying event extensions. Therefore I propose that tracing middleware to the HTTP protocol used by NewDefaultClient.

ian-mi:
Fixes #572

### DIFF
--- a/v2/client/client_default.go
+++ b/v2/client/client_default.go
@@ -12,7 +12,7 @@ import (
 // client, all outbound events will have a time and id set if not already
 // present.
 func NewDefault() (Client, error) {
-	p, err := http.New()
+	p, err := http.NewObserved()
 	if err != nil {
 		return nil, err
 	}
--- a/v2/protocol/http/protocol.go
+++ b/v2/protocol/http/protocol.go
@@ -15,6 +15,8 @@ import (
 	cecontext "github.com/cloudevents/sdk-go/v2/context"
 	"github.com/cloudevents/sdk-go/v2/event"
 	"github.com/cloudevents/sdk-go/v2/protocol"
+	"go.opencensus.io/plugin/ochttp"
+	"go.opencensus.io/plugin/ochttp/propagation/tracecontext"
 )
 
 const (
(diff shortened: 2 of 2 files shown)


v2/client/client_default.go
```diff
@@ -12,7 +16,17 @@ import (
 // client, all outbound events will have a time and id set if not already
 // present.
 func NewDefault() (Client, error) {
-	p, err := http.New()
+	p, err := http.New(
+		http.WithMiddleware(func(h nethttp.Handler) nethttp.Handler {
+			return &ochttp.Handler{
```
n3wscott, line 21: We do not want to introduce the opencensus libs in the default client, please apply this change to https://github.com/cloudevents/sdk-go/blob/589ccf58301886a6aa2165ba77f16f4c02cf4260/v2/client/client_observed.go#L14

ian-mi: Since this is HTTP specific it probably belongs in the HTTP protocol instead, maybe as a protocol option?

n3wscott: then `http.NewObserved()`?

v2/client/client_default.go
```diff
@@ -12,7 +12,7 @@ import (
 // client, all outbound events will have a time and id set if not already
 // present.
 func NewDefault() (Client, error) {
-	p, err := http.New()
+	p, err := http.NewObserved()
```
slinkydeveloper, line 15: Can we avoid the default client to use the observed protocol? The user can easily create its own protocol and the related client

ian-mi: Preserving trace context is the correct behavior for most users, I expect that there are very few use cases where dropping the incoming trace headers or ignoring the tracecontext in ctx is desirable. Using the NewObserved protocol will therefore create a more useful default behavior.
Note that observed is not really the best name for this protocol as trace propagation is orthogonal to exporting telemetry, I've chosen this name mostly for consistency with the observed client.

slinkydeveloper: My point is that we're introducing an hidden dependency and as @n3wscott pointed out https://github.com/cloudevents/sdk-go/pull/573#discussion_r478669817 that's something we wish to avoid. I see that this is the correct behaviour, but there might be users that are not using tracing, because the work to enable it is minimal (and already documented in examples etc) I prefer to avoid this change

ian-mi: I'm not sure I follow. The opencensus libs are already used by client.NewObserved which is what the default client uses, so using the http.NewObserved protocol is not adding a new indirect dependency to the default client. There are no module level dependency changes as opencensus is already a dependency there as well.

n3wscott:
Let's let this in, and we might change the default at a later date to make it more clean of a cut between observability and the core lib.

LGTM

Issue #569, opened 2020-08-21
CE receiver client can only handle GOMAXPROCS concurrent requests


liu-cong:
While debugging an issue in knative-gcp with [@tommyreddad](https://github.com/tommyreddad), we found that the CE v2 http client can only support up to GOMAXPROCS concurrent requests. As a result, if some receiver `fn` are taking long time to finish, they essentially block other event processing.

The issue is that in `client.go`, it spawns GOMAXPROCS goroutines. Each goroutine is an infinite loop to call the `invoker` to process events. The problem is that `invoker` is called synchronously, therefore each goroutine up to GOMAXPROCS is essentially processing events one by one.

https://github.com/cloudevents/sdk-go/blob/3db01d4122c22c0d825ecf0ec776e8520d24af2c/v2/client/client.go#L211-L241

Another potential performance blocker is that the incoming channel in http protocol is blocking. I believe a buffered channel is better.

https://github.com/cloudevents/sdk-go/blob/3db01d4122c22c0d825ecf0ec776e8520d24af2c/v2/protocol/http/protocol.go#L73-L76

### How to reproduce

Run the example receiver code: https://gist.github.com/liu-cong/a132fe8629715419c7a58ee20001d071
It retrieves the `Ce-Sleep` attribute from the event and sleep for that period of time.

Then send a few events to the receiver using curl, (note that we set `Ce-Sleep` to 1min).
```
curl -v localhost:8080
  -X POST \
  -H "Ce-Sleep: 1m" \
  -H "Ce-Id: 1" \
  -H "Ce-Specversion: 1.0" \
  -H "Ce-Type: type" \
  -H "Ce-Source: source" \
  -H "Content-Type: application/json" \
  -d '{"msg":"Test!"}
```

Once you reach GOMAXPROCS events, the `receive` function for next event will be blocked until the sleep period is done.

n3wscott:
Any suggestions on fixing? Can you increase your `GOMAXPROCS`?

liu-cong:
If I understand correctly, GOMAXPROCS configures how many CPUs to use, typically equal to or less than the number of CPU cores.

I think a proper fix is to run the `receiver` function in a goroutine instead of blocking. This is how HTTP server handles requests.

n3wscott:
Interesting. Ok let me try it locally and see what's up.

PR #574, opened 2020-08-28
Client should not block on GOMAXPROCS receivers.

n3wscott:
Fixes: https://github.com/cloudevents/sdk-go/issues/569

- Change client to fork off the call to invoker inside the polling loop.
- adding a test to validate the blocking GOMAXPROCS is an issue in the client


n3wscott:
@liu-cong can you try this patch and let me know if it solves your issue?

v2/client/client.go
```diff
@@ -233,9 +233,14 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 					continue
 				}
 
-				if err := c.invoker.Invoke(ctx, msg, respFn); err != nil {
-					cecontext.LoggerFrom(ctx).Warnf("Error while handling a message: %s", err)
-				}
+				// Do not block on the invoker.
+				wg.Add(1)
```
slinkydeveloper, line 237: Do we really need to synchronize the `Invoke` on the wg? If `Invoke` ends after these goroutines loop, is this a potential issue?

n3wscott: it just lets us know to block if invoke is happening when the process is spinning down

slinkydeveloper: that's my point, do we need to wait for invoke to finish before spinning down these goroutines? isn't this done automatically by the runtime?

n3wscott: It is not done automatically. We would return from this loop and exit the process before the invokers had a chance to finish on a soft sig-term

n3wscott:
Ping on @liu-cong

@slinkydeveloper want to LGTM?

slinkydeveloper (approved):

liu-cong:
/lgtm Thanks @n3wscott for the quick fix!

I haven't tried this patch. However, I previously tried a similar/same patch locally so I believe it will work as expected :)

PR #575, opened 2020-08-28
Add nats.io jetstream pull consumer support

tamalsaha:
Signed-off-by: Tamal Saha <tamal@appscode.com>


tamalsaha:
nats.io jetstream supports 2 types of consumers: (1) push based consumer and (2) pull-based consumer. The push based consumer is currently supported by cloudevents go-sdk. But it is lacking support for pull-based consumer. This pr adds support for pull-based consumer.

Issue: https://github.com/cloudevents/sdk-go/issues/258#issuecomment-601693073

Lib: https://github.com/nats-io/jsm.go#consuming

Few things to note:

- I passed the `stream` name via Options.
- I changed the meaning of `subject` to `consumer` in the `Subscribe` method.
- Should the timeout for `NextMsg` be configurable?
- I am going to add test if this general structure look acceptable.

cc: @ripienaar , I would appreciate any comments from you or nats team on this pr.

n3wscott:
SWEEET! Will look next week.

protocol/nats/v2/subscriber.go
```diff
@@ -1,20 +1,27 @@
...
 import (
+	"github.com/nats-io/jsm.go"
 	"github.com/nats-io/nats.go"
+	"time"
 )
 
 // The Subscriber interface allows us to configure how the subscription is created
 type Subscriber interface {
-	Subscribe(conn *nats.Conn, subject string, cb nats.MsgHandler) (*nats.Subscription, error)
+	Subscribe(conn *nats.Conn, subject string, cb nats.MsgHandler) (Subscription, error)
```
n3wscott, line 11: This change will require a v3.0.0 semver bump, and break the downstream integrators.

Can you think of a way to do this change without changing the API? Do we need to introduce a new protocol type or option for NATS?

n3wscott:
Ping on the question around another way to add this without changing the API?

tamalsaha:
We have decided to use the Event format but use the nats libraries directly for our use-case. This gives us maximum flexibility to use various nats features. So, this pr is not needed for us anymore . Closing!

PR #581, opened 2020-09-09
Fix client context cancellation

### LINKED ISSUE #580: pubsub receiver does not return EOF when the context is cancelled
`pubsub.Receiver` does not return `io.EOF` when the context is cancelled. That makes `client.StartReceiver` keeps retrying and never ends.

https://github.com/cloudevents/sdk-go/blob/v2.2.0/v2/client/client.go#L231-L234

inigohu:
fixes #580

Signed-off-by: Iñigo Horcajo <inigohu@gmail.com>

### DIFF
--- a/v2/client/client.go
+++ b/v2/client/client.go
@@ -228,6 +228,10 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 					return
 				}
 
+				if err == context.Canceled { // Context canceled
+					return
+				}
+
 				if err != nil {
 					cecontext.LoggerFrom(ctx).Warnf("Error while receiving a message: %s", err)
 					continue


v2/client/client.go
```diff
@@ -228,6 +228,10 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 					return
```
slinkydeveloper, line 228: That's weird, when the context is closed the receiver or the responder returns `io.EOF`. Which protocol are you using? To me it sounds like a problem of the protocol you're using which doesn't properly fullfill the contract.

inigohu: I'm using the pubsub protocol

slinkydeveloper: Then the pubsub protocol is bugged, because it should return io.EOF when the context is closed. Please change name and description of #580 so the pubsub protocol maintainers can take care of it

inigohu: Issue #580 updated and created a new PR #582

slinkydeveloper:
closing in favour of https://github.com/cloudevents/sdk-go/pull/581

PR #587, opened 2020-09-16
Fix handler input types variance

matejvasek:
Input types should be contravariant or invariant.

### DIFF
--- a/v2/client/receiver.go
+++ b/v2/client/receiver.go
@@ -113,16 +113,16 @@ func (r *receiverFn) validateInParamSignature(fnType reflect.Type) error {
 	switch fnType.NumIn() {
 	case 2:
 		// has to be (context.Context, event.Event)
-		if !fnType.In(1).ConvertibleTo(eventType) {
-			return fmt.Errorf("%s; cannot convert parameter 2 from %s to event.Event", inParamUsage, fnType.In(1))
+		if !eventType.ConvertibleTo(fnType.In(1)) {
+			return fmt.Errorf("%s; cannot convert parameter 2 to %s from event.Event", inParamUsage, fnType.In(1))
 		} else {
 			r.hasEventIn = true
 		}
 		fallthrough
 	case 1:
-		if !fnType.In(0).ConvertibleTo(contextType) {
-			if !fnType.In(0).ConvertibleTo(eventType) {
-				return fmt.Errorf("%s; cannot convert parameter 1 from %s to context.Context or event.Event", inParamUsage, fnType.In(0))
+		if !contextType.ConvertibleTo(fnType.In(0)) {
+			if !eventType.ConvertibleTo(fnType.In(0)) {
+				return fmt.Errorf("%s; cannot convert parameter 1 to %s from context.Context or event.Event", inParamUsage, fnType.In(0))
 			} else if r.hasEventIn {
 				return fmt.Errorf("%s; duplicate parameter of type event.Event", inParamUsage)
 			} else {
--- a/v2/client/receiver_test.go
+++ b/v2/client/receiver_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"testing"
+	"time"
 
 	"github.com/cloudevents/sdk-go/v2/event"
 	"github.com/cloudevents/sdk-go/v2/protocol"
@@ -38,6 +39,9 @@ func TestReceiverFnValidTypes(t *testing.T) {
 		"ctx in, Event+Result out":       func(context.Context) (*event.Event, protocol.Result) { return nil, nil },
 		"Event in, Event+Result out":     func(event.Event) (*event.Event, protocol.Result) { return nil, nil },
 		"ctx+Event in, Event+Result out": func(context.Context, event.Event) (*event.Event, protocol.Result) { return nil, nil },
+
+		"input contravariance; may accept supertype": func(event.EventReader) {},
+		"output covariance; may return subtype":      func() *myErr { return nil },
 	} {
 		t.Run(name, func(t *testing.T) {
 			if _, err := receiver(fn); err != nil {
@@ -65,6 +69,8 @@ func TestReceiverFnInvalidTypes(t *testing.T) {
 		"Event as non-ptr out":     func() event.Event { return event.Event{} },
 		"extra Event in":           func(event.Event, event.Event) {},
 		"not a function":           map[string]string(nil),
+
+		"input covariance; must not accept subtype": func(*myCtx) {},
 	} {
 		t.Run(name, func(t *testing.T) {
 			if _, err := receiver(fn); err == nil {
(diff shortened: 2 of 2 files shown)


v2/client/receiver_test.go
```diff
@@ -38,6 +39,9 @@ func TestReceiverFnValidTypes(t *testing.T) {
 		"ctx in, Event+Result out":       func(context.Context) (*event.Event, protocol.Result) { return nil, nil },
 		"Event in, Event+Result out":     func(event.Event) (*event.Event, protocol.Result) { return nil, nil },
 		"ctx+Event in, Event+Result out": func(context.Context, event.Event) (*event.Event, protocol.Result) { return nil, nil },
+
+		"input contravariance; may accept supertype": func(event.EventReader) {},
```
matejvasek, line 43: @n3wscott  Maybe this is too loose/flexible, I mean: maybe we don't want to allow (sub|super) types. That would mean replace `ConvertibleTo()` by simple equals `==`.

n3wscott: `event.Event` is a struct, not an interface so I think it has to be an `Event` and not an `EventReader`.

https://github.com/cloudevents/sdk-go/blob/master/v2/event/event.go#L11

v2/client/receiver.go
```diff
@@ -113,16 +113,16 @@ func (r *receiverFn) validateInParamSignature(fnType reflect.Type) error {
 	switch fnType.NumIn() {
 	case 2:
 		// has to be (context.Context, event.Event)
-		if !fnType.In(1).ConvertibleTo(eventType) {
-			return fmt.Errorf("%s; cannot convert parameter 2 from %s to event.Event", inParamUsage, fnType.In(1))
+		if !eventType.ConvertibleTo(fnType.In(1)) {
```
n3wscott, line 116: Thinking about this more, the usecase is the function consumer would like to accept one of the interfaces that event.Event implements? Like even a subset? interesting. I had not considered this usage.

v2/client/receiver_test.go
n3wscott: oooh I see, you are asking for just a subset of `event.Event` and in this case it is the `event.EventReader` interface but you could invent your own...

v2/client/receiver_test.go
```diff
@@ -65,6 +69,8 @@ func TestReceiverFnInvalidTypes(t *testing.T) {
 		"Event as non-ptr out":     func() event.Event { return event.Event{} },
 		"extra Event in":           func(event.Event, event.Event) {},
 		"not a function":           map[string]string(nil),
+
+		"input covariance; must not accept subtype": func(*myCtx) {},
```
n3wscott, line 73: Agreed, you can't expect a custom type to be created by the SDK.

v2/client/receiver_test.go
```diff
@@ -38,6 +39,9 @@ func TestReceiverFnValidTypes(t *testing.T) {
 		"ctx in, Event+Result out":       func(context.Context) (*event.Event, protocol.Result) { return nil, nil },
 		"Event in, Event+Result out":     func(event.Event) (*event.Event, protocol.Result) { return nil, nil },
 		"ctx+Event in, Event+Result out": func(context.Context, event.Event) (*event.Event, protocol.Result) { return nil, nil },
+
+		"input contravariance; may accept supertype": func(event.EventReader) {},
+		"output covariance; may return subtype": func() *myErr { return nil },
```
n3wscott, line 44: I would not expect this to work. the `protocol.Result` is an interface and what it creates internally will not be cast-able to a `*myEr`r unless it was an interface.

v2/client/receiver_test.go
```diff
@@ -229,3 +235,30 @@ func TestReceiverFnInvoke_5(t *testing.T) {
...
+	panic("implement me")
+}
+
+func (m myCtx) Err() error {
+	panic("implement me")
+}
+
+func (m myCtx) Value(key interface{}) interface{} {
+	panic("implement me")
+}
```
n3wscott, line 264: nit: newline

n3wscott (commented):
Looks good!! Thanks so much, one question around `myErr` and why that worked.

v2/client/receiver_test.go
matejvasek: The `protocol.Result` interface is indeed not convertible to `*myErr`, it is the other way around. The `*myErr` type implements `protocol.Result` and therefore is convertible to it, and that's what is needed for `output types`. It works exactly opposite for `input types`.

matejvasek: I am little surprised that `event.Event` (non-pointer type) is convertible to interface type `event.EventReader`, I totally understand that pointer type `*event.Event` is convertible `event.EventReader`.

matejvasek: I tried to run:
```go
func Handle(ctx context.Context, event event.EventReader) {
	fmt.Println(event.ID())
}
```
and it worked. Not saying it's particularly useful.

v2/client/receiver.go
matejvasek: Yeah this would for example use `event.EventReader` which exposes only part of an event. Not sure if this is useful, I changed this because older version was even less useful or less type safe.

matejvasek:
@n3wscott  why is the CI is failing? I run test locally and it worked.

v2/client/receiver.go
matejvasek: Again am not saying it is much useful, it is simply more type correct/safe, than having that `ConvertibleTo()` otherway around.

v2/client/receiver_test.go
matejvasek: >but you could invent your own
Not sure what you meant by this, AFAIK you cannot implement custom interface for builtin type.

matejvasek: But it might be interesting sometimes if you could.

matejvasek: Fewer languages allow that (e.g. https://clojure.org/reference/protocols).

matejvasek:
After another commit the CI has not been executed at all, weird.

n3wscott:
the CI is failing because of linting issues. 

I will make an action item to move to github actions, makes it more clear.

v2/client/receiver_test.go
n3wscott: I think go does some magic to make non-pointers translate to interfaces when the pointer of that struct does implement the interface.

matejvasek:
To be able to see CI log I need to login into circleci, I don't have an account. I might use github account but it requires some permissions and I am too paranoid to enable them :smile:

v2/client/receiver_test.go
matejvasek: >I think go does some magic to make non-pointers translate to interfaces when the pointer of that struct does implement the interface.

Seems it really does, but only if receivers of given interfaces are all non-pointer, interesting.

matejvasek:
I fixed the CI (go fmt) issue and squashed commits.

n3wscott:
LGTM!!!

thanks for the fix.

PR #589, opened 2020-09-24
StartReceiver returns if it can't start listening

matejvasek:


matejvasek:
It passes CI now, @n3wscott do you thing that the last commit is improvement?

v2/client/client.go
```diff
@@ -244,8 +238,18 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 			}
 		}()
 	}
+
+	// Start the opener, if set.
```
slinkydeveloper, line 242: Why did you moved the opener after the goroutine polling start? I wonder if this breaks some protocols, because they may assume that when you invoke OpenInbound you mutate some internal state which is then queried by Receive

matejvasek: I not only moved it below polling, but I also put it away from gorutine, I did that so that I can share error without introducing channel.
This was older version I force pushed away, I can use that if you think it's more safe.
```diff
diff --git a/v2/client/client.go b/v2/client/client.go
index f670306..d9d5b5e 100644
--- a/v2/client/client.go
+++ b/v2/client/client.go
@@ -170,6 +170,9 @@ func (c *ceClient) Request(ctx context.Context, e event.Event) (*event.Event, pr
 // StartReceiver sets up the given fn to handle Receive.
 // See Client.StartReceiver for details. This is a blocking call.
 func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
+       ctx, cancel := context.WithCancel(ctx)
+       defer cancel()
+
        c.receiverMu.Lock()
        defer c.receiverMu.Unlock()
 
@@ -198,11 +201,15 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
        }()
 
        // Start the opener, if set.
+       openInboundErr := make(chan error, 1)
        if c.opener != nil {
                go func() {
-                       if err := c.opener.OpenInbound(ctx); err != nil {
+                       err = c.opener.OpenInbound(ctx)
+                       if err != nil {
                                cecontext.LoggerFrom(ctx).Errorf("Error while opening the inbound connection: %s", err)
+                               cancel()
                        }
+                       openInboundErr <- err
                }()
        }
 
@@ -245,7 +252,7 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
                }()
        }
        wg.Wait()
-       return nil
+       return <-openInboundErr
 }
 
 // noRespFn is used to simply forward the protocol.Result for receivers that aren't responders

```

>I wonder if this breaks some protocols, because they may assume that when you invoke OpenInbound you mutate some internal state which is then queried by Receive

If it were so, then there would be potential bug even before, since `OpenInbound()` is called in go block it may be invoked after the polling depending on scheduler.

slinkydeveloper: I think this other solution might be better, but before rushing in implementing it, i would double check all our protocol implementations to find out if there is a hidden relationship between OpenInbound() and Receive()

> If it were so, then there would be potential bug even before, since OpenInbound() is called in go block it may be invoked after the polling (depending on how gorutines/threads are scheduled).

I see, it may be that we never found out it because it's far more likely that the runtime executes that goroutine directly than invoking the other goroutines of the polling first

matejvasek: Yes, such a scheduling is really unlikely (at leas on modern multicore machines)

matejvasek: I tried to put some debugging output to see what is called first. To my surprise on my machine `Respond` is called at least once before `OpenInbound` in most cases. But then again, those debugging prints might have affected result.

n3wscott: moving where the go fork is should make no change

n3wscott (approved):
LGTM

PR #596, opened 2020-10-07
add ability to specify custom func for retry policy

### LINKED ISSUE #595: Allow customization of which errors to retry.
https://github.com/cloudevents/sdk-go/blob/master/v2/protocol/http/protocol_retry.go#L45

These are hard coded, would it be possible to pass this in somehow. I think using context might be a bit hacky, but a thought.

vaikas:
Fixes #595 

Introduce an option for overriding the retry policy.


slinkydeveloper (approved):
LGTM, thanks!

PR #597, opened 2020-10-08
pass http.Option through

vaikas:
In order to pass HTTP.Option into the client, I basically had to duplicate the NewDefault as seen here:
https://github.com/knative-sandbox/eventing-rabbitmq/pull/113/files#diff-e45c0bc45ea5877a2e315d983df2d17eR77

If we just allow the protocol Option optionally here (so doesn't change the signature), I could use this method as is.

WDYT?


slinkydeveloper:
It's a vararg, so it's fine and it doesn't break. LGTM

slinkydeveloper (approved):

PR #600, opened 2020-10-08
Use errgroup to wait for subscribers in OpenInbound

### LINKED ISSUE #599: Pubsub OpenInbound Infinite Loop on Context Cancellation
When ctx is cancelled, pubsub OpenInbound will loop infinitely: https://github.com/cloudevents/sdk-go/blob/58f826d67d9177e176e6336703ba21b34a2f970f/protocol/pubsub/v2/protocol.go#L213-L229

ian-mi:
Fixes #599


n3wscott (commented):
LGTM

PR #606, opened 2020-11-09
Websocket protocol binding (only client)

### LINKED ISSUE #598: Implement WebSockets protocol binding
Implement the new https://github.com/cloudevents/spec/blob/master/websockets-protocol-binding.md

slinkydeveloper:
Part of #598

* Implemented websocket protocol binding, only client-side (server-side in a followup)
* Added a sample
* Added a new utils package to encapsulate some common messages (Later we can refactor some nats code to use that, deduplicating code)
* Fixed docs of protocol/inbound.go


protocol/ws/v2/client_protocol.go
```diff
@@ -0,0 +1,131 @@
...
+		})
+	} else {
+		c.receiverLock.Unlock()
+	}
+	return m, err
+}
+
+// UnsafeReceive is like Receive, except it doesn't guard from multiple invocations
+// from different goroutines.
+func (c *ClientProtocol) UnsafeReceive(ctx context.Context) (binding.Message, error) {
```
n3wscott, line 89: might not export this, seems like a footgun

protocol/ws/v2/client_protocol.go
```diff
@@ -0,0 +1,131 @@
...
+		consumeStream(reader)
+		return nil, fmt.Errorf("wrong message type: %s, expected %s", messageType, c.messageType)
+	}
+
+	return utils.NewStructuredMessage(c.format, reader), nil
+}
+
+func consumeStream(reader io.Reader) {
+	//TODO is there a less expensive way to consume the stream?
+	ioutil.ReadAll(reader)
```
n3wscott, line 109: There might be some fancy seek thing we can do later... so we try to find the edge of a single event at a time

slinkydeveloper: We don't need to find the edge of the websocket message, this is already done by the lib itself. But yeah this is open to optimizations

protocol/ws/v2/client_protocol.go
slinkydeveloper: It's not a footgun, on the contrary this is expected for a websocket user to be unable to invoke multiple times from different goroutines a method to read from the same connection, because she/he knows the wasm stream is not multiplexed.

slinkydeveloper: If you feel this is really problematic, we can make it unexported and export it later

n3wscott:
LGTM

I want to see the server side now :D

protocol/ws/v2/subprotocols.go
```diff
@@ -0,0 +1,22 @@
...
+	"nhooyr.io/websocket"
+
+	"github.com/cloudevents/sdk-go/v2/binding/format"
+)
+
+const JsonSubprotocol = "cloudevents.json"
+
+var SupportedSubprotocols = []string{JsonSubprotocol}
+
+func resolveFormat(subprotocol string) (format.Format, websocket.MessageType, error) {
```
zhongduo, line 15: The `format` has already supported different format types, are you planing to support different types? I am wondering if we can leverage or improve the existing framework. 
https://github.com/cloudevents/sdk-go/blob/7bec7ba0a7bbc90a67a5fc594a217ee0b01c1bf7/v2/binding/format/format.go#L66

v2/protocol/inbound.go
```diff
@@ -9,6 +9,7 @@ import (
 // Receiver receives messages.
 type Receiver interface {
 	// Receive blocks till a message is received or ctx expires.
+	// Receive can be invoked safely from different goroutines.
```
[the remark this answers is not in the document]
slinkydeveloper: They already are, otherwise they couldn't pass the integration tests :smile:

protocol/ws/v2/subprotocols.go
slinkydeveloper: In sdk-go we support only the EventFormat atm, so here i'm just mapping to what we support currently

v2/protocol/inbound.go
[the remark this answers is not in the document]
slinkydeveloper: This missing comment was more a mistake than a "new feature"

protocol/ws/v2/subprotocols_test.go
```diff
@@ -0,0 +1,37 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			fmt, messageType, err := resolveFormat(tt.subprotocol)
+			require.NoError(t, err)
+			require.Equal(t, tt.wantFormat, fmt)
+			require.Equal(t, tt.wantMessageType, messageType)
+		})
+	}
+}
+
+func Test_resolveFormat_error(t *testing.T) {
```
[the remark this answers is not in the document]
slinkydeveloper: I have no idea why Goland generates these weeeird names

protocol/ws/v2/subprotocols.go
slinkydeveloper: Ah I see your point now, well this "mapping" of subprotocols is very specific to websockets protocol binding https://github.com/cloudevents/spec/blob/master/websockets-protocol-binding.md#15-cloudevents-subprotocols, i guess this code should live only in the websockets protocol. Also because it uses a websocket type (`websocket.MessageType`)

n3wscott (commented):
/lgtm
/approve

PR #607, opened 2020-11-09
[Braindump] Proposal for redesigning the client interface

slinkydeveloper:
I came out with this proposal after working on #606.

My problem with the `Client` interface is that, while `Send` and `Request` maps the interfaces in protocol, `StartReceiver` doesn't map the semantics of `Receiver` and `Responder`. This is particularly weird with `websocket` for example, when you have a bi-directional client per stream, so you can do something like:

```golang
c.Send()
[...]
c.Receive()
```

In particular, I find that `StartReceiver` is opinionated, maps to the concepts of http more than the concepts of messaging and it's less golang idiomatic than a channel or a method like `Receive`. Another point to consider is that going from the method invocation to the handler mode is relatively straightforward (and we could easily provide the method to do it, like showed in this PR), while going from the handler mode to the method invocation is harder and, generally, more error prone. 

### DIFF
--- a/v2/client/client.go
+++ b/v2/client/client.go
@@ -17,32 +17,52 @@ import (
 )
 
 // Client interface defines the runtime contract the CloudEvents client supports.
-type Client interface {
+type Sender interface {
 	// Send will transmit the given event over the client's configured transport.
 	Send(ctx context.Context, event event.Event) protocol.Result
 
 	// Request will transmit the given event over the client's configured
 	// transport and return any response event.
 	Request(ctx context.Context, event event.Event) (*event.Event, protocol.Result)
+}
 
-	// StartReceiver will register the provided function for callback on receipt
-	// of a cloudevent. It will also start the underlying protocol as it has
-	// been configured.
-	// This call is blocking.
-	// Valid fn signatures are:
-	// * func()
-	// * func() error
-	// * func(context.Context)
-	// * func(context.Context) protocol.Result
-	// * func(event.Event)
-	// * func(event.Event) protocol.Result
-	// * func(context.Context, event.Event)
-	// * func(context.Context, event.Event) protocol.Result
-	// * func(event.Event) *event.Event
-	// * func(event.Event) (*event.Event, protocol.Result)
-	// * func(context.Context, event.Event) *event.Event
-	// * func(context.Context, event.Event) (*event.Event, protocol.Result)
-	StartReceiver(ctx context.Context, fn interface{}) error
+type ReceiverEvent interface {
+	Event() event.Event
+	Ack()
+	Reply(event *event.Event)
+	ReplyWithResult(*event.Event, protocol.Result)
+}
+
+// Client interface defines the runtime contract the CloudEvents client supports.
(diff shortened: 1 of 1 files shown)


v2/client/client.go
```diff
@@ -25,24 +25,32 @@ type Client interface {
...
-	// * func(context.Context) protocol.Result
-	// * func(event.Event)
-	// * func(event.Event) protocol.Result
-	// * func(context.Context, event.Event)
-	// * func(context.Context, event.Event) protocol.Result
-	// * func(event.Event) *event.Event
-	// * func(event.Event) (*event.Event, protocol.Result)
-	// * func(context.Context, event.Event) *event.Event
-	// * func(context.Context, event.Event) (*event.Event, protocol.Result)
-	StartReceiver(ctx context.Context, fn interface{}) error
```
n3wscott, line 45: how about leaving StartReceiver and also add the blocking single pulling functions? 

you still need a hook on client to give it a thread, I am not seeing how that is happening without dramatically changing the contract.

slinkydeveloper: > how about leaving StartReceiver and also add the blocking single pulling functions?

So one of the points with this proposal is that I would love to relax this assumption on the protocol side https://github.com/cloudevents/sdk-go/pull/606/files#diff-484d929efc01852c00d94e8e519500757204bacd6a67224c50db3c805eabc8f5R12 and enforce only in `StartReceiver`, where it really matters. Nor `StartReceiver` or invoking from multiple threads `Receive` makes sense in the context of Websockets.

My worry is that, If we put everything in `Client`, we'll have races everywhere, because we need to handle a lot of different cases: 

* user invokes Receive, it's blocked on Receive but then invokes StartReceiver
* user invokes StartReceiver, then invokes Receive
* OpenInbound for StartReceiver, then user invokes Receive too
* StartReceiver context is canceled, but then user invokes Receive

While, if we make it clear that `StartReceiver` is like one level up/different to the client, e.g. accepting directly the `Protocol` in `StartReceiver`, then we have a clean implementation, both for `StartReceiver` and for `Client.Receive`. And also makes pretty clear to the user the usage aka don't mess up with `Receive` and `StartReceiver` at the same time.

n3wscott: Maybe want you are saying is we need a new implementation for a stream based client. Client is for more like pub/sub style apis, and we can create a new one specific for streams like WS provides?

slinkydeveloper: Pretty much, but in particular my point is that I see the pub/sub APIs as a layer on top of a stream API (in fact, protocol APIs are already stream apis)

n3wscott:
Think about a v3?

slinkydeveloper:
Maybe, let's get the 2.4 out first :smile:

duglin:
@embano1 I'm not sensing a lot of momentum to redesign things right now. You ok with closing this?

embano1:
Ok to close for me

duglin:
Closing due to lack of interest. We can reopen if someone wants to drive it.

Issue #609, opened 2020-11-12
Investigate improving json decoding performance through json streaming


slinkydeveloper:
Golang json module has a streaming parser we can leverage in our UnmarshalJson implementation for CloudEvent.

We should investigate if this is doable and if increases performances of json decoding.

PR #613, opened 2020-11-20
Benchmark Json marshal/unmarshal

slinkydeveloper:
Part of #609 

These benchmarks evaluates the performance of MarshalJSON() and UnmarshalJSON(), used to implement the Json Event Format


n3wscott (approved):
LGTM

Never seen a bad test (ok ok that is a lie, but this is good :D)

PR #616, opened 2020-11-23
URI must be absolute

### LINKED ISSUE #615: Should '#' be a valid entry for dataschema?
This currently gets rejected.

This simple test case currently fails:

```
func TestSimple(t *testing.T) {
	ev := cloudev.NewEvent()
	ev.SetDataSchema("#")
	ev.SetID("id")
	ev.SetSource("mysrc")
	ev.SetType("mytype")

	require.NoError(t, ev.Validate())
}
```

https://github.com/cloudevents/spec/blob/v1.0/spec.md#dataschema
Indicates:

> URI - Absolute uniform resource identifier.
> String encoding: Absolute URI as defined in RFC 3986 Section 4.3.

The RFC is a little hard to follow on this part and it seems like '#' might considered as valid (meaning current URL).
However, in an event that has very limited sense.

This question mainly comes up because I've noticed that Azure Cloudevent Subscription send '#' as dataschema for example:

```json
{
  "id": "uuid",
  "source": "/subscriptions/sub-uuid/resourceGroups/rg/providers/Microsoft.Storage/storageAccounts/acc",
  "specversion": "1.0",
  "type": "Microsoft.Storage.BlobCreated",
  "dataschema": "#",
  "subject": "/blobServices/default/containers/foo/blobs/message",
  "time": "2020-11-20T18:35:53.2272766Z",
  "data": {
    }
  }
}
```

Because the `Event validation` is non flexible it makes it currently impossible to use the SDK with this service.

slinkydeveloper:
Fix #615

Note: This might be seen as a breaking change because the previous validation was wrong, accepting relative uris more than absolute uris

### DIFF
--- a/v2/event/eventcontext_v1.go
+++ b/v2/event/eventcontext_v1.go
@@ -226,10 +226,8 @@ func (ec EventContextV1) Validate() ValidationError {
 	//  OPTIONAL
 	//  If present, MUST adhere to the format specified in RFC 3986
 	if ec.DataSchema != nil {
-		dataSchema := strings.TrimSpace(ec.DataSchema.String())
-		// empty string is not RFC 3986 compatible.
-		if dataSchema == "" {
-			errors["dataschema"] = fmt.Errorf("if present, MUST adhere to the format specified in RFC 3986")
+		if !ec.DataSchema.Validate() {
+			errors["dataschema"] = fmt.Errorf("if present, MUST adhere to the format specified in RFC 3986, Section 4.3. Absolute URI")
 		}
 	}
 
--- a/v2/event/eventcontext_v1_test.go
+++ b/v2/event/eventcontext_v1_test.go
@@ -1,7 +1,6 @@
 package event_test
 
 import (
-	"net/url"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


lahabana:
Thanks! Maybe making the validation optional (and default to the old buggy one) would help with the breaking change?

It would also help with dealing with badly implemented providers (.e.g Azure as in #615)

slinkydeveloper:
> Maybe making the validation optional

The validation is already optional

lahabana:
Yes but it's all or nothing atm either you validate no fields or all the fields.

slinkydeveloper:
> Yes but it's all or nothing atm either you validate no fields or all the fields.

Which I think it's reasonable for our use cases. If you need something more complex, you can wrap Validate in your own running the full validation and then filtering out errors you're not interested to

lahabana:
Seems reasonable indeed. Didn't think that filtering out was actually pretty straightforward as validationError is just a map. Thanks!

v2/event/eventcontext_v1.go
```diff
@@ -226,10 +226,8 @@ func (ec EventContextV1) Validate() ValidationError {
 	//  OPTIONAL
 	//  If present, MUST adhere to the format specified in RFC 3986
 	if ec.DataSchema != nil {
-		dataSchema := strings.TrimSpace(ec.DataSchema.String())
-		// empty string is not RFC 3986 compatible.
-		if dataSchema == "" {
-			errors["dataschema"] = fmt.Errorf("if present, MUST adhere to the format specified in RFC 3986")
+		if !ec.DataSchema.Validate() {
+			errors["dataschema"] = fmt.Errorf("if present, MUST adhere to the format specified in RFC 3986, Section 4.3. Absolute URI")
```
n3wscott, line 230: Interesting, this is a change to the spec since we started.

n3wscott (approved):
LGTM

PR #617, opened 2020-11-23
Remove regex to check the extension name

slinkydeveloper:
There's no need to use a regex for this simple check, this PR removes the regex and replaces it with a simple logic, which is a bit more efficient and allocates less

### DIFF
--- a/v2/event/extensions.go
+++ b/v2/event/extensions.go
@@ -25,7 +25,7 @@ func IsExtensionNameValid(key string) bool {
 		return false
 	}
 	for _, c := range key {
-		if !((c >= 'a' && c <= 'z') || (c >= '0' && c <= '9')) {
+		if !((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9')) {
 			return false
 		}
 	}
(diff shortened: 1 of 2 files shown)


v2/event/extensions.go
```diff
@@ -21,4 +20,14 @@ func caseInsensitiveSearch(key string, space map[string]interface{}) (interface{
 	return nil, false
 }
 
-var IsAlphaNumeric = regexp.MustCompile(`^[a-zA-Z0-9]+$`).MatchString
+func IsExtensionNameValid(key string) bool {
+	if len(key) < 1 || len(key) > 20 {
+		return false
+	}
+	for _, c := range key {
+		if !((c >= 'a' && c <= 'z') || (c >= '0' && c <= '9')) {
```
n3wscott, line 28: A-Z is allowed and ignored, this change will be breaking for anyone who is using mixed-case keys and letting the sdk lowercase it.

slinkydeveloper: Should we keep the uppercase behaviour or remove it? If we want to keep it, then i'll just add it to the condition here

n3wscott (commented):
/lgtm
/approve

PR #627, opened 2020-11-25
Add test for corner-case of having data_base64 as an extension in V03

markusthoemmes:
Adds a test that catches the corner-case of having a `data_base64` as an extension in V03.

Signed-off-by: Markus Thömmes <markusthoemmes@me.com>

### DIFF
--- a/v2/event/event_unmarshal_test.go
+++ b/v2/event/event_unmarshal_test.go
@@ -829,7 +829,7 @@ func TestUnmarshalWithOrdering(t *testing.T) {
 				DataBase64:  true,
 			},
 		},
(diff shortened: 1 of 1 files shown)


v2/event/event_unmarshal_test.go
```diff
@@ -829,6 +829,34 @@ func TestUnmarshalWithOrdering(t *testing.T) {
 				DataBase64:  true,
 			},
 		},
+		"base64 json encoded data v0.3 with data/data_base64 -> datacontenttype -> specversion -> datacontentencoding": {
```
slinkydeveloper, line 832: Can you do a similar one without a base64 data (e.g. like struct data)

markusthoemmes: Done!

PR #632, opened 2020-11-26
Completely remove tokenQueue

markusthoemmes:
This is not a performance enhancement. Results are precisely equal. It is however an enhancement in readability and explicitness imo, which is why I send it anyway.

Signed-off-by: Markus Thömmes <markusthoemmes@me.com>


slinkydeveloper (approved):
Simpler to read, i like it

PR #621, opened 2020-11-25
Implemented streaming writer

slinkydeveloper:
Like #618, but from event to bytes.

A thing to notice of this streaming writer is that it writes the event fields in an order such that the parser of #618 can efficiently parse it, without buffering some fields.

Note that the marshal tests and others had to be changed because they were doing a string comparison, but the new marshaller writes the elements in a different order. Also, the unmarshalling to map uses always floats even for integers. But, semantically speaking, **there is no diff with the previous code**. 

Results:

```
name                                                                               old time/op    new time/op    delta
Marshal/number_data_v1.0-8                                                           17.1µs ± 5%     5.9µs ± 4%  -65.44%  (p=0.008 n=5+5)
Marshal/struct_data_v0.3-8                                                           27.5µs ± 3%     9.2µs ± 2%  -66.52%  (p=0.008 n=5+5)
Marshal/nil_data_v0.3-8                                                              26.0µs ± 0%     9.0µs ± 1%  -65.39%  (p=0.008 n=5+5)
Marshal/string_data_v0.3-8                                                           28.2µs ± 0%     9.2µs ± 2%  -67.30%  (p=0.008 n=5+5)
Marshal/struct_data_v1.0-8                                                           29.0µs ± 0%     8.3µs ± 0%  -71.26%  (p=0.016 n=5+4)
Marshal/nil_data_v1.0-8                                                              26.5µs ± 0%     8.6µs ± 4%  -67.48%  (p=0.008 n=5+5)
Marshal/string_data_v1.0-8                                                           28.9µs ± 1%     8.8µs ± 1%  -69.55%  (p=0.008 n=5+5)
Marshal/base64_json_encoded_data_v1.0-8                                              19.2µs ± 0%     5.9µs ± 1%  -68.99%  (p=0.008 n=5+5)

name                                                                               old alloc/op   new alloc/op   delta
Marshal/number_data_v1.0-8                                                           5.16kB ± 0%    3.88kB ± 0%  -24.70%  (p=0.008 n=5+5)
Marshal/struct_data_v0.3-8                                                           8.76kB ± 0%    4.48kB ± 0%     ~     (p=0.079 n=4+5)
Marshal/nil_data_v0.3-8                                                              7.15kB ± 0%    4.55kB ± 0%  -36.34%  (p=0.029 n=4+4)
Marshal/string_data_v0.3-8                                                           8.76kB ± 0%    4.45kB ± 0%  -49.16%  (p=0.008 n=5+5)
[...]

PR #633, opened 2020-11-30
[WIP] Format now supports streaming

slinkydeveloper:
Depends on #621

Format now uses io.Reader/Writer, allowing binding.ToEvent and binding.ToMessage to use the new json streaming reader/writer

**Note:** this is breaking, although it's modifying an "internal" API (yeah it's public, but it's not really documented for users to use it)


duglin:
Closing for now - we can reopen it if someone would like to continue to work on it

Issue #619, opened 2020-11-24
Migrate away from direct dependency on observibility.


n3wscott:
Today we have observibility code peppered throughout the codebase, and this is useful sometimes, a lot of the usage is likely just noise and slowdown for the integrator. 

What if we added a go channel and a light weight observability structure to capture the current measurements?

Code like this:

```go
// UnmarshalJSON implements the json unmarshal method used when this type is
// unmarshaled using json.Unmarshal.
func (e *Event) UnmarshalJSON(b []byte) error {
	_, r := observability.NewReporter(context.Background(), eventJSONObserved{o: reportUnmarshal})
	err := ReadJson(e, bytes.NewReader(b))

	// Report the observable
	if err != nil {
		r.Error()
	} else {
		r.OK()
	}
	return err

}
```

becomes:

```go
// UnmarshalJSON implements the json unmarshal method used when this type is
// unmarshaled using json.Unmarshal.
func (e *Event) UnmarshalJSON(b []byte) error {
	r := observability.Vent(<some little struct>)

	err := ReadJson(e, bytes.NewReader(b))

	// Report the observable
	if err != nil {
		r.Error()
	} else {
		r.OK()
	}
	return err

}
```

And internally, `Vent' pushes a start span onto a go chan, and `Error` or `OK` closes that span via that same go chan using callbacks to hold the context.

PR #634, opened 2020-12-02
Abstraction for metrics and tracing

slinkydeveloper:
The goal of this PR is to introduce an abstraction to remove from the core sdk module the dependencies over metrics and tracing. Part of #619

## Content of this PR:

* the core module doesn't depend anymore on opencensus
* the core module provides an abstraction to plug tracing and metrics in send/receive/request, called `ObservabilityService`
* there is a new separate module that provides the same support as before for opencensus + extras
* Observability is now used only in Client and not anymore in marshalling/unmarshalling hot paths
* Fixed some wrong context propagation with http using the client receiver

## Breaking changes:

Users of metrics and tracing will have to modify their code to support this change. Now they won't need to use `client.NewObservable` and `http.NewObservable`, but they'll must use the observability service and the http protocol factory from the opencensus package

Deprecated (and short-circuited functions):

* client.NewObservable (now it's like client.New)
* http.NewObservable (now it's like http.New)

Removed functions:

* Most of the stuff in v2/observability module which were exposing opencensus types
* All the `ObsEncode` and `ObsDecode`
* Some methods on `DistributedTracingExtension` which were exposing opencensus types (which were implementing the distributed tracing extension in the wrong way https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md#using-the-distributed-tracing-extension)


v2/client/http_receiver.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func NewHTTPReceiveHandler(ctx context.Context, p *thttp.Protocol, fn interface{}) (*EventReceiver, error) {
-	invoker, err := newReceiveInvoker(fn)
+	invoker, err := newReceiveInvoker(fn, noopObservabilityService{})
```
n3wscott, line 12: We should get ahead of this and add a observability.Get() that defaults to `noopObservabilityService` here and everywhere else. We can do this as a follow up but it should happen before we cut

v2/client/client_default.go
```diff
@@ -5,19 +5,17 @@ import (
...
 
 // NewDefault provides the good defaults for the common case using an HTTP
-// Protocol client. The http transport has had WithBinaryEncoding http
-// transport option applied to it. The client will always send Binary
-// encoding but will inspect the outbound event context and match the version.
+// Protocol client.
 // The WithTimeNow, and WithUUIDs client options are also applied to the
 // client, all outbound events will have a time and id set if not already
 // present.
 func NewDefault(opts ...http.Option) (Client, error) {
```
n3wscott, line 12: I wonder if we should deprecate this method and comment how to get the same result with `client.NewHTTP` and then alias that to `cloudevents.NewHTTPClient`

v2/client/http_receiver.go
slinkydeveloper: sure

n3wscott:
LGTM
APPROVE

Issue #636, opened 2020-12-08
Opener always returns "error while opening the inbound connection" after context cancellation


antoineco:
Given the following client:

```go
func cloudeventsClient(idleConnTimeout time.Duration) (cloudevents.Client, error) {
        t := http.DefaultTransport.(*http.Transport)
        t.IdleConnTimeout = idleConnTimeout

        p, err := cehttp.New(cehttp.WithRoundTripper(t))
        if err != nil {
                return nil, fmt.Errorf("creating cehttp.Protocol: %w", err)
        }

        return cloudevents.NewClient(p)
}
```

And the following code to run a receiver using that client:

```go
err := cli.StartReceiver(ctx, receiveFn)
```

When `ctx` get cancelled (upon a normal and expected termination), `StartReceiver` seems to consistently return the error:

```
error while opening the inbound connection: context deadline exceeded
```

https://github.com/cloudevents/sdk-go/blob/f0303ff82cc2951c2dc089f1f87bc4ea9dd6eac9/v2/client/client.go#L242-L248

The default shutdown timeout is 1 min. By looking at my logs, I can confirm the duration between the context cancellation and the error is exactly 1 min:

```
2020/12/08 11:33:49 Stopped event recorder
2020/12/08 11:34:49 Failure during runtime of CloudEvents handler: error while opening the inbound connection: context deadline exceeded
```

I need to investigate why the server can not shut down gracefully.

antoineco:
It seems like setting `IdleConnTimeout` to `5s` was causing some harm. Things seem fine after increasing it to something less aggressive (`30s`).

Closing as a non-issue.

PR #637, opened 2020-12-08
Provide additional context in shutdown errors

antoineco:
I found out while investigating #636 that the error received from `errChan` was discarded during a server shutdown. I used the occasion to wrap those errors with some additional context to ease troubleshooting a little bit.

Example:

`2020/12/08 12:55:49 Failure during runtime of CloudEvents handler: error while opening the inbound connection: combined error during shutdown of HTTP server: shutting down HTTP server: context deadline exceeded, server failed during shutdown: http: some terrible error`


slinkydeveloper (approved):
Sounds good, thanks for the PR!

PR #639, opened 2020-12-11
Remove validation from the string representation

slinkydeveloper:
The reasons for this change are:

* the validation it's not related to string representation at all.
* avoid the overhead of the validation when logging events. One that needs both just invokes both methods


n3wscott (commented):
LGTM

n3wscott:
I think the tests are looking for that string.

n3wscott:
LGTM

PR #641, opened 2020-12-15
Add required call to SetID when marshalling JSON CloudEvent.

### LINKED ISSUE #640: README deserialization code missing SetID
The demo code for deserializing CloudEvents in [`README.md`](https://github.com/cloudevents/sdk-go#serializedeserialize-a-cloudevent) is missing a call to `event.SetID()`.  Without this, the call to `json.Marshal` will error out with "event.SetId undefined".

Suggest adding an example of setting the ID to some static string: `e.SetID("32943bac6fea")`

scsibug:
Fixes #640.  Simple addition of SetID so that this code can be used as-is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ To marshal a CloudEvent into JSON:
 
 ```go
 event := cloudevents.NewEvent()
+event.SetID("example-uuid-32943bac6fea")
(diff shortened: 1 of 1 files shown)


slinkydeveloper:
Thanks for your contribution!

Issue #643, opened 2020-12-15
Kafka Receiver: Data Loss


pastequo:
Hello again,

Continuing the (small) discussion here: https://github.com/cloudevents/sdk-go/issues/642

The client [here](https://github.com/cloudevents/sdk-go/blob/ab0ab9f7e0824a8ce40f394b8871de0e8465d725/v2/client/client.go#L203-L240) starts a new goroutine each time an event is pulled. At the end of "invoke" [here](https://github.com/cloudevents/sdk-go/blob/ab0ab9f7e0824a8ce40f394b8871de0e8465d725/v2/client/invoker.go#L42), Finish() is called. For kafka receiver it calls MarkMessage() [here](https://github.com/cloudevents/sdk-go/blob/ab0ab9f7e0824a8ce40f394b8871de0e8465d725/protocol/kafka_sarama/v2/receiver.go#L57)
This method is extracting the message offset, adding 1 to it and finally (not synchroneously) inform kafka that this consumer-group has processed all messages before message.offset+1

Since there is no control over the processing time (the time spent in invoke), the goroutine processing the last kafka message (for example) could finish first. Then the offset of this message is pushed "first", saying that this consumer group is up to date with the current kafka log for this partition. If the process is stopped / crashed / etc, before the other goroutine finished, when the process will be restarted (or when the rebalance is done for other consumer in the same group). Only new messages will be sent. All the other messages that were computed by the different goroutines are lost

For example, I pushed 100 messages in kafka in a single partition topic
After that, I start a consumer with this configuration
```golang
	saramaConfig := sarama.NewConfig()
	saramaConfig.Version = sarama.V2_0_0_0
	saramaConfig.Consumer.Offsets.AutoCommit.Enable = true
	saramaConfig.Consumer.Offsets.Initial = sarama.OffsetOldest
```

for each message, I apply a pseudo processing with a random duration
```
func receive(ctx context.Context, event cloudevents.Event) {
	// Random processing time
	n := rand.Intn(1000)
	time.Sleep(time.Duration(n) * time.Second)

	fmt.Printf("After %vs, %s\n", n, event.DataEncoded)
}
```
After a few second I stop the process using Ctrl-C
![image](https://user-images.githubusercontent.com/40832495/102256606-e12dd700-3f0b-11eb-8811-90aefaf98938.png)

Looking at the kafka offset I see
[...]

PR #645, opened 2020-12-16
Fix Kafka ConsumeClaim loop to Mark right offset

pastequo:
https://github.com/cloudevents/sdk-go/issues/643

message is a pointer varying in the for loop, when MarkMessage is called, the pointer points to something potentially different to the current iteration

Signed-off-by: Matthieu Bernardin <matt.bernardin@gmail.com>


slinkydeveloper:
Wanna double check @n3wscott?

slinkydeveloper (approved):

n3wscott (commented):
There is no go call, so there is no fork and the closure fix does not seem to be required. I do not see how this code is any different than what was before. Add a test?

slinkydeveloper:
> There is no go call,

But the message is pushed inside a channel and used later, i think that's the issue here

pastequo:
I think it is related to the anonymous function:
https://play.golang.org/p/JD0mqu98uus

n3wscott:
LGTM

Issue #630, opened 2020-11-26
NATS flaky tests


slinkydeveloper:
```
 === RUN   TestSendStructuredMessagedToStructures/regular_subscriber/Event{"data":"hello","datacontenttype":"text/json","dataschema":"http://example.com/schema","exbinary":"AAECAw==","exbool":true,"exint":42,"exstring":"exstring","extime":"2020-03-21T12:34:56.78Z","exurl":"http://example.com/source","id":"full-event-0","source":"http://example.com/source","specversion":"1.0","subject":"topic","time":"2020-03-21T12:34:56.78Z","type":"com.example.FullEvent"}
coverage: [no statements]
panic: test timed out after 5m0s

goroutine 15 [running]:
testing.(*M).startAlarm.func1()
	/opt/hostedtoolcache/go/1.14.12/x64/src/testing/testing.go:1514 +0x11c
created by time.goFunc
	/opt/hostedtoolcache/go/1.14.12/x64/src/time/sleep.go:168 +0x52

goroutine 1 [chan receive]:
testing.(*T).Run(0xc0001b7560, 0xbdc7fb, 0x26, 0xbec968, 0x0)
	/opt/hostedtoolcache/go/1.14.12/x64/src/testing/testing.go:1096 +0x570
testing.runTests.func1(0xc0001b7560)
	/opt/hostedtoolcache/go/1.14.12/x64/src/testing/testing.go:1339 +0xa7
testing.tRunner(0xc0001b7560, 0xc0002bfce8)
	/opt/hostedtoolcache/go/1.14.12/x64/src/testing/testing.go:1050 +0x1ec
testing.runTests(0xc00000f6c0, 0x1087850, 0x1, 0x1, 0x0)
	/opt/hostedtoolcache/go/1.14.12/x64/src/testing/testing.go:1337 +0x595
testing.(*M).Run(0xc000156300, 0x0)
	/opt/hostedtoolcache/go/1.14.12/x64/src/testing/testing.go:1252 +0x300
main.main()
	_testmain.go:88 +0x348

goroutine 6 [select]:
go.opencensus.io/stats/view.(*worker).start(0xc00013c280)
	/home/runner/go/pkg/mod/go.opencensus.io@v0.22.3/stats/view/worker.go:154 +0x1d6
created by go.opencensus.io/stats/view.init.0
	/home/runner/go/pkg/mod/go.opencensus.io@v0.22.3/stats/view/worker.go:32 +0x9a

[...]

PR #653, opened 2021-01-07
Updating client to use NewClientHTTP, fixing new opencensus package

n3wscott:


v2/client/client.go
```diff
@@ -123,6 +121,11 @@ func (c *ceClient) Send(ctx context.Context, e event.Event) protocol.Result {
 	if err = e.Validate(); err != nil {
 		return err
 	}
+
+	// Event has been defaulted and validated, record we are going to preform send.
+	ctx, cb := c.observabilityService.RecordSendingEvent(ctx, e)
```
slinkydeveloper, line 126: If you move it down here, you don't record failed validation. Is that what you want?

v2/client/client.go
```diff
@@ -152,6 +152,10 @@ func (c *ceClient) Request(ctx context.Context, e event.Event) (*event.Event, pr
 		return nil, err
 	}
 
+	// Event has been defaulted and validated, record we are going to perform request.
+	ctx, cb := c.observabilityService.RecordRequestEvent(ctx, e)
```
slinkydeveloper, line 156: same as above

v2/client/client_http.go
```diff
@@ -22,3 +22,8 @@ func NewDefault(opts ...http.Option) (Client, error) {
 
 	return c, nil
 }
+
+// NewDefault has been replaced by NewHTTP
+// Deprecated. To get the same as NewDefault provided, please use the following:
+// TODO
```
slinkydeveloper, line 28: leftover?

slinkydeveloper (commented):
To be honest, I don't like the naming `observability/opencensus/v2/client`, I prefer `observability/opencensus/v2` so the user can alias-import with `ceopencensus observability/opencensus/v2`. But it's just a matter of style

v2/client/client.go
n3wscott: I moved this because if we do not do the defaulting first, the recorder does not see the true event on the wire and you do not get to record the `id`

v2/client/client.go
n3wscott: I moved this because if we do not do the defaulting first, the recorder does not see the true event on the wire and you do not get to record the `id`

v2/client/client_http.go
n3wscott: yeah :D

v2/client/client.go
slinkydeveloper: Makes sense, maybe in future we could have another signal for the validation failure to provide to the obs service, but that's fine now

slinkydeveloper:
LGTM

slinkydeveloper (approved):

n3wscott:
hm,.,, I need to look into why the integration tests are failing still.

slinkydeveloper:
@n3wscott it's natss #630

Issue #657, opened 2021-01-25
Set *stan.Msg metadata on receiver ctx


dan-j:
After discussion on slack, it it desirable to set NATS Streaming message metadata on the `context.Context` which can be passed into receivers.

This can be implemented by implementing the `binding.MessageContext` interface at the protocol layer.

At the moment, the context passed to receivers is the same as passed to `client.StartReceiver(ctx)`, so this functionality should prevent breaking changes for existing users where they expect `ctx` to have a certain state.

PR #658, opened 2021-01-25
Allow stan protocol to set native *stan.Msg as a context value

dan-j:
See #657 

This is an approach which doesn't introduce any breaking changes, if users do nothing their `.Receive()` functions will receive the same `ctx` object they always have, however using the new `PropogateContext()` option will override it.

I'm going to add some comments to other parts of the codebase in this MR because I think I have an alternative approach, but needs discussion

Signed-off-by: dan-j <5727701+dan-j@users.noreply.github.com>

### DIFF
--- a/protocol/stan/v2/context.go
+++ b/protocol/stan/v2/context.go
@@ -1,0 +1,44 @@
+package stan
+
+import (
+	"context"
+	"github.com/cloudevents/sdk-go/v2/binding"
+)
+
+// MsgMetadata holds metadata of a received *stan.Msg. This information is kept in a separate struct so that users
+// cannot interact with the underlying message outside of the SDK.
+type MsgMetadata struct {
+	Sequence        uint64
+	Redelivered     bool
+	RedeliveryCount uint32
+}
+
+type msgKeyType struct{}
+
+var msgKey msgKeyType
+
+// MetadataContextDecorator returns an inbound context decorator which adds STAN message metadata to
+// the current context. If the inbound message is not a *stan.Message then this decorator is a no-op.
+func MetadataContextDecorator() func(context.Context, binding.Message) context.Context {
+	return func(ctx context.Context, m binding.Message) context.Context {
+		if msg, ok := m.(*Message); ok {
+			return context.WithValue(ctx, msgKey, MsgMetadata{
+				Sequence:        msg.Msg.Sequence,
+				Redelivered:     msg.Msg.Redelivered,
+				RedeliveryCount: msg.Msg.RedeliveryCount,
+			})
+		}
+
+		return ctx
+	}
+}
+
+// MessageMetadataFrom extracts the STAN message metadata from the provided ctx. The bool return parameter is true if
+// the metadata was set on the context, or false otherwise.
+func MessageMetadataFrom(ctx context.Context) (MsgMetadata, bool) {
+	if v, ok := ctx.Value(msgKey).(MsgMetadata); ok {
+		return v, true
+	}
+
+	return MsgMetadata{}, false
+}
--- a/protocol/stan/v2/context_test.go
+++ b/protocol/stan/v2/context_test.go
@@ -1,0 +1,89 @@
+package stan
+
+import (
+	"context"
+	"github.com/cloudevents/sdk-go/v2/binding"
+	"github.com/nats-io/stan.go"
+	"github.com/nats-io/stan.go/pb"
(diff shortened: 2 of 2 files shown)


v2/client/invoker.go
```diff
@@ -118,12 +118,18 @@ func (r *receiveInvoker) IsResponder() bool {
...
-		return mctx.Context()
-	} else {
+		ctx = mctx.Context()
+	}
+
+	if ctx == nil {
 		return fallback
 	}
+
+	return ctx
```
dan-j, line 134: changes to this function means that if `MessageContext.Context()` returns `nil`, then `fallback` is returned, whereas previously it would return `nil`. 

Not sure if we can do this as it may be a breaking change, however I'd say this is safer since using a `nil` context is typically a bad idea anyway.

dan-j:
An alternative approach, but not sure of the consequences:

Is there a reason why `context.Background()` is passed in here? If we used `ctx` instead, then we can get rid of the `PropogateContext()` option.

https://github.com/cloudevents/sdk-go/blob/93f9fa1cd3103d71d1b609a75ec3160ed601b436/v2/binding/to_event.go#L45-L52

slinkydeveloper:
> Is there a reason why context.Background() is passed in here? If we used ctx instead, then we can get rid of the PropogateContext() option.

AFAIR I pass `context.Background()` here to avoid propagating any option related to the writes (ForceBinary, ForceStructured) which will prevent this DirectWrite to work

v2/client/invoker.go
slinkydeveloper: I'm ok with this change, thus as you outlined before, using another context is a breaking change

slinkydeveloper (commented):
I'm ok with this change, although before merging it, I would love to find out if we can avoid swapping context, but more "decorating it"

slinkydeveloper:
@dan-j Look at https://github.com/cloudevents/sdk-go/pull/659, the idea of this PR is to reuse the same context, jut wrap in it the additional informations we want to propagate from the message up to the client

slinkydeveloper:
I think we should not propagate the whole stan messages, but only the metadata we're interested in. How does it sounds?

n3wscott (approved):

dan-j:
> I think we should not propagate the whole stan messages, but only the metadata we're interested in. How does it sounds?

Sure, the three things which come to mind are: `Redelivered`, `RedeliveryCount` and `Sequence`.

Do you think these should be standard context options and the getters/setters somewhere in the core SDK, or keep them contained within the `stan` package? I know protocols such as http won't be able to support this functionality but others may, and since they're values on context consumers can't expect they be set, but can check if they are and act appropriately.

slinkydeveloper:
> keep them contained within the stan package?

Let's keep them in the stan package

dan-j:
@slinkydeveloper 

OK, I've updated this PR to expose a new context decorator which can be used with your new `WithInboundContextDecorator()` option. This still sets the whole message on the context, but uses an unexported key so the data is only available by using the accessor functions in `context.go`.

Thoughts?

protocol/stan/v2/context.go
```diff
@@ -0,0 +1,49 @@
...
+			// we store the raw *stan.Msg on the context, but under a unexported key so only the below accessor
+			// functions can be used to extract this info.
+			return context.WithValue(ctx, msgKey, msg.Msg)
+		}
+
+		return ctx
+	}
+}
+
+func SequenceFrom(ctx context.Context) (uint64, bool) {
```
slinkydeveloper, line 27: Can you add a godoc to this and below methods?

slinkydeveloper (commented):
That's good! Can you add the godocs to the methods?

protocol/stan/v2/context.go
dan-j: done, I've actually changed the behaviour to keep all the data in a single struct, seems unnecessary to have multiple functions to extract this data because you would have ended up with multiple `if value, ok := FooFrom(ctx); ok { ... }` statements if you wanted to view all of the properties.

slinkydeveloper (commented):
One last thing, can you add a small fancy test for it? Just add it to the actual test an "assert" that MsgMetadata is not empty

dan-j:
Added tests with a valid `*stan.Message` and one where it is not a `*stan.Message`

slinkydeveloper (approved):

PR #659, opened 2021-01-27
Propagate informations from inbound messages up to Client

slinkydeveloper:
Now MessageContext is not replaced anymore (like done in #634) but it can be used together with a context decorator function.

### DIFF
--- a/observability/opencensus/v2/http.go
+++ b/observability/opencensus/v2/http.go
@@ -1,14 +1,11 @@
 package v2
 
 import (
-	"context"
 	"net/http"
 
 	"go.opencensus.io/plugin/ochttp"
 	"go.opencensus.io/plugin/ochttp/propagation/tracecontext"
-	"go.opencensus.io/trace"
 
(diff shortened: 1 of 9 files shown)


slinkydeveloper:
cc @n3wscott

dan-j:
For this to be useful towards #658, we'd also need a way to add custom decorators to `ceClient.inboundContextDecorators`.

Perhaps a new option?

```go
// it may also be useful to type the decorator signature to make it easier to read
type ContextDecorator func(context.Context) context.Context

func WithInboundContextDecorators(decorators ...ContextDecorator) Option {
  return func(i interface{}) error {
    if c, ok := i.(*ceClient); ok {
      c.inboundContextDecorators = append(c.inboundContextDecorators, decorators...)
    }
    return nil
  }
}
```

slinkydeveloper:
@dan-j Added a new option

dan-j:
LGTM

n3wscott:
LGTM

Issue #656, opened 2021-01-19
sdk-go for amp protocol is not able to settle 1 out of 10 messages in my POC consitently 


aneeshkp:
When I used the example to send 10 messages , the client is able to receive only 9 out of 10 messages.
The sender  hangs waiting for  acknowledge for ever. In my opnion it should at least fail or time out, so the sender can resend the message.
 The POC I created  to test this scenario  can be found here, You can recreate this problem either by running in the cluster or stand alone
https://github.com/aneeshkp/cloudevents-amqp
I see following in the QDR logs
```
 ROUTER_CORE (info) [C29][L63] Stuck delivery: At least one delivery on this link has been undelivered/unsettled for more than 10 seconds
 ```
 Sending 10 messages , you can see following message in the log ( its same behaviour if I send 100 messages too, 1 will be not settled)
 ```
 2021/01/19 14:50:26 All 10 message was sent
2021/01/19 14:50:26 Total 1 messages were unsettled
2021/01/19 14:50:26 Unsettled messages
2021/01/19 14:50:26 --------------------
2021/01/19 14:50:26 Message id `2` was not settled and waitng
2021/01/19 14:50:26 --------------------
2021/01/19 14:50:26 Out of 10 messages ,Only 9 was settled

````
with 100 messages 
```
2021/01/19 15:15:59 100 MESSAGE SUCCESSFULLY DELIVERED <nil>
2021/01/19 15:15:59 --------- Summary ----------
2021/01/19 15:15:59 All 100 message was sent
2021/01/19 15:15:59 Total 1 messages were unsettled
2021/01/19 15:15:59 Unsettled messages
2021/01/19 15:15:59 --------------------
2021/01/19 15:15:59 Message id `2` was not settled and waiting
2021/01/19 15:15:59 --------------------
2021/01/19 15:15:59 Out of 100 messages ,Only 99 was settled
```

aneeshkp:
I just found out, if  I blast the qdr with +1 messages with zero sleep time then almost half of them are not settled.
The new code in the above link is able to test this scenario.

slinkydeveloper:
Hi @aneeshkp I tried to investigate a bit and it sounds like a problem of the amqp library, maybe some wrong configuration? The sdk-go code correctly sends the ack back for every message (i checked with your repro), but after the 5th message the receiver remains stuck, because nothing is received from the amqp library receiver.

aneeshkp:
Hi @slinkydeveloper 
[...]

PR #664, opened 2021-03-01
This will resolve issue mentioned in 656

aneeshkp:
This will resolve issue mentioned  in [656](https://github.com/cloudevents/sdk-go/issues/656)
 Current  implements creates both receiver and sender amqp transport pointing to same address and causes issues as mentioned here.
The link source address of the receiver and the link target address of the transmitter are set to the same value. This means that transmitted msgs can loop back to the sender. The address uses balanced deliver so that the first msg goes to the proper destination, but the second msg loops back to sender.

This PR resolves that issue by creating separate implementation of NewProtocol as NewSenderProtocol and NewReceiver Protocol, while maintaining old implementation of NewProtocol.

### DIFF
--- a/protocol/amqp/v2/protocol.go
+++ b/protocol/amqp/v2/protocol.go
@@ -86,6 +86,102 @@ func NewProtocol(server, queue string, connOption []amqp.ConnOption, sessionOpti
 	return p, nil
 }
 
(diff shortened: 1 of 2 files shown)


slinkydeveloper (approved):

slinkydeveloper:
Thanks for this PR!

PR #665, opened 2021-03-11
consumers of the SDK should be given power to throttle incoming messages

sks:
- Invoker does not have to execute in goroutine.
- This way the users of the library can add in any kind of throttle that they want to put in place

### DIFF
--- a/v2/client/client.go
+++ b/v2/client/client.go
@@ -246,13 +246,9 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 				}
 
 				// Do not block on the invoker.
-				wg.Add(1)
-				go func() {
-					if err := c.invoker.Invoke(ctx, msg, respFn); err != nil {
-						cecontext.LoggerFrom(ctx).Warn("Error while handling a message: ", err)
-					}
-					wg.Done()
-				}()
+				if err := c.invoker.Invoke(ctx, msg, respFn); err != nil {
+					cecontext.LoggerFrom(ctx).Warn("Error while handling a message: ", err)
+				}
 			}
 		}()
 	}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,5 +31,5 @@ Code style for this module is maintained using [`go fmt`](https://golang.org/cmd
 
 ### Maintainers
 
-If you are a maintainer of this repository, [here](maintainers_guide.pr) is a brief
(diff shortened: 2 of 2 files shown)


v2/client/client.go
```diff
@@ -246,13 +246,9 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 				}
 
 				// Do not block on the invoker.
-				wg.Add(1)
```
sks, line 249: This change was introduced in #574 to solve #569, But I think this change was not required to solve #569

n3wscott: this change makes the invoker single threaded, and is required to solve #569

n3wscott:
This PR does not add throttle control to the integrator, it adds it to the invoked connection downstream. Downstream connections can hold open and the throughput of the SDK will be limited by threads I think.

sks:
@n3wscott : Thanks you are right, This might not be the way to solve my problem.

here is what I did instead, 

```
func (s *Subscriber) start(ctx context.Context) error {
	consumer, _ := kafka_sarama.NewConsumer("localhost:9092", sarama.NewConfig(), "some_group_id", "some_topic")
	defer consumer.Close(context.Background())
	g, ctx := errgroup.WithContext(ctx)

	g.Go(func() error {
		return s.recieveMessageAndProcess(ctx, consumer)
	})
	g.Go(func() error {
		return consumer.OpenInbound(ctx)
	})
	return g.Wait()
}

func (s *Subscriber) recieveMessageAndProcess(ctx context.Context, consumer *kafka_sarama.Consumer) error {
	wg := &sync.WaitGroup{}
	throttle := make(chan struct{}, runtime.NumCPU())
	defer close(throttle)
	defer wg.Wait()
	select {
	case <-ctx.Done():
		return ctx.Err()
	default:
		for {
			msg, err := consumer.Receive(ctx)
			if err != nil {
				return fmt.Errorf("error receiving message: %w", err)
			}
			throttle <- struct{}{}
			wg.Add(1)
			go func(m binding.Message) {
				defer wg.Done()
				err = myLongRunningProcess(m)
				<-throttle
				if err == nil {
					_ = m.Finish(protocol.NewReceipt(true, "processed at %d", time.Now().Unix()))
					return
				}
				_ = m.Finish(protocol.NewReceipt(false, "error processing the message: %w", err))
			}(msg)
		}
	}
}

```

Looks like this solution is working at the moment, so will close this PR.

Thanks again 😅

PR #674, opened 2021-04-12
Remove costly parsing of content media type on the hot path

slinkydeveloper:
This PR avoids to invoke `mime.ParseMediaType` which is quite a costly method, just to grab the media type. Per https://tools.ietf.org/html/rfc7231#section-3.1.1.1, just checking for the `;` is enough. This is the same approach we use to parse the content type header in the http binding https://github.com/cloudevents/sdk-go/blob/946ebd9a535413a9cb0c2a7f910d76194a00e21a/v2/binding/format/format.go#L48

### DIFF
--- a/v2/event/eventcontext_v03_test.go
+++ b/v2/event/eventcontext_v03_test.go
@@ -162,10 +162,6 @@ func TestGetMediaTypeV03(t *testing.T) {
 		"nil": {
 			want: "",
 		},
-		"just encoding": {
-			t:    "charset=utf-8",
-			want: "",
-		},
 		"text/html with encoding": {
 			t:    "text/html; charset=utf-8",
 			want: "text/html",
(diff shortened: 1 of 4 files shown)


v2/event/eventcontext_v03_test.go
```diff
@@ -162,10 +162,6 @@ func TestGetMediaTypeV03(t *testing.T) {
 		"nil": {
 			want: "",
 		},
-		"just encoding": {
```
n3wscott, line 165: why delete the test?

slinkydeveloper: Because per spec https://tools.ietf.org/html/rfc7231#section-3.1.1.1, it seems to me that a content type with just encoding doesn't make sense

n3wscott (approved):

PR #675, opened 2021-04-12
Fix regression with json marshaller/unmarshaller

slinkydeveloper:
Fix the unmarshalling when there is some json in the event data, but there is no content type.

### DIFF
--- a/v2/event/event_unmarshal.go
+++ b/v2/event/event_unmarshal.go
@@ -359,7 +359,8 @@ func consumeDataAsBytes(e *Event, isBase64 bool, b []byte) error {
 	}
 
 	mt, _ := e.Context.GetDataMediaType()
-	if mt != ApplicationJSON && mt != TextJSON {
+	// Empty content type assumes json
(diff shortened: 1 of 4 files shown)


v2/event/event_unmarshal.go
```diff
@@ -359,7 +359,8 @@ func consumeDataAsBytes(e *Event, isBase64 bool, b []byte) error {
 	}
 
 	mt, _ := e.Context.GetDataMediaType()
-	if mt != ApplicationJSON && mt != TextJSON {
+	// Empty content type assumes json
+	if mt != "" && mt != ApplicationJSON && mt != TextJSON {
```
n3wscott, line 363: This will throw new errors when the body is just an unquoted string or random bytes.

slinkydeveloper: Why? Can you provide a test case where do you think this is going to fail?

n3wscott (approved):

matzew:
@n3wscott @slinkydeveloper will this fix result in a 2.4.1 release ?

Issue #670, opened 2021-04-08
Extension name with more than 20 characters throws error


deissnerk:
[`SetExtension()`](https://github.com/cloudevents/sdk-go/blob/f322543770b0da85f3e001dbda6f9e686cbe745f/v2/event/eventcontext_v1.go#L75) of the v1.0 event context throws an error, if the name of the extension is longer than 20 characters. In the spec this length restriction is a [SHOULD NOT](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#attribute-naming-convention) restriction. So failing with an error seems to be too strong. Shouldn't it be possible to violate SHOULD requirements? 😉

n3wscott:
It should be possible. Good point. What version of the sdk are you seeing this in? it is in the latest 2.4.0?

deissnerk:
We experienced it first with the 2.4.0 release. I think it might have sneaked in with this [commit](https://github.com/cloudevents/sdk-go/commit/2c1245bfced37e33cc33c00d1abadb8d11f1a63f).

n3wscott:
Yep. ok thanks! I will get a point fix release out in the next couple days.

deissnerk:
Thank you. If it is only about removing the length check, I can come up with a little PR, too.

n3wscott:
@deissnerk If you want to write the PR, that would be great, I was thinking of having a global setting that an importer can set the max length to something they wish, but have the default be unlimited (0).

PR #678, opened 2021-04-13
By default, do not restrict the length of extension attribute names

deissnerk:
I introduced a global variable `MaxExtensionNameLength` in package event, which by default is 0. If you prefer other mechanisms to globally set this value, please let me know.

As I wanted to focus on #670, I kept the changes minimal. I think that some of the functions would belong into different files now. E.g. `SetExtension()` could now go into `eventcontext_v1_writer.go`, and `extension.go` is not really needed as a separate file. 

Signed-off-by: Klaus Deissner <klaus.deissner@sap.com>

### DIFF
--- a/v2/event/eventcontext_v1.go
+++ b/v2/event/eventcontext_v1.go
@@ -1,7 +1,6 @@
 package event
 
 import (
-	"errors"
 	"fmt"
 	"mime"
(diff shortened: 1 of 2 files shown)


n3wscott (approved):
LGTM

PR #681, opened 2021-04-22
Adding SPDX Authors headers using Apache 2.0.

n3wscott:

### DIFF
--- a/hack/boilerplate/add-boilerplate.sh
+++ b/hack/boilerplate/add-boilerplate.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The CloudEvents Authors
+# SPDX-License-Identifier: Apache-2.0
+
+USAGE=$(cat <<EOF
+Add boilerplate.<ext>.txt to all .<ext> files missing it in a directory.
+
+Usage: (from repository root)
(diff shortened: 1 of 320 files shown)


AUTHORS
```diff
@@ -0,0 +1,9 @@
+# This is the list of CloudEvents authors for copyright purposes.
```
slinkydeveloper, line 1: Do we must have this file? I would prefer if we don't, because it needs to be kept updated

slinkydeveloper (commented):
Just a comment on the authors fine, otherwise lgtm

AUTHORS
n3wscott: I think it is used to define what "CloudEvents Authors" mean. Whenever I see a group name in the copyright, I always see an authors file.

n3wscott: I could be wrong... I am not a lawyer...

slinkydeveloper: if that's the case, this doesn't represent all the authors, because there are people that have contributed to this project outside these 4 companies. I think the git commits suffice to define "cloudevents authors", but yeah i'm not a lawyer too :smile:

n3wscott: I am gonna drop it...

n3wscott:
cc @grantr

PR #686, opened 2021-05-12
Adding error to protocol result if not 200 status code

danielgrist:
Adding the error message to the protocol result so it can be retrieved when there are non 200 status codes.

Signed-off-by: Daniel Grist <dgrist@ibm.com>

### DIFF
--- a/v2/protocol/http/protocol_retry.go
+++ b/v2/protocol/http/protocol_retry.go
@@ -8,6 +8,7 @@ package http
 import (
 	"context"
 	"errors"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"time"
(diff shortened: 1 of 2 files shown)


duglin:
@n3wscott could you take a look at this one?

v2/protocol/http/protocol_retry.go
```diff
@@ -42,7 +43,12 @@ func (p *Protocol) doOnce(req *http.Request) (binding.Message, protocol.Result)
 	if resp.StatusCode/100 == 2 {
 		result = protocol.ResultACK
 	} else {
-		result = protocol.ResultNACK
+		respbody, err := ioutil.ReadAll(resp.Body)
```
n3wscott, line 46: Reading the body here will cause line 54 to error I think, because the body will have already been read. 

The body of the response is already returned within the message, this change should not be required.

danielgrist: The change was made because the message is dropped later on here: https://github.com/cloudevents/sdk-go/blob/main/v2/protocol/http/protocol.go#L142

n3wscott: How about we read and wrap the returned message from Request in Send and return it if err is not nil on L142? So fix the problem one bump up, rather than in protocol_retry. Seems like protocol_retry is doing the right thing, we just need to not drop the body if there is an error in Send.

n3wscott (approved):

Issue #661, opened 2021-02-03
Protobuf codec and format support


kconwayinvision:
Hey folks, I have an implementation of https://github.com/cloudevents/spec/blob/v1.0.1/protobuf-format.md that I'm looking to contribute. Before I open a PR, I figured I'd give folks a chance to see what is involved. I've put some code up for demonstration here: https://github.com/kconwayinvision/cloudevents-go-protobuf. It's a fairly straightforward implementation of `format.Format`, `datacodec.Encoder`, and `datacodec.Decoder`. The real complexity is in implementing the requirement to use the `ProtoData` field when the format and data codec are both protobuf.

If this belongs in the SDK then I can open a PR for us to talk about more specific issues such as whether or not to automate protobuf code generation or the complexity of the protobuf type juggling. I figured I'd start by asking whether or not this even belongs here or if this should go in a contrib space somewhere.

slinkydeveloper:
Hi @kconwayinvision, thanks for the proposal. We would really love to see the protobuf format, what about starting with a PR for the implementation of `format.Format`?

For the datacodec implementation, let's figure it out after the format

kconwayinvision:
A pull request for the Format is up here: https://github.com/cloudevents/sdk-go/pull/662

kconwayinvision:
If there are folks watching this issue and not the PRs then https://github.com/cloudevents/sdk-go/pull/688 is waiting on reviews.

kconwayinvision:
This is ready to close.

Format support added: https://github.com/cloudevents/sdk-go/pull/662
Codec support added: https://github.com/cloudevents/sdk-go/pull/688

PR #688, opened 2021-05-19
Add protobuf datacodec

kconwayinvision:
This adds the ability to send and receive protobuf encoded data within
the envelope.

### DIFF
--- a/binding/format/protobuf/v2/protobuf.go
+++ b/binding/format/protobuf/v2/protobuf.go
@@ -6,6 +6,7 @@ import (
 	stdtime "time"
 
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/cloudevents/sdk-go/v2/binding/format"
@@ -30,9 +31,6 @@ var (
 
 const (
 	ApplicationCloudEventsProtobuf = "application/cloudevents+protobuf"
-	// applicationProtobuf embedded here but unexported temporarily until it
-	// is moved to the datacodec implementation.
-	applicationProtobuf = "application/protobuf"
 )
 
 // StringOfApplicationCloudEventsProtobuf  returns a string pointer to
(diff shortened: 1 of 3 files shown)


binding/format/protobuf/v2/protobuf.go
```diff
@@ -106,19 +104,19 @@ func sdkToProto(e *event.Event) (*pb.CloudEvent, error) {
...
-	// }
+	if e.DataContentType() == ContentTypeProtobuf {
+		anymsg := &anypb.Any{}
+		if err := proto.Unmarshal(e.Data(), anymsg); err != nil {
+			if e.DataSchema() == "" {
+				return nil, fmt.Errorf("cannot encode direct protobuf message without dataschema. set dataschema to the appropriate protobuf type like type.googleapis.com/packge.v1.Type or make sure you are using the appropriate data content type %s", ContentTypeProtobuf)
+			}
+			anymsg.TypeUrl = e.DataSchema()
+			anymsg.Value = e.Data()
+		}
```
kconwayinvision, line 115: I want to highlight this section because it demonstrates the big question I have for this PR. There's a circular relationship between the codec and the format defined in the [protobuf spec](https://github.com/cloudevents/spec/blob/v1.0.1/protobuf-format.md) that is difficult to resolve without either changing the spec or introducing non-specified implementations details.

When both protobuf format and encoding are used together then we must use the `ProtoData` field of the protobuf CloudEvent type to store the data. I believe the intention of the protobuf specification is to match the behavior of JSON+JSON which embeds the data as a native type within the envelope rather than a JSON ecoded string within a JSON envelope. In the case of protobuf, though, the only generic type available is `Any` which contains two fields: a type URI and a protobuf encoded message.

In order to create an `Any` we must know the source type URI. The `Encode` implementation in this draft takes a source protobuf message and generates an `Any` from it such that the type URI is correct. However, this means the `Encode` and `Decode` methods practically only work correctly when used with the protobuf format. If used with the JSON format then `Encode` will incorrectly produce an encoded `Any` rather than an encoded form of the original message. Only users of this Go SDK would handle this case.

The problem I've highlighted here, though, is that there is not enough information within the `Event` during formatting to construct the `Any` required to comply with the spec. The special behavior of the format should be entirely encapsulated here but the type erasure that happens during the datacodec `Encode` leaves us without a type URI. The same type URI could be set as the dataschema value but this value is not required by either the primary CloudEvent spec or the protobuf extension.

Some options I've considered that I'd like feedback on:

1. Only convert from source data to `Any` in the format when protobuf content type is used. Use the dataschema value, including no value, as the type URI in the `Any`. Rely on the use of the non-type URI validating decoding feature of `Any` to prevent malformed or missing type URIs from blocking decoding back to a valid protobuf message.  This option technically results in an [invalid `Any`](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto#L125) but this would be hidden behind the format abstraction and never surfaced to users. The risk is that other languages protobuf implementations may not offer the option to bypass invalid or missing type URIs.

2. Change the protobuf specification to require the `dataschema` value and match it to the type URI that should be included in an `Any`. However, the type URI of a protobuf message is most often not a valid URI as required by the CloudEvent specification but a URI-reference because it is missing the required URI scheme. This means we would need to modify the actual type URI to something CloudEvent compliant in order to use that field. Additionally, the API for getting a message type URI is not a top-level feature but an artifact of the protobuf reflection interface in Go (ex: `string(msg.ProtoReflect().Descriptor().FullName())`). I'm not certain how other language implementation of protobuf expose this.

3. Artificially restrict the protobuf datacodec to only be used with the protobuf format. This would hide the issues without addressing them. However, this option also seems to require that we leak through the format/datacodec abstraction and add a special case somewhere in the main SDK that identifies and bails on this condition. Otherwise, it's not clear how we could impose that restriction only through the scope of this plugin.


I think Option 1 is the better of the batch because it best hides the implementation detail of using an `Any` and allows the various CloudEvent SDKs to manage the complexity of handling a technically invalid `Any` due to the potentially missing or malformed type URI. That puts the least burden on users of the SDK and works within the constraints of the current spec. That being said, I'm open to hearing more options or what folks think of the options I've listed.

kconwayinvision:
Hey folks, checking in. Any preference on how to go or would folks rather I choose?

kconwayinvision:
I went ahead an implemented Option 1 which was to move the `Any` type wrapping into the format. This ensures that the datacodec implementation is generally useful and can be used in conjunction with non-protobuf formats for the envelope.

Whenever folks get a chance to review, I believe this is ready to go. When it's merged it will close https://github.com/cloudevents/sdk-go/issues/661.

kconwayinvision:
Checking in again. I'm still interested in getting this line of work closed out. Let me know if there are any comments or questions.

ezbercih:
does this PR also address leveraging a schema registry like the one that comes in Kafka ecosystem in order to validate the the protobuf data payload?

kconwayinvision:
@ezbercih No. It's limited to only implementing the schema agnostic data encoding.

There's nothing here that prevents some kind of integration with a schema registry in the future. This PR is limited to completing https://github.com/cloudevents/spec/blob/master/protobuf-format.md.

n3wscott:
APPROVE thanks!

PR #689, opened 2021-06-03
Finish() no longer used binding-message in Send()

xinydev:
Finish() no longer used binding-message in Send().  and I found that some trace mechanisms must rely on Finish() to complete.

Signed-off-by: XinYang <xinydev@gmail.com>

### DIFF
--- a/v2/protocol/http/protocol.go
+++ b/v2/protocol/http/protocol.go
@@ -141,6 +141,9 @@ func (p *Protocol) Send(ctx context.Context, m binding.Message, transformers ...
 	}
 
 	msg, err := p.Request(ctx, m, transformers...)
+	if msg != nil {
+		defer func() { _ = msg.Finish(err) }()
+	}
 	if err != nil && !protocol.IsACK(err) {
 		var res *Result
 		if protocol.ResultAs(err, &res) {


slinkydeveloper:
I think this PR is wrong, finish is invoked by the `cloudevents.Client`. If you're using the protocol manually, you should invoke finish by yourself

xinydev:
> I think this PR is wrong, finish is invoked by the `cloudevents.Client`. If you're using the protocol manually, you should invoke finish by yourself

yes, when the user invokes [Request()](https://github.com/cloudevents/sdk-go/blob/main/v2/client/client.go#L139) function in `cloudevents.Client`, the Finished() is invoked  by `cloudevents.Client`. 

but when the user invokes [Send()](https://github.com/cloudevents/sdk-go/blob/main/v2/client/client.go#L111) function in `cloudevents.Client`,  the `cloudevents.Client` can not get the original `binding.Message`, the `binding.Message` only exists in protocol's [Send()](https://github.com/cloudevents/sdk-go/blob/main/v2/protocol/http/protocol.go#L136) function.

slinkydeveloper:
> but when the user invokes Send() function in cloudevents.Client, the cloudevents.Client can not get the original binding.Message, the binding.Message only exists in protocol's Send() function.

Maybe that's what needs to be fixed? I mean, on the cloudevents.Client side?

xinydev:
We can indeed fix this on the `cloudevents.Client ` side, by changing [this](https://github.com/cloudevents/sdk-go/blob/main/v2/client/client.go#L135) to invoke protocol's Request() function . 

But the problem still exists, users who use the protocol manually, they invoke the [protocol's Send() function](https://github.com/cloudevents/sdk-go/blob/main/v2/protocol/http/protocol.go#L136), and then, they can not invoke the Finish() because the Send() function does not return the Message

xinydev:
> > but when the user invokes Send() function in cloudevents.Client, the cloudevents.Client can not get the original binding.Message, the binding.Message only exists in protocol's Send() function.
> 
> Maybe that's what needs to be fixed? I mean, on the cloudevents.Client side?

We can indeed fix this on the `cloudevents.Client ` side, by changing [this](https://github.com/cloudevents/sdk-go/blob/main/v2/client/client.go#L135) to invoke protocol's Request() function . 

But the problem still exists, users who use the protocol manually, they invoke the [protocol's Send() function](https://github.com/cloudevents/sdk-go/blob/main/v2/protocol/http/protocol.go#L136), and then, they can not invoke the Finish() because the Send() function does not return the Message

xinydev:
@slinkydeveloper Would you please take another look?
The Message I finished is the response of the HTTP request, not the body of the request. I think the user who uses Send() function does not care about this response Message(if they care response, they should call Request() instead), they only care about the request's success or failure.

https://github.com/cloudevents/sdk-go/blob/43b8eca508219f50e528e86de1aa3df0f97546fa/v2/protocol/http/protocol.go#L143-L156

slinkydeveloper:
@n3wscott any ideas about this one?

xinydev:
@n3wscott  ping

dan-j:
Looking at a few of the other protocols, they all set up a deferred call to `Finish()`, but the http protocol doesn't. Maybe this PR is correct?

dan-j:
Ah, ignore my last comment, I misread the PR and just realised you explained you'd finished the response, not the original message (which is already finished)

I can't really comment on this either, I've never used the http protocol before so I don't know if this is going to cause any issues with other integrations.

n3wscott (approved):

Issue #699, opened 2021-08-04
Memory leak in daemon


bluntik:
Hi,
we are using client send in daemon and we have memory leak. During debug I found next
Example of code
```
package main

import (
	"context"
	cloudevents "github.com/cloudevents/sdk-go/v2"
	"log"
	"net/http"
	_ "net/http/pprof"
	"sync"
)

func main() {
	wg := sync.WaitGroup{}

	wg.Add(1)
	go func() {
		defer wg.Done()
		log.Println(http.ListenAndServe("localhost:6060", nil))
	}()

	c, err := cloudevents.NewClientHTTP()
	if err != nil {
		log.Fatalf("failed to create client, %v", err)
	}

	queue := make(chan struct{}, 100)

	for i := 0; i < 500; i++ {
		queue <- struct{}{}

		go func() {
			// Create an Event.
			event :=  cloudevents.NewEvent()
			event.SetSource("example/uri")
			event.SetType("example.type")
			event.SetData(cloudevents.ApplicationJSON, map[string]string{"hello": "world"})

			// Set a target.
			ctx := cloudevents.ContextWithTarget(context.Background(), "http://localhost:14046/health-check")

			// Send that Event.
			if result := c.Send(ctx, event); cloudevents.IsUndelivered(result) {
				log.Fatalf("failed to send, %v", result)
			}
			<-queue
		}()
	}

	wg.Wait()
}
```

If I sending 500 request in example in debug goruitine we will see 1000 not close gorutine 500 for read 500 for write
http://localhost:6060/debug/pprof/goroutine?debug=1 

```
goroutine profile: total 1005
500 @ 0x103b8a5 0x104c7d7 0x132aae7 0x1071c61
#	0x132aae6	net/http.(*persistConn).readLoop+0x966	/usr/local/Cellar/go/1.16.4/libexec/src/net/http/transport.go:2203

500 @ 0x103b8a5 0x104c7d7 0x132bef7 0x1071c61
#	0x132bef6	net/http.(*persistConn).writeLoop+0xf6	/usr/local/Cellar/go/1.16.4/libexec/src/net/http/transport.go:2382

1 @ 0x103b8a5 0x103421b 0x106c155 0x10c8f25 0x10c9b55 0x10c9b37 0x12289cf 0x1234571 0x130a418 0x1071c61
[...]

PR #700, opened 2021-08-04
Fix memory leak(#699)

bluntik:
Signed-off-by: Alexey Ryazanov <a.ryazanov@corp.mail.ru>

* Update v2/protocol/http/protocol_retry.go

### DIFF
--- a/v2/protocol/http/protocol_retry.go
+++ b/v2/protocol/http/protocol_retry.go
@@ -34,6 +34,11 @@ func (p *Protocol) do(ctx context.Context, req *http.Request) (binding.Message, 
 
 func (p *Protocol) doOnce(req *http.Request) (binding.Message, protocol.Result) {
 	resp, err := p.Client.Do(req)
+
+	if resp != nil {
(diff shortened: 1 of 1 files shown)


v2/protocol/http/protocol_retry.go
```diff
@@ -34,6 +34,11 @@ func (p *Protocol) do(ctx context.Context, req *http.Request) (binding.Message,
 
 func (p *Protocol) doOnce(req *http.Request) (binding.Message, protocol.Result) {
 	resp, err := p.Client.Do(req)
+
+	if resp != nil {
+		defer resp.Body.Close()
```
n3wscott, line 39: This happens when the returned `binding.Message` has `.Finish` called here: https://github.com/cloudevents/sdk-go/blob/main/v2/protocol/http/protocol.go#L143-L146

This change will break the contract with the the `binding.Message`; So there is something else going on. Still looking.

n3wscott:
Already fixed at HEAD

Issue #697, opened 2021-07-15
Deprecate nats and stan?


n3wscott:
@zhaojizhuang should we deprecate the nats and stan packages in favor of jetstream?

zhaojizhuang:
@n3wscott maybe we should deprecate `stan`, But `nats` should be kept. cc @ripienaar @wallyqs

ripienaar:
Yes agree let’s deprecate Stan but nats can be kept

n3wscott:
ok thanks! I will make that PR and close this issue.

bruth:
I believe this needs to be closed.

PR #701, opened 2021-08-11
Deprecate the stan protocol

n3wscott:
fixes https://github.com/cloudevents/sdk-go/issues/697

`stan` has been replaced with `nats_jetstream`.


dan-j:
So I don't have a problem with deprecating support for STAN, however I don't believe telling users in the documentation that they should use the `nats_jetstream` package instead is the correct wording.

Jetstream is a new component of core NATS which implements the same functionality as NATS Streaming. If a user still had a NATS Streaming cluster, they can't use the Jetstream library. 

I've also reviewed this on my phone, so if I've misinterpreted anything let me know.

n3wscott:
thanks @dan-j 

HOLD

PR #703, opened 2021-08-13
update the dep on nats-server to >= 2.1.9

n3wscott:
GHSA-2c64-vj8g-vwrq
high severity
Vulnerable versions: < 2.1.9
Patched version: 2.1.9
(This advisory is canonically https://advisories.nats.io/CVE/CVE-2020-26892.txt )

NATS nats-server through 2020-10-07 has Incorrect Access Control because of how expired credentials are handled.

The NATS accounts system has expiration timestamps on credentials; the https://github.com/nats-io/jwt library had an API which encouraged misuse and an IsRevoked() method which misused its own API.

A new IsClaimRevoked() method has correct handling and the nats-server has been updated to use this. The old IsRevoked() method now always returns true and other client code will have to be updated to avoid calling it.

The CVE identifier should cover any application using the old JWT API, where the nats-server is one of those applications.


n3wscott:
APPROVE, this is a dep update, no sdk code change.

Issue #684, opened 2021-05-03
calling close on amqp  protocol object , generates infinite loop.


aneeshkp:
When calling close on protocol
``` 
func (t *Protocol) Close(ctx context.Context) (err error) {
````
The code goes  on an infinite loop 
``` 
client/client.go line 244
if err != nil {
					cecontext.LoggerFrom(ctx).Warn("Error while receiving a message: ", err)
					continue
}
```
With error 
```
go:244","msg":"Error while receiving a message: amqp: link closed"}
{"level":"warn","ts":1620052511.8445728,"logger":"fallback","caller":"client/client.go:244","msg":"Error while receiving a message: amqp: link closed"}
```

PR #707, opened 2021-08-16
Refactor client to return error on receiver.Receive

### LINKED ISSUE #685: How to reconnect amqp if session ended by server
Hi!
When start receiving, check if something wrong only when starting.

// this is blocking
if err := client.StartReceiver(ctx, eventHandler.Handle); err != nil {
		log.Fatal().Err(err).Msg("EventListener failed to start")
		return err
	}

But, if server down when this function already running, I got this warning:
{"level":"warn","ts":1620560979.9585598,"logger":"fallback","caller":"client/client.go:244","msg":"Error while receiving a message: session ended by server: *Error{Condition: amqp:internal-error, Description: shutdown, Info: map[]}"}

And I can't do anything. My service is still running but got an infinity loop, and all that I get - this message. How I can stop receiving and start a new session?

oleg-kulyk:
Signed-off-by: Oleg Kulyk <oleg.kulyk@checkmarx.com>

Fixes #685 and #684 

Current behavior locks down consumer in an infinite loop if connection to Rabbitmq is closed. When analyzing the client code I saw no real purpose in "hiding" error coming from Receiver and just continuing running.

Current change should not introduce any b/c breaking issues, unless the client code depends on consumer not "dying" if connection is lost (for which I struggle to come up with a meaningful example).

The fix for test came after inspecting locally what would be the outcome if err is retuned (in which case test panicked because of "fail" in running goroutine after the `t.Run` process is finished).

### DIFF
--- a/v2/client/client.go
+++ b/v2/client/client.go
@@ -224,11 +224,10 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 	}()
 
 	// Start Polling.
+	errChan := make(chan error)
 	wg := sync.WaitGroup{}
 	for i := 0; i < c.pollGoroutines; i++ {
-		wg.Add(1)
 		go func() {
-			defer wg.Done()
 			for {
 				var msg binding.Message
 				var respFn protocol.ResponseFn
@@ -241,13 +240,9 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 					respFn = noRespFn
 				}
 
-				if err == io.EOF { // Normal close
-					return
-				}
-
 				if err != nil {
-					cecontext.LoggerFrom(ctx).Warn("Error while receiving a message: ", err)
-					continue
+					errChan <- err
+					return
 				}
 
 				// Do not block on the invoker.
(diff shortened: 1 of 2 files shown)


v2/client/client.go
```diff
@@ -241,13 +240,9 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
...
 				}
 
-				if err == io.EOF { // Normal close
-					return
-				}
-
 				if err != nil {
-					cecontext.LoggerFrom(ctx).Warn("Error while receiving a message: ", err)
-					continue
+					errChan <- err
```
n3wscott, line 244: not all errors cause all protocols to not respond, some would like the message to be returned via Invoke, this change does not seem like it will work for all protocols.

Is there a change to the amqp impl that would allow it to handle the error? Like, perhaps it needs to respond with a `io.EOF` when the the server is disconnected.

oleg-kulyk: Thanks for the review.

I think it does make send to send `io.EOF` in case connection is down, I'll look in that direction.

oleg-kulyk:
closing in favor of #750

PR #708, opened 2021-08-17
Use message context as inbound context when available

benmoss:
This way middlewares are able to mutate the message context and have that get threaded through to our invoker. Right now the context that the invoker will receive is the one passed into `StartReceiver`, which isn't as helpful.


n3wscott:
the failure is something with nats and not related to this PR. 

LGTM

PR #715, opened 2021-08-30
Propagate error to observability service

### LINKED ISSUE #713: Observability service getting empty errors in the callback function
While implementing the `ObservabilityService` interface to work with OpenTelemetry, I noticed that when errors happen during the several operations in cloudevents (`client.go` `Send` method), the callback in the observability service `func(errOrResult error)` is invoked with an empty error. Depending on the use case, developers would want to maybe mark the span created by the ObservabilityService as failed, but without the error result it's not possible to figure it out.

I started debugging to see why, and I found out that in several places the callback is called in a `defer` but *before* the error is actually collected. For example, [here](https://github.com/cloudevents/sdk-go/blob/70febdc855835091a1d5f111d9223f226eb0ab0a/v2/client/client.go#L133)

joaopgrassi:
Call defer in the appropriate location, so the observability service callback is invoked with the error object populated. This allows to record the error into the span, or set the span as failed. 

Fixes #713

### DIFF
--- a/observability/opencensus/v2/client/client_test.go
+++ b/observability/opencensus/v2/client/client_test.go
@@ -15,23 +15,25 @@ import (
 	"testing"
 	"time"
 
-	"github.com/lightstep/tracecontext.go/traceparent"
-	"github.com/stretchr/testify/require"
-	"go.opencensus.io/trace"
-
 	obshttp "github.com/cloudevents/sdk-go/observability/opencensus/v2/http"
+	"github.com/cloudevents/sdk-go/v2/binding"
 	"github.com/cloudevents/sdk-go/v2/client"
 	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/cloudevents/sdk-go/v2/observability"
 	"github.com/cloudevents/sdk-go/v2/protocol"
 	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
 	"github.com/cloudevents/sdk-go/v2/types"
+	"github.com/lightstep/tracecontext.go/traceparent"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opencensus.io/trace"
 )
 
-func simpleTracingBinaryClient(t *testing.T, target string) client.Client {
+func simpleTracingBinaryClient(t *testing.T, target string, os client.ObservabilityService) client.Client {
 	p, err := obshttp.NewObservedHTTP(cehttp.WithTarget(target))
 	require.NoError(t, err)
 
-	c, err := client.New(p, client.WithObservabilityService(New()))
+	c, err := client.New(p, client.WithObservabilityService(os))
 	require.NoError(t, err)
 	return c
 }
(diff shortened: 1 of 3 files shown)


v2/client/client.go
```diff
@@ -186,7 +184,7 @@ func (c *ceClient) Request(ctx context.Context, e event.Event) (*event.Event, pr
 	} else {
 		resp = rs
 	}
-
+	defer cb(err, resp)
```
n3wscott, line 187: this change should not do anything as err and resp are defined here: https://github.com/cloudevents/sdk-go/pull/715/files#diff-bfa445e7f19a76da55d2f4a1f51a4dd9cb7e2bf54efd28fd96428078c1453732R139-R140 and are never recreated in the function

v2/client/invoker.go
```diff
@@ -81,9 +81,9 @@ func (r *receiveInvoker) Invoke(ctx context.Context, m binding.Message, respFn p
 
 			var cb func(error)
 			ctx, cb = r.observabilityService.RecordCallingInvoker(ctx, e)
-			defer cb(result)
 
 			resp, result = r.fn.invoke(ctx, e)
+			defer cb(result)
```
n3wscott, line 86: same here, can you confirm the new test fails when the defer order is changed to the original?

n3wscott:
You are correct. https://play.golang.org/p/PyVqhb3LVKM

Interesting! I expected the interface pointers to not be reassigned, but clearly there is a go-ism I don't understand on that one. 

Could you please resort the imports, and then LGTM

Thanks for the PR and the issue.

observability/opencensus/v2/client/client_test.go
```diff
@@ -15,23 +15,25 @@ import (
...
-
 	obshttp "github.com/cloudevents/sdk-go/observability/opencensus/v2/http"
+	"github.com/cloudevents/sdk-go/v2/binding"
 	"github.com/cloudevents/sdk-go/v2/client"
 	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/cloudevents/sdk-go/v2/observability"
 	"github.com/cloudevents/sdk-go/v2/protocol"
 	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
 	"github.com/cloudevents/sdk-go/v2/types"
+	"github.com/lightstep/tracecontext.go/traceparent"
```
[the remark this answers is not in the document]
joaopgrassi: Ah I'll fix these. VSCode just auto-sorted them.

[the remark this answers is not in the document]
joaopgrassi: Done.

joaopgrassi:
@n3wscott not sure why the integration tests failed, they also failed once in our fork and re-triggering it worked. I guess it's some network issue? Since it seems it was waiting for 12min..

v2/client/client.go
n3wscott: I was wrong. It seems to matter for `error` type. which is annoying...

n3wscott:
LGTM

Thanks a ton for the fix @joaopgrassi !!

Issue #693, opened 2021-06-23
How to set OpenTelemetry tracing parameters?


boramalper:
I am using [Knative Eventing](https://knative.dev/docs/eventing/) and want to set [OpenTelemetry](https://opentelemetry.io/) tracing parameters (tracing ID, span ID etc.) for the event that I am sending to [Knative Eventing Broker](https://knative.dev/docs/eventing/broker/).

I thought `WithTracePropagation()` could have been helpful but it is [deprecated](https://pkg.go.dev/github.com/cloudevents/sdk-go/v2@v2.4.1/client#WithTracePropagation) and no alternatives are listed. [`ObservabilityService`](https://pkg.go.dev/github.com/cloudevents/sdk-go/v2@v2.4.1/client#ObservabilityService) could have been an alternative solution but it does not have an `OutboundContextDecorators()`... Though I would like to be able to set OpenTelemetry tracing parameters myself freely, without using middlewares or other magic.

What I currently have is this (simplified):
```go
import(
	cloudevents "github.com/cloudevents/sdk-go/v2"
	"github.com/google/uuid"
)


type MyStruct struct {
	// ...
}

func main() {	
	event := cloudevents.NewEvent("1.0")
	event.SetID(uuid.New().String())
	event.SetType("my-type")
	event.SetSource("//my-source")

	_ = event.SetData(cloudevents.ApplicationJSON, MyStruct{})

	ceClient, _ := cloudevents.NewClientHTTP()
	// I want to set OpenTelemetry trace ID here!
	if result := ceClient.Send(cloudevents.ContextWithTarget(ctx, env.Sink), event); !cloudevents.IsACK(result) {
		log.Fatalf("failed to send CloudEvent: %+v", result)
	}
}
```

boramalper:
@ustiugov

ustiugov:
hi @joaopgrassi, this is definitely something we would be interested in. Would be great if you could upstream that. 

thanks @boramalper for bringing me into this discussion.

joaopgrassi:
Hi @ustiugov cool, I'll working on it. I feel though there's some scattered issues about this. I found this old one https://github.com/cloudevents/sdk-go/issues/554 and commented there about the sdk spec about this observability service.

n3wscott:
Yes please, glad you got it working. @benmoss was also helping clear some of this up with some examples added to the sdk. Maybe Ben has more to say?

benmoss:
[...]

PR #717, opened 2021-09-20
Introduce OpenTelemetry observability service

### LINKED ISSUE #554: Migrate to OpenTelemetry
We should look at what it will take to migrate to OpenTelemetry.

One option could be to break out all telemetry integrations and produce two modules, one for what we have with OpenSenses and one with the new OpenTelemetry (it is not 1.0 yet).

Ref:

https://github.com/open-telemetry/opentelemetry-go/

https://opentelemetry.io/

joaopgrassi:
Resolves #554, #693

## This PR

- Adds a new implementation of the `ObservabilityService` interface for [OpenTelemetry](https://opentelemetry.io). The new observability service is similar to the existing one for OpenCensus. 
- Exposes inject and extract functions to read and write [W3C Trace Context data](https://www.w3.org/TR/trace-context/) to/from events respectively.
- Exposes helper methods to create (and extend) CloudEvents HTTP clients enabled with OpenTelemetry
- A sample project demonstrating how to use the "observed" client with OpenTelemetry (with spans that can be viewed with Jaeger)
- Tests (in a separated module to not impose the OTel default SDK as a transitive dependency for users)

Note: The observability service is only producing spans at the moment. Metrics support can be added once the SDK is more stable.


n3wscott:
AWESOME!! I am reviewing now, give me a day or two, thanks so much!

joaopgrassi:
@n3wscott there might be some changes related to the attributes that are added to the spans created by the ObservabilityService. It's currently being discussed in my PR on the OTel spec. In case you want to see it: https://github.com/open-telemetry/opentelemetry-specification/pull/1978

Not sure which is going to be reviewed/merged first, but if this is merged first, then I can change it later to adhere with the OTel spec :).

n3wscott (approved):
amazing work, with tests, samples and READMEs and everything!! I can't wait to use this instead of the open census integration.

Issue #709, opened 2021-08-18
SetExtension seems to override core attributes for http at least.


n3wscott:
Test and fix. SetExtensions should not override core attributes.

PR #720, opened 2021-10-04
fix: SetExtension cannot override core attributes

embano1:
Fail if `SetExtension` is called with a core attribute.

TODO:

- [ ] Implement for `V03`
- [ ] Update func docs

Closes: #709


embano1:
Up for discussion:

Where can I reference the `V03` and `V1` spec attributes instead of creating a custom map? I tried using `Attributes()` from `spec` using but that introduces a circular dep.

v2/event/extensions.go
```diff
@@ -48,6 +48,24 @@ func validateExtensionName(key string) error {
...
+		"time":            {},
+		"source":          {},
+		"specversion":     {},
+		"type":            {},
+		"datacontenttype": {},
+		"dataschema":      {},
+	}
+
+	// TODO: not safe, can be bypassed with a ce-ce- prefix
+	trimmed := strings.TrimPrefix(key, "ce-")
```
n3wscott, line 63: not needed, if the key has a `-` it is not a valid extension per the test on 70.

n3wscott:
> Up for discussion:
> 
> Where can I reference the `V03` and `V1` spec attributes instead of creating a custom map? I tried using `Attributes()` from `spec` using but that introduces a circular dep.

Issue #59, opened 2019-03-07
[Webhook] Implement Allowed Rate


n3wscott:

slinkydeveloper:
Is this solved by https://github.com/cloudevents/sdk-go/pull/491 ?

n3wscott:
Allowed rate is now negotiated but rate limit is not implemented.

embano1:
Shall I take a step and create a PR with something like https://github.com/sethvargo/go-limiter (see httplimit) or rather basic using x/rate? Background: I currently need this functionality (also #56) so happy to support :)

PR #722, opened 2021-10-11
Add inbound rate limiting

embano1:
WIP, first proposal up for discussion to implement inbound rate limiting. 

Features/Behavior:

- Applied in `ServeHTTP`, i.e. cannot be bypassed
- `Interface,` i.e. bring your own via HTTP `Option`
- Defaults to no rate limiting via `noOpLimiter` to keep current behavior, even though we have been advertising a `magic` number `1000` when asked for in HTTP options

See #59 for more context.

Closes: #59


benmoss:
cc @n3wscott

PR #724, opened 2021-10-13
Pub/Sub message ordering support

### LINKED ISSUE #723: Pub/Sub ordering key support
See https://cloud.google.com/pubsub/docs/ordering#go

skymeyer:
Fixes #723 

Message ordering may require producing the events using an explicit [regional endpoint](https://cloud.google.com/pubsub/docs/reference/service_apis_overview#service_endpoints). Not sure if we should add a helper to produce a pubsub client using a regional service endpoint or leave this as an exercise to the integrator.


rriolobos:
I tested pubsub ordering key with this library but I fond an issue. When you send 3 or more messages, with same ordering key, receiver only waits for first one, second and third messages are consumed concurrently. I added a 5 second sleep in message handler in order to check in the total amount of time was 15seconds as expected, but it was 10 second. It can also be checked with logs. When I try the same directly invoking pubsub receiver, ir worked as expected.

skymeyer:
Question from @rriolobos is tracked under #753.

PR #741, opened 2021-11-24
736 feat pubsub subscription filter

irejanodev:
My apologies, I forgot to include the filter value in the `Connection` struct to the creation method. 
This fix allows server filtering as the pubsub implementation.

CC/ @n3wscott


n3wscott (approved):

n3wscott:
It looks like you need to sign the commit to pass DCO, but LGTM other than that

irejanodev:
> It looks like you need to sign the commit to pass DCO, but LGTM other than that

Yes! thanks again!

Issue #742, opened 2021-11-25
CE HTTP: keep request data at the context


odacremolbap:
The HTTP CE Client for receiving requests is designed to be used like this:

```golang
...
client := NewClient(options)
err := client.StartReceiver(ctx, dispatch)
...
func dispatch(ctx context.Context, event ce.Event) (*ce.Event, ce.Result) {
   return process(event)
}
```

I would like to have more information on the received request at the CloudEvent client to be able to create multi-tenant clients. That would be:

```golang
func dispatch(ctx context.Context, event ce.Event) (*ce.Event, ce.Result) {
   req  := ce.HTTPRequestFromContext(ctx)
   process := processLookupEventProcessor(req)
   return process(event)
}
```

For this to happen the request or a subset of their data should be added to the context before the switch here:

https://github.com/cloudevents/sdk-go/blob/1170e89edb9b504a806f2c6a26563c3c26b68276/v2/protocol/http/protocol.go#L285-L301

---

The scenario I am working on are generic transformations of CloudEvent payloads where users can configure their own transformation parameters getting a different URL path for each configuration. Requesting the same CloudEvents client at each of those paths would call the dispatcher, and inside it the developer could use the raw request (request.URL.Path in my case) to load the configuration:

- my.ce.transformation/configuration-for-jane
- my.ce.transformation/configuration-for-joan

---

**Alternatives**

- Using the an extended attribute or even the payload to inform the expected configured tenancy.

I have no control on the received events from users, but as a Sinks developer I can expose a distinct URL Address for each configuration. This alternative is not possible for me.

- Not using CE Client.

While this is a solution, I would love to use the retries and abuse protection of the CE Client without duplicating that code at my end.

n3wscott:
Originally the SDK did something like this but it was not very reliable for the other protocols. But I still think it might be ok to add additional values in the context of the message (here https://github.com/cloudevents/sdk-go/blob/1170e89edb9b504a806f2c6a26563c3c26b68276/v2/protocol/http/message.go#L75-L76) and you can probe for them in the invoked receiver (which is what I think you are wanting to do now?)

Would this work for you?

odacremolbap:
Yes, that would be perfect. If I am undertanding that right that would be

```go
message := NewMessage(req.Header, req.Body) 
message.ctx = WithRequest(req.Context(), req)
```

so that I can do later something like

```go
func dispatch(ctx context.Context, event ce.Event) (*ce.Event, ce.Result) {
[...]

PR #747, opened 2021-12-09
HTTP Protocol: add RequestData in Context option.

odacremolbap:
Closes https://github.com/cloudevents/sdk-go/issues/742

@n3wscott this is not implemented exactly as described in the issue.

> The way I would do this is I would make a context helper in the cloudevents http package, make a new struct with these fields, copy them over before leaving the serve http method into the context using the helper context method.

Instead of adding it at the [ServeHTTP](https://github.com/cloudevents/sdk-go/blob/77f73c28a5a032a26ce0c344addc3ef4b14d60ac/v2/protocol/http/protocol.go#L285) method this PR leverages the http.Options, so this feature is an opt-in.

My main concern is that requests will hit the middleware prior to the `RateLimiter` which feels like the wrong order, but that's the case for all middlewares here. I don't consider that an issue in this PR and if we were to improve the pattern I would suggest moving the `RateLimiting` feature also as an option (in a different PR).

### DIFF
--- a/v2/protocol/http/options.go
+++ b/v2/protocol/http/options.go
@@ -287,3 +287,15 @@ func WithRateLimiter(rl RateLimiter) Option {
 		return nil
 	}
 }
+
+// WithRequestDataAtContextMiddleware adds to the Context RequestData.
+// This enables a user's dispatch handler to inspect HTTP request information by
+// retrieving it from the Context.
+func WithRequestDataAtContextMiddleware() Option {
+	return WithMiddleware(func(next nethttp.Handler) nethttp.Handler {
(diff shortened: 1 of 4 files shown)


v2/protocol/http/options.go
```diff
@@ -287,3 +287,15 @@ func WithRateLimiter(rl RateLimiter) Option {
 		return nil
 	}
 }
+
+// WithRequestDataAtContextMiddleware adds to the Context RequestData.
+// This enables a user's dispatch handler to inspect HTTP request information by
+// retrieving it from the Context.
+func WithRequestDataAtContextMiddleware() Option {
```
n3wscott, line 294: nice idea!!

n3wscott (approved):
Awesome job!

PR #758, opened 2022-02-15
Make SdkToProto and ProtoToSDK public

pingsutw:
I want to use custom API to send cloudevent to my topic, but I can't convert cloudevent objects to cloudevent proto due to private API.

hope we can make these APIs public.

### DIFF
--- a/binding/format/protobuf/v2/protobuf.go
+++ b/binding/format/protobuf/v2/protobuf.go
@@ -56,7 +56,7 @@ func (protobufFmt) MediaType() string {
 }
 
 func (protobufFmt) Marshal(e *event.Event) ([]byte, error) {
-	pbe, err := sdkToProto(e)
+	pbe, err := ToProto(e)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


pingsutw:
cc @n3wscott

binding/format/protobuf/v2/protobuf.go
```diff
@@ -76,7 +76,7 @@ func (protobufFmt) Unmarshal(b []byte, e *event.Event) error {
 }
 
 // convert an SDK event to a protobuf variant of the event that can be marshaled.
-func sdkToProto(e *event.Event) (*pb.CloudEvent, error) {
```
n3wscott, line 79: The method is used but not renamed, and I am not sure this is a great name for an exported method. 

I would think `ToProto` and `FromProto` would be better names to export.

n3wscott (approved):

pingsutw:
@n3wscott Could we create a release for this changed? thanks!

PR #764, opened 2022-03-23
fix http result errors by wrapping the upstream error, allowing for isNack

n3wscott:


n3wscott:
@sbgreene1307 would you mind reviewing this? Thanks for your unit test and your issue!!

v2/protocol/http/protocol.go
```diff
@@ -157,7 +157,12 @@ func (p *Protocol) Send(ctx context.Context, m binding.Message, transformers ...
 				buf := new(bytes.Buffer)
 				buf.ReadFrom(message.BodyReader)
 				errorStr := buf.String()
-				err = NewResult(res.StatusCode, "%s", errorStr)
+				// If the error is not wrapped, then just add the message body.
+				if _, ok := err.(*Result); ok {
+					err = NewResult(res.StatusCode, "%s", errorStr)
```
n3wscott, line 162: ah, I think this loses the inner error.

v2/protocol/http/protocol.go
```diff
@@ -157,7 +157,17 @@ func (p *Protocol) Send(ctx context.Context, m binding.Message, transformers ...
 				buf := new(bytes.Buffer)
 				buf.ReadFrom(message.BodyReader)
 				errorStr := buf.String()
-				err = NewResult(res.StatusCode, "%s", errorStr)
+				// If the error is not wrapped, then append the original error string.
+				if og, ok := err.(*Result); ok {
+					if len(og.Format) == 0 {
+						og.Format = "%s"
+					} else {
+						og.Format = og.Format + ": %s"
```
n3wscott, line 165: still not quite there. Still testing

test/integration/http/tap_handler.go
```diff
@@ -113,6 +114,9 @@ func (t *tapHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(500)
 		return
 	}
+	if t.statusCode > 299 {
+		w.WriteHeader(t.statusCode)
```
sbgreene1307, line 118: I forgot to add a return here which I think is what is causing the "superfluous" log line that you see in the tests.  Do you want to just add that?

test/integration/http/direct.go
```diff
@@ -83,6 +86,7 @@ func ClientDirect(t *testing.T, tc DirectTapTest, copts ...client.Option) {
 				t.Errorf("expected ACK, got %s", result)
 			}
 		} else if !cloudevents.ResultIs(result, tc.wantResult) {
+			t.Errorf("result.IsUndelivered = %v", cloudevents.IsUndelivered(result))
```
sbgreene1307, line 89: This can be removed now as well

sbgreene1307 (changes requested):
A couple minor things to button up

test/integration/http/tap_handler.go
n3wscott: Will do!

test/integration/http/direct.go
n3wscott: I don't mind the extra output

test/integration/http/tap_handler.go
n3wscott, line 118:
```suggestion
		w.WriteHeader(t.statusCode)
		return
```

Issue #770, opened 2022-05-07
CE receiver client does not have option to make callback a blocking call


smilence-yu:
https://github.com/cloudevents/sdk-go/pull/574 changed the behavior of the CE v2 client so the callback `fn` is always asynchronous ( one go routine is spawn for each event ). 

https://github.com/cloudevents/sdk-go/blob/main/v2/client/client.go#L251-L258

This indeed addressed https://github.com/cloudevents/sdk-go/issues/569 but also made any of the following use cases a lot trickier.

1. Don't want two events to be processed at the same time
2. Want events to be handled in the order they were received
3. Want to resume from last check point in the event of service restart ( e.g. in Kafka case, `fn` callback may commit the offset for the messages they has been received and restarted service will pick up any event hasn't been committed offset for)

I'd propose adding an option `WithBlockingInvokerCall` to allow the old behavior. I'll create a pull request for it to be considered.

### How to reproduce

*This is basically the counter of the same section in https://github.com/cloudevents/sdk-go/issues/569*

Run the same example receiver code as in https://github.com/cloudevents/sdk-go/issues/569: https://gist.github.com/liu-cong/a132fe8629715419c7a58ee20001d071

Then send a few events to the receiver using curl, 

```shell
curl -v localhost:8080
  -X POST \
  -H "Ce-Sleep: 1m" \
  -H "Ce-Id: 1" \
  -H "Ce-Specversion: 1.0" \
  -H "Ce-Type: type" \
  -H "Ce-Source: source" \
  -H "Content-Type: application/json" \
  -d '{"msg":"Test!"}
```

The `receive` function won't be blocked until the sleep period is done

PR #771, opened 2022-05-08
feat(client): Add option to make receive callback blocking

smilence-yu:
Addresses https://github.com/cloudevents/sdk-go/issues/770 

- Add an option for the callback on receiver to be blocking or nonblocking
- Updated `blocking_test` to ensure when `pollRoutines` is `1` AND it's blocking, the events are processed in serialized manner.

### Manual Testing

Run the example receiver code: https://gist.github.com/liu-cong/a132fe8629715419c7a58ee20001d071
It retrieves the Ce-Sleep attribute from the event and sleep for that period of time.

Then send a few events to the receiver using curl, 
```shell
curl -v localhost:8080
  -X POST \
  -H "Ce-Sleep: 20s" \
  -H "Ce-Id: 1" \
  -H "Ce-Specversion: 1.0" \
  -H "Ce-Type: type" \
  -H "Ce-Source: source" \
  -H "Content-Type: application/json" \
  -d '{"msg":"Test!"}
```

#### Without `blockingCallback` flag

`cloudevents.NewClient(p, client.WithPollGoroutines(1))`

Sent 3 events, finish at the same time

image

#### With `blockingCallback` flag

`cloudevents.NewClient(p, client.WithBlockingCallback(), client.WithPollGoroutines(1))`

image

Finished one after another

Signed-off-by: James Yu <jyu@confluent.io>

### DIFF
--- a/test/integration/http/blocking_test.go
+++ b/test/integration/http/blocking_test.go
@@ -45,6 +45,7 @@ type BlockingSenderReceiverTestOutput struct {
 type BlockingSenderReceiverTestCases map[string]BlockingSenderReceiverTest
 
 func TestNonBlockingSenderReceiver(t *testing.T) {
+	t.Parallel()
 	now := time.Now()
 
 	testCases := BlockingSenderReceiverTestCases{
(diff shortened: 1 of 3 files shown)


n3wscott:
The lint error seems to be related to go 1.18 and the linter. Not related to this PR.

Issue #773, opened 2022-06-03
`req.Body` not reset in `doWithRetry`


embano1:
# Scenario: 

When using a `net/http` HTTP server as remote/receiver (instead of the receiver from this package), `ceClient` has issues when retrying a non-successful HTTP response. 

Error: `"msg":"Invalid result type, not HTTP Result: Post \"http://always-fail.default.127.0.0.1.sslip.io\": http: ContentLength=34 with Body length 0 (type: *protocol.Receipt)"}`

This happens when the remote server returns a message in the response, e.g. „server down“ (not CE) but not when just returning a status code and empty response body.

The issue is caused by `func (p *Protocol) doWithRetry()` because it does not reset the `req.Body` during subsequent attempts.

## Expected behavior: 

Return encoding error, but correctly reset `req.Body`, e.g. return `"failed to convert response into event: unknown Message encoding\n500:  (3x)"`

## Client code:

```go
/*
 Copyright 2021 The CloudEvents Authors
 SPDX-License-Identifier: Apache-2.0
*/

package main

import (
	"context"
	"fmt"
	"net/http"
	"time"

	"github.com/cloudevents/sdk-go/observability/opencensus/v2/client"
	"go.uber.org/zap"
	"knative.dev/eventing/pkg/kncloudevents"
	"knative.dev/pkg/logging"

	cloudevents "github.com/cloudevents/sdk-go/v2"
	"github.com/cloudevents/sdk-go/v2/protocol"
	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
)

func main() {
	ctx := cloudevents.ContextWithTarget(context.Background(), "http://always-fail.default.127.0.0.1.sslip.io")

	topts := []cehttp.Option{cehttp.WithIsRetriableFunc(func(statusCode int) bool {
		retry, _ := kncloudevents.SelectiveRetry(ctx, &http.Response{StatusCode: statusCode}, nil)
		return retry
	})}

	ceClient, err := client.NewClientHTTP(topts, nil)
	if err != nil {
		panic(err)
	}

	e := cloudevents.NewEvent()
	e.SetType("com.cloudevents.sample.sent")
	e.SetSource("https://github.com/cloudevents/sdk-go/v2/samples/httpb/sender")
	_ = e.SetData(cloudevents.ApplicationJSON, map[string]interface{}{
		"id":      1,
		"message": "Hello, World!",
	})

[...]

PR #774, opened 2022-06-09
Reset request body on retry

embano1:
Fixes a bug where requests with a body (event) are not always retried due to the underlying `net/http` throwing `*url.Error` because of body and content-length mismatch (body is empty and not reset upon retry). See #773 for details.

The tests did not catch this as they were not doing behaving as a real HTTP server (just mocking `RoundTripper`) and also not carrying an event body (which would not trigger this behavior). 

- [x] Changed retry tests to use a network connection and HTTP (test) server
- [x] Reset body (if not `nil`) during retries
- [x] Set `GetBody` func only if it is not already set (the sdk already does this currently)

Closes: #773

### DIFF
--- a/v2/protocol/http/protocol_retry.go
+++ b/v2/protocol/http/protocol_retry.go
@@ -6,8 +6,11 @@
 package http
 
 import (
+	"bytes"
 	"context"
 	"errors"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"time"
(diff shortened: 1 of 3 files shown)


embano1:
Linter workflow seems broken.

n3wscott (approved):

n3wscott:
Thank you so much @embano1 !! Both for the detailed bug report and the follow-up fix. I will cherry-pick this into the current release as a bug fix and we will get this out. 🙏 🙏 🙏

PR #778, opened 2022-06-24
bump golang to 1.18

inigohu:
Bumping golang version to 1.18
Signed-off-by: Iñigo Horcajo <inigohu@gmail.com>


n3wscott:
If there is a good reason for this, let me know. This change forces all downstreams to update to go 1.18.

n3wscott:
I think we could bump to 1.17 to fix the test?

Issue #783, opened 2022-07-18
The sender can't get the error message from receiver


wenfengwang:
Hi `sdk-go` community, I started a receiver like:
```go
func receive(ctx context.Context, event cloudevents.Event) cloudevents.Result {
	fmt.Printf("Got an Event: %s", event)
	return cloudevents.NewHTTPResult(401, "test")
}
```
but the sender can't get the error message:
![image](https://user-images.githubusercontent.com/9766821/179494617-a976cdaa-c83b-4a81-b36b-6f2866a9b779.png)

and I found in https://github.com/cloudevents/sdk-go/blob/main/v2/protocol/http/protocol.go#L366, that only `StatusCode` be used, but `Format` and `Args` are ignored.
![image](https://user-images.githubusercontent.com/9766821/179495129-445e504c-c2b0-418a-9418-14d44c46ebe4.png)

I was wondering if it was a bug?

n3wscott:
Seems like a bug, want to fix?

wenfengwang:
> Seems like a bug, want to fix?

OK, I will submit a PR to fix this.

n3wscott:
Thanks for the PR!!

PR #785, opened 2022-07-20
fix: missing error message when return HTTPResult

wenfengwang:
fix issue #783

### DIFF
--- a/samples/http/receiver-result/main.go
+++ b/samples/http/receiver-result/main.go
@@ -1,5 +1,5 @@
 /*
- Copyright 2021 The CloudEvents Authors
+ Copyright 2022 The CloudEvents Authors
  SPDX-License-Identifier: Apache-2.0
 */
 
--- a/samples/http/receiver-result/main.go
+++ b/samples/http/receiver-result/main.go
@@ -8,9 +8,10 @@ package main
 import (
 	"context"
 	"fmt"
-	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"log"
 	"net/http"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
 )
 
 func main() {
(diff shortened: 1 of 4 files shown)


samples/http/receiver-result/main.go
```diff
@@ -0,0 +1,39 @@
+/*
+ Copyright 2021 The CloudEvents Authors
```
n3wscott, line 2:
```suggestion
 Copyright 2022 The CloudEvents Authors
```

v2/protocol/http/protocol.go
```diff
@@ -390,6 +391,11 @@ func (p *Protocol) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		}
 
 		rw.WriteHeader(status)
+		if errMsg != "" {
```
n3wscott, line 394: I believe go does not like to write any body after writing the status header. Can you flip these lines?

n3wscott (commented):
Thanks! couple of nits and it LGTM

samples/http/receiver-result/main.go
wenfengwang: DONE

v2/protocol/http/protocol.go
wenfengwang: Thanks for your advice @n3wscott . It seems that we have to write body after writing the header. In Go source code https://github.com/golang/go/blob/dev.boringcrypto.go1.18/src/net/http/server.go#L1600, which shows that it will automatically set header when calls `Write`, and in https://github.com/golang/go/blob/dev.boringcrypto.go1.18/src/net/http/server.go#L1146, which shows that after any ` WriteHeader` called, the flag will be set to `true`, and in https://github.com/golang/go/blob/dev.boringcrypto.go1.18/src/net/http/server.go#L1140, which checks the flag and if it was the `true`, the new `StatusCode` will be ignored. I was wondering if my understanding was right, hope your feedback. : )

wenfengwang:
@n3wscott Hi, please have a look on the PR : )

n3wscott (approved):

Issue #817, opened 2022-11-02
After a rebalance, the receiver keeps receiving messages until it errors out


nbajaj90:
After a rebalance, the receiver keeps receiving messages until its error out, and most of these messages fails to commit. Do we need to fix the ConsumeClaim to return once the consumer session ends (because of rebalance), so that the receiver won't keep receiving messages after rebalance.

To fix here:
https://github.com/cloudevents/sdk-go/blob/6784a56ad84051544d141e597a226e8e48cb8a98/protocol/kafka_sarama/v2/receiver.go#L55

References:
https://github.com/Shopify/sarama/issues/2118
https://github.com/Shopify/sarama/blob/5e2c2ef0e429f895c86152189f625bfdad7d3452/examples/consumergroup/main.go

nbajaj90:
Expecting something like - 
```
func (r *Receiver) ConsumeClaim(session sarama.ConsumerGroupSession, claim sarama.ConsumerGroupClaim) error {
	for {
		select {
		case msg, ok := <-claim.Messages():
			if !ok {
				return nil
			}
			m := NewMessageFromConsumerMessage(msg)

			r.incoming <- msgErr{
				msg: binding.WithFinish(m, func(err error) {
					if protocol.IsACK(err) {
						session.MarkMessage(msg, "")
					}
				}),
			}
		case <-session.Context().Done():
			return nil
		}
	}
}
```

PR #818, opened 2022-11-03
Patch sarama_kafka rebalance fix

nbajaj90:
Issue: https://github.com/cloudevents/sdk-go/issues/817
Issue Explanation and Fix: https://github.com/Shopify/sarama/issues/2118
Fix reference: https://github.com/Shopify/sarama/blob/5e2c2ef0e429f895c86152189f625bfdad7d3452/examples/consumergroup/main.go#L177

### DIFF
--- a/protocol/kafka_sarama/v2/receiver.go
+++ b/protocol/kafka_sarama/v2/receiver.go
@@ -76,7 +76,7 @@ func (r *Receiver) ConsumeClaim(session sarama.ConsumerGroupSession, claim saram
 		// Also checked Shopify/sarama code which calls this ConsumeClaim method, and don't see if there is any difference
 		// whether this method returns error or not. If it returns the error, as per current implementation, it could
 		// get printed in logs and later drained when the ConsumerGroup gets closed.
-		// For now, to be on safer side, returning nil instead of session.Context().Err() as sugeested in
+		// For now, to be on safer side, returning nil instead of session.Context().Err() as suggested in
 		// https://github.com/Shopify/sarama/blob/5e2c2ef0e429f895c86152189f625bfdad7d3452/examples/consumergroup/main.go
 		case <-session.Context().Done():
 			return nil
--- a/protocol/kafka_sarama/v2/receiver.go
+++ b/protocol/kafka_sarama/v2/receiver.go
@@ -52,13 +52,17 @@ func (r *Receiver) Close(context.Context) error {
 	return nil
 }
 
+// ConsumeClaim must start a consumer loop of ConsumerGroupClaim's Messages().
+// Also the method should return when `session.Context()` is done.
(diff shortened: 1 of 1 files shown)


duglin:
LGTM
ping @lionelvillard for final review

protocol/kafka_sarama/v2/receiver.go
```diff
@@ -53,19 +53,25 @@ func (r *Receiver) Close(context.Context) error {
...
+
+			r.incoming <- msgErr{
+				msg: binding.WithFinish(m, func(err error) {
+					if protocol.IsACK(err) {
+						session.MarkMessage(msg, "")
+					}
+				}),
+			}
+		case <-session.Context().Done():
+			return nil
```
embano1, line 72: nit: I generally recommend the context error here instead of surpressing it, e.g. `return session.Context().Err()`

protocol/kafka_sarama/v2/receiver.go
```diff
@@ -53,19 +53,25 @@ func (r *Receiver) Close(context.Context) error {
...
-			msg: binding.WithFinish(m, func(err error) {
-				if protocol.IsACK(err) {
-					session.MarkMessage(msg, "")
-				}
-			}),
+	for {
+		select {
+		case msg, ok := <-claim.Messages():
+			if !ok {
+				return nil
```
embano1, line 60: question: since the behavior is now different from the previous implementation (the previous one basically never returned IIUC), should we add a comment here to describe when `ConsumeClaim` returns and what `error` would be (if any)?

protocol/kafka_sarama/v2/receiver.go
lionelvillard: +1

lionelvillard:
LGTM

@nbajaj90 can you look at @embano1 comments?

protocol/kafka_sarama/v2/receiver.go
nbajaj90: Added a comment in code for this, please review.

protocol/kafka_sarama/v2/receiver.go
nbajaj90: Added a comment in code for this, please review.

protocol/kafka_sarama/v2/receiver.go
embano1: You're the expert on the implementation so I think it's ok. Probably better to put that comment in the Method comment section so it's exported and doc-ed with the method?

lionelvillard:
LGTM

waiting for @embano1 for final review.

protocol/kafka_sarama/v2/receiver.go
nbajaj90: No, I am not an expert, just wanted that fix patched here. Let me move the comment.

embano1 (approved):

embano1:
Approved

duglin:
LGTM
Merging

Issue #301, opened 2020-02-04
Batch encoding support in bindings interfaces


slinkydeveloper:

slinkydeveloper:
After playing with the [multipart experiment](https://github.com/slinkydeveloper/sdk-go/commits/multipart_experiment), i come out with [`MultiMessage`](https://github.com/slinkydeveloper/sdk-go/blob/multipart_experiment/v2/binding/multi_message.go#L3).

`MultiMessage` is a `Message` iterator with a `Finish()`. I didn't included this interface in `Message` because it requires dropping the assumption **everywhere in the sdk** that 1 message corresponds to 1 event. `MultiMessage` is easily implementable, an example: https://github.com/slinkydeveloper/sdk-go/blob/multipart_experiment/v2/binding/multi_event_message.go

My proposal is that we should provide at bindings level this interface and in http protocol implement it for batch encoding. Then, we won't have any support for it at protocol level (and at client level) and we'll provide to users an example for using it (like https://github.com/slinkydeveloper/sdk-go/blob/multipart_experiment/v2/cmd/samples/http_multipart/random_multi_event_send/main.go).

My problem with this plan is [`http.NewMessage`](https://github.com/slinkydeveloper/sdk-go/blob/multipart_experiment/v2/protocol/http/message.go#L50): What this function should do, if we open the doors to `MultiMessage`? 

* Should it return a void pointer that could be **or** a single `Message` **or** a `MultiMessage`?
* We want to have another function called  `NewMultiMessage` that returns always a `MultiMessage` and, in case of single structured/binary message, returns a one element `MultiMessage`?

n3wscott:
I think I agree with a new `MultiMessage` type, it can implement the message interface but to built it up is more involved, like maybe it consumes `Message`s? And then to use this we add a `SendMulti(ctx, []events)` or something.

So I can see this being implemented without an API breaking change on the surfaces we care about.

acodetailor:
support batch event now ?

Xopek:
Really looking forward to be able to send multiple events at once. Can i help somehow to implement it?

PR #829, opened 2022-12-22
http: Batch Events from HTTP Request and Response

markmandel:
Simple implementations of NewEventsFromHTTPRequest and NewEventsFromHTTPResponse as a step towards providing some support for `application/cloudevents-batch+json`.

This doesn't help with batch sending of Events, but this provides some basics of being able to process batch requests and/or responses from standard Go HTTP handling.

This could be improved over time (See previous work on #301) as well, but optimised for having something that works to start, rather than implementing big design changes.

Signed-off-by: Mark Mandel <markmandel@google.com>

### DIFF
--- a/v2/alias.go
+++ b/v2/alias.go
@@ -140,8 +140,8 @@ var (
 	NewEventFromHTTPResponse  = http.NewEventFromHTTPResponse
 	NewEventsFromHTTPRequest  = http.NewEventsFromHTTPRequest
 	NewEventsFromHTTPResponse = http.NewEventsFromHTTPResponse
-	NewHTTPRequestFromEvents  = http.NewHTTPRequestFromEvents
 	NewHTTPRequestFromEvent   = http.NewHTTPRequestFromEvent
+	NewHTTPRequestFromEvents  = http.NewHTTPRequestFromEvents
 	IsHTTPBatch               = http.IsHTTPBatch
 
 	// HTTP Messages
--- a/v2/alias.go
+++ b/v2/alias.go
@@ -135,8 +135,10 @@ var (
 	ToMessage = binding.ToMessage
 
 	// Event Creation
-	NewEventFromHTTPRequest  = http.NewEventFromHTTPRequest
-	NewEventFromHTTPResponse = http.NewEventFromHTTPResponse
+	NewEventFromHTTPRequest   = http.NewEventFromHTTPRequest
+	NewEventFromHTTPResponse  = http.NewEventFromHTTPResponse
+	NewEventsFromHTTPRequest  = http.NewEventsFromHTTPRequest
+	NewEventsFromHTTPResponse = http.NewEventsFromHTTPResponse
 
 	// HTTP Messages
 
--- a/v2/alias.go
+++ b/v2/alias.go
@@ -135,11 +135,14 @@ var (
 	ToMessage = binding.ToMessage
 
 	// Event Creation
+
 	NewEventFromHTTPRequest   = http.NewEventFromHTTPRequest
 	NewEventFromHTTPResponse  = http.NewEventFromHTTPResponse
 	NewEventsFromHTTPRequest  = http.NewEventsFromHTTPRequest
 	NewEventsFromHTTPResponse = http.NewEventsFromHTTPResponse
 	NewHTTPRequestFromEvents  = http.NewHTTPRequestFromEvents
+	NewHTTPRequestFromEvent   = http.NewHTTPRequestFromEvent
+	IsHTTPBatch               = http.IsHTTPBatch
 
 	// HTTP Messages
 
--- a/v2/alias.go
+++ b/v2/alias.go
@@ -139,6 +139,7 @@ var (
 	NewEventFromHTTPResponse  = http.NewEventFromHTTPResponse
 	NewEventsFromHTTPRequest  = http.NewEventsFromHTTPRequest
 	NewEventsFromHTTPResponse = http.NewEventsFromHTTPResponse
+	NewHTTPRequestFromEvents  = http.NewHTTPRequestFromEvents
 
 	// HTTP Messages
 
--- a/v2/binding/format/format.go
+++ b/v2/binding/format/format.go
@@ -51,6 +51,9 @@ func (jb jsonBatchFmt) MediaType() string {
 	return event.ApplicationCloudEventsBatchJSON
 }
 
+// Marshal will return an error for jsonBatchFmt since the Format interface doesn't support batch Marshalling, and we
+// know it's structured batch json, we'll go direct to the json.UnMarshall() (see `ToEvents()`) since that is the best
+// way to support batch operations for now.
 func (jb jsonBatchFmt) Marshal(e *event.Event) ([]byte, error) {
 	return nil, errors.New("not supported for batch events")
 }
--- a/v2/binding/format/format.go
+++ b/v2/binding/format/format.go
@@ -7,6 +7,7 @@ package format
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -41,12 +42,30 @@ func (jsonFmt) Unmarshal(b []byte, e *event.Event) error {
 	return json.Unmarshal(b, e)
 }
 
+// JSONBatch is the built-in "application/cloudevents-batch+json" format.
+var JSONBatch = jsonBatchFmt{}
+
+type jsonBatchFmt struct{}
+
+func (jb jsonBatchFmt) MediaType() string {
+	return event.ApplicationCloudEventsBatchJSON
+}
+
+func (jb jsonBatchFmt) Marshal(e *event.Event) ([]byte, error) {
+	return nil, errors.New("not supported for batch events")
+}
+
+func (jb jsonBatchFmt) Unmarshal(b []byte, e *event.Event) error {
+	return errors.New("not supported for batch events")
+}
+
 // built-in formats
 var formats map[string]Format
 
 func init() {
 	formats = map[string]Format{}
 	Add(JSON)
+	Add(JSONBatch)
 }
 
 // Lookup returns the format for contentType, or nil if not found.
--- a/v2/binding/to_event.go
+++ b/v2/binding/to_event.go
@@ -77,8 +77,7 @@ func ToEvents(ctx context.Context, message MessageReader, body io.Reader) ([]eve
 	// Since Format doesn't support batch Marshalling, and we know it's structured batch json, we'll go direct to the
 	// json.UnMarshall(), since that is the best way to support batch operations for now.
 	var events []event.Event
-	err := json.NewDecoder(body).Decode(&events)
-	return events, err
+	return events, json.NewDecoder(body).Decode(&events)
 }
 
 type messageToEventBuilder event.Event
--- a/v2/binding/to_event.go
+++ b/v2/binding/to_event.go
@@ -8,6 +8,7 @@ package binding
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -20,6 +21,9 @@ import (
 
 // ErrCannotConvertToEvent is a generic error when a conversion of a Message to an Event fails
 var ErrCannotConvertToEvent = errors.New("cannot convert message to event")
+
+// ErrCannotConvertToEvents is a generic error when a conversion of a Message to a Batched Event fails
+var ErrCannotConvertToEvents = errors.New("cannot convert message to batched events")
 
 // ToEvent translates a Message with a valid Structured or Binary representation to an Event.
 // This function returns the Event generated from the Message and the original encoding of the message or
@@ -59,6 +63,22 @@ func ToEvent(ctx context.Context, message MessageReader, transformers ...Transfo
 		return nil, err
 	}
 	return &e, Transformers(transformers).Transform((*EventMessage)(&e), encoder)
+}
+
+// ToEvents translates a Batch Message and corresponding Reader data to a slice of Events.
+// This function returns the Events generated from the body data, or an error that points
+// to the conversion issue.
+func ToEvents(ctx context.Context, message MessageReader, body io.Reader) ([]event.Event, error) {
+	messageEncoding := message.ReadEncoding()
+	if messageEncoding != EncodingBatch {
+		return nil, ErrCannotConvertToEvents
+	}
+
+	// Since Format doesn't support batch Marshalling, and we know it's structured batch json, we'll go direct to the
+	// json.UnMarshall(), since that is the best way to support batch operations for now.
+	var events []event.Event
+	err := json.NewDecoder(body).Decode(&events)
+	return events, err
 }
 
 type messageToEventBuilder event.Event
--- a/v2/protocol/http/utility.go
+++ b/v2/protocol/http/utility.go
@@ -27,19 +27,20 @@ func NewEventFromHTTPResponse(resp *nethttp.Response) (*event.Event, error) {
 	return binding.ToEvent(context.Background(), msg)
 }
 
-// NewEventsFromHTTPRequest returns a batched set of Events from a http.Request
+// NewEventsFromHTTPRequest returns a batched set of Events from a HTTP Request
 func NewEventsFromHTTPRequest(req *nethttp.Request) ([]event.Event, error) {
 	msg := NewMessageFromHttpRequest(req)
 	return binding.ToEvents(context.Background(), msg, msg.BodyReader)
 }
 
-// NewEventsFromHTTPResponse returns a batched set of Events from a http.Response
+// NewEventsFromHTTPResponse returns a batched set of Events from a HTTP Response
 func NewEventsFromHTTPResponse(resp *nethttp.Response) ([]event.Event, error) {
 	msg := NewMessageFromHttpResponse(resp)
 	return binding.ToEvents(context.Background(), msg, msg.BodyReader)
 }
 
 // NewHTTPRequestFromEvent creates a http.Request object that can be used with any http.Client for a singular event.
+// This is an HTTP POST action to the provided url.
 func NewHTTPRequestFromEvent(ctx context.Context, url string, event event.Event) (*nethttp.Request, error) {
 	if err := event.Validate(); err != nil {
 		return nil, err
@@ -57,7 +58,7 @@ func NewHTTPRequestFromEvent(ctx context.Context, url string, event event.Event)
 }
 
 // NewHTTPRequestFromEvents creates a http.Request object that can be used with any http.Client for sending
-// a batched set of events.
+// a batched set of events. This is an HTTP POST action to the provided url.
 func NewHTTPRequestFromEvents(ctx context.Context, url string, events []event.Event) (*nethttp.Request, error) {
 	// Sending batch events is quite straightforward, as there is only JSON format, so a simple implementation.
 	for _, e := range events {
@@ -81,7 +82,7 @@ func NewHTTPRequestFromEvents(ctx context.Context, url string, events []event.Ev
 	return request, nil
 }
 
-// IsHTTPBatch returns of the current http.Request or http.Response is a batch event operation, by checking the
+// IsHTTPBatch returns if the current http.Request or http.Response is a batch event operation, by checking the
 // header `Content-Type` value.
 func IsHTTPBatch(header nethttp.Header) bool {
 	return header.Get(ContentType) == event.ApplicationCloudEventsBatchJSON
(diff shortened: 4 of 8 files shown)


v2/binding/format/format.go
```diff
@@ -41,12 +42,30 @@ func (jsonFmt) Unmarshal(b []byte, e *event.Event) error {
 	return json.Unmarshal(b, e)
 }
 
+// JSONBatch is the buitl-in "application/cloudevents-batch+json" format.
```
duglin, line 45: s/buitl/built/

markmandel: Nice catch - thank you!

duglin:
How does someone send a batch with this change? I feel like I'm missing something (but it could be me :-) ).

duglin:
doi never mind - I just reread your origin post.

duglin:
@lionelvillard any thoughts on this?

Will this feel incomplete w/o the sending side?

markmandel:
> Will this feel incomplete w/o the sending side?

Thankfully sending of JSON HTTP batch events is quite straightforward if someone wanted to do it by hand  -- but `Client` seems very coupled to singular events, so I couldn't see an easy way to extend it to handle multiple events without breaking every implementation of the Client interface, and having massive changes.

What I could do - which is inline with the utility package, is add a `NewHTTPRequestFromEvents(ctx context.Context, url string, events[]Events) *http.Request` function, which would spit out a `http.Request` object that would send batched events, that the end user could use with their own http client as they so desired.

So an end user could do something like:

```go
var events []event.Event
event :=  cloudevents.NewEvent()
event.SetSource("example/uri")
event.SetType("example.type")
event.SetData(cloudevents.ApplicationJSON, map[string]string{"hello": "world"})

events := append(events, event) // repeat multiple times

request := NewHTTPRequestFromEvents(context.Background(), "http://mywebsiteaddress", events")
client := http.Client{}
resposne, err := client.Do(request)
// do something with the response and error.
```

Pretty easy addition, and allows for sending as well as recieving.

WDYT?

markmandel:
Actually, looking at this, I really like the idea of having a `NewHTTPRequestFromEvents` (and probably a `NewHTTPRequestFromEvent` (singular), because if you pair that with `NewEventFromHTTPResponse` - if the batch event endpoint returns a batched event, you can easily parse this as well.

Basically you can escape the `Client` if you need to break the glass for whatever reason, which is not a bad thing!

markmandel:
Actually, I like the idea of `NewHTTPRequestFromEvents` so much, I implemented it -- so now you can send batched events!

v2/protocol/http/utility_test.go
```diff
@@ -89,3 +90,67 @@ func TestNewEventFromHttpResponse(t *testing.T) {
 		})
 	}
 }
+
+func TestNewEventsFromHTTPRequest(t *testing.T) {
```
lionelvillard, line 94: Can you add a test where `NewEventsFromHTTPRequest` returns an error and maybe another one with more than one event?

lionelvillard:
Have you considered having `NewHTTPRequestFromEvents` also handling single, non-batched events? It feels wrong that the caller must look at in the request headers what's the content mode to determine whether to call `NewHTTPRequestFromEvents` or `NewHTTPRequestFromEvent`.

v2/protocol/http/utility_test.go
markmandel: Thanks for this suggestion! Ended up leading me to the code to see how easy it would be to do a `NewHTTPRequestFromEvent`, so added that in as well.

Updates to tests also implemented.

markmandel:
> Have you considered having NewHTTPRequestFromEvents also handling single, non-batched events?

Hmnn.. that's an interesting question 🤔 

My thinking in this part of the library was to give people the low level building blocks they needed to put together what they needed to parse data as they needed. In which case, checking the `Content-Type` header seems like the appropriate thing to if they wanted a singular endpoint to handle either batch or singular events.

I'd honestly also assumed people would have specific endpoints to handle singular and a different endpoint to handle batched events -- but I could definitely be wrong with that assumption.

Also assumed that people may only want to accept either singular/batched -- so didn't want to bake in any higher level abstractions.

A good middle ground might something like adding a ` func IsRequestBatch(req http.Request) bool` and `func IsResponseBatch(req http.Response) bool` which makes checking the header easy if people want to combine parsing logic into a singular endpoint. WDYT?

lionelvillard:
`func IsRequestBatch(req http.Request) bool` and `func IsResponseBatch(req http.Response) bool` seems like a great addition to me. +1

markmandel:
Done!

I ended up at `IsHTTPBatch(header nethttp.Header) bool` since both `http.Request` and `http.Response` both use the same `http.Header` type, so I figured it made sense to just check that value.

markmandel:
Just gently bumping this PR, see if it's good to merge now?

lionelvillard:
LGTM

@embano1 @duglin wanna take a quick look at this one?

v2/binding/to_event.go
```diff
@@ -61,6 +65,22 @@ func ToEvent(ctx context.Context, message MessageReader, transformers ...Transfo
...
+func ToEvents(ctx context.Context, message MessageReader, body io.Reader) ([]event.Event, error) {
+	messageEncoding := message.ReadEncoding()
+	if messageEncoding != EncodingBatch {
+		return nil, ErrCannotConvertToEvents
+	}
+
+	// Since Format doesn't support batch Marshalling, and we know it's structured batch json, we'll go direct to the
+	// json.UnMarshall(), since that is the best way to support batch operations for now.
+	var events []event.Event
+	err := json.NewDecoder(body).Decode(&events)
```
embano1, line 80: You could directly return here

v2/binding/format/format.go
```diff
@@ -41,12 +42,30 @@ func (jsonFmt) Unmarshal(b []byte, e *event.Event) error {
...
+// JSONBatch is the built-in "application/cloudevents-batch+json" format.
+var JSONBatch = jsonBatchFmt{}
+
+type jsonBatchFmt struct{}
+
+func (jb jsonBatchFmt) MediaType() string {
+	return event.ApplicationCloudEventsBatchJSON
+}
+
+func (jb jsonBatchFmt) Marshal(e *event.Event) ([]byte, error) {
```
embano1, line 54: The error threw me off until I saw your implementation in `to_event.go` - perhaps a short comment that this is handled by the other package?

v2/protocol/http/utility.go
```diff
@@ -24,3 +26,63 @@ func NewEventFromHTTPResponse(resp *nethttp.Response) (*event.Event, error) {
 	msg := NewMessageFromHttpResponse(resp)
 	return binding.ToEvent(context.Background(), msg)
 }
+
+// NewEventsFromHTTPRequest returns a batched set of Events from a http.Request
+func NewEventsFromHTTPRequest(req *nethttp.Request) ([]event.Event, error) {
+	msg := NewMessageFromHttpRequest(req)
+	return binding.ToEvents(context.Background(), msg, msg.BodyReader)
```
embano1, line 33: Question: use req.Context() here? Not sure if that provides value/advantage in this case.

v2/protocol/http/utility.go
```diff
@@ -24,3 +26,63 @@ func NewEventFromHTTPResponse(resp *nethttp.Response) (*event.Event, error) {
...
+}
+
+// NewEventsFromHTTPResponse returns a batched set of Events from a http.Response
+func NewEventsFromHTTPResponse(resp *nethttp.Response) ([]event.Event, error) {
+	msg := NewMessageFromHttpResponse(resp)
+	return binding.ToEvents(context.Background(), msg, msg.BodyReader)
+}
+
+// NewHTTPRequestFromEvent creates a http.Request object that can be used with any http.Client for a singular event.
```
embano1, line 42: nit: line wrapping seem off

v2/protocol/http/utility.go
```diff
@@ -24,3 +26,63 @@ func NewEventFromHTTPResponse(resp *nethttp.Response) (*event.Event, error) {
 	msg := NewMessageFromHttpResponse(resp)
 	return binding.ToEvent(context.Background(), msg)
 }
+
+// NewEventsFromHTTPRequest returns a batched set of Events from a http.Request
```
embano1, line 30: nit: better from a HTTP request ?

v2/binding/to_event.go
```diff
@@ -61,6 +65,22 @@ func ToEvent(ctx context.Context, message MessageReader, transformers ...Transfo
 	return &e, Transformers(transformers).Transform((*EventMessage)(&e), encoder)
 }
 
+// ToEvents translates a Batch Message and corresponding Reader data to a slice of Events.
+// This function returns the Events generated from the body data, or an error that points
+// to the conversion issue.
+func ToEvents(ctx context.Context, message MessageReader, body io.Reader) ([]event.Event, error) {
```
embano1, line 71: Question: can we have a test with a partial invalid event in the batch request body? I would like to understand the semantics of this API, i.e. which states []event.Event can have depending on the input.

v2/protocol/http/utility.go
```diff
@@ -24,3 +26,63 @@ func NewEventFromHTTPResponse(resp *nethttp.Response) (*event.Event, error) {
...
+	return binding.ToEvents(context.Background(), msg, msg.BodyReader)
+}
+
+// NewHTTPRequestFromEvent creates a http.Request object that can be used with any http.Client for a singular event.
+func NewHTTPRequestFromEvent(ctx context.Context, url string, event event.Event) (*nethttp.Request, error) {
+	if err := event.Validate(); err != nil {
+		return nil, err
+	}
+
+	req, err := nethttp.NewRequestWithContext(ctx, nethttp.MethodPost, url, nil)
```
embano1, line 48: nit: I know this is a helper, but since we're hardcoding POST behavior here, would be nice to see this doc-ed in the comment. Not all webhooks use POST, unfortunately.

v2/protocol/http/utility.go
```diff
@@ -24,3 +26,63 @@ func NewEventFromHTTPResponse(resp *nethttp.Response) (*event.Event, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	request.Header.Set(ContentType, event.ApplicationCloudEventsBatchJSON)
+
+	return request, nil
+}
+
+// IsHTTPBatch returns of the current http.Request or http.Response is a batch event operation, by checking the
```
embano1, line 84: nit: returns of?

v2/alias.go
```diff
@@ -135,8 +135,14 @@ var (
...
 	// Event Creation
-	NewEventFromHTTPRequest  = http.NewEventFromHTTPRequest
-	NewEventFromHTTPResponse = http.NewEventFromHTTPResponse
+
+	NewEventFromHTTPRequest   = http.NewEventFromHTTPRequest
+	NewEventFromHTTPResponse  = http.NewEventFromHTTPResponse
+	NewEventsFromHTTPRequest  = http.NewEventsFromHTTPRequest
+	NewEventsFromHTTPResponse = http.NewEventsFromHTTPResponse
+	NewHTTPRequestFromEvents  = http.NewHTTPRequestFromEvents
+	NewHTTPRequestFromEvent   = http.NewHTTPRequestFromEvent
```
embano1, line 144: nit: be consistent and switch order to FromEvent then FromEvents - easy one to miss when going through this list.

v2/protocol/http/utility_test.go
embano1: Can you please add test cases which actually assert an error?

embano1:
Nice work @markmandel! Since you added a bit of code, would be nice to see a coverage report here. IMHO currently you don't have all code paths covered with tests.

markmandel:
> Nice work @markmandel! Since you added a bit of code, would be nice to see a coverage report here. IMHO currently you don't have all code paths covered with tests.

Thanks for the detailed feedback! I'll jump on it.

v2/protocol/http/utility.go
```diff
@@ -24,3 +26,63 @@ func NewEventFromHTTPResponse(resp *nethttp.Response) (*event.Event, error) {
...
+// a batched set of events.
+func NewHTTPRequestFromEvents(ctx context.Context, url string, events []event.Event) (*nethttp.Request, error) {
+	// Sending batch events is quite straightforward, as there is only JSON format, so a simple implementation.
+	for _, e := range events {
+		if err := e.Validate(); err != nil {
+			return nil, err
+		}
+	}
+	var buffer bytes.Buffer
+	err := json.NewEncoder(&buffer).Encode(events)
```
duglin, line 69: Just for my own education, here we just encode the events directly, but above for a single event we use `WriteRequest` which seems to have quite a bit more logic. Is any of that logic needed for batch?

v2/binding/to_event.go
markmandel: Nice trick. I like it 👍🏻

v2/protocol/http/utility.go
markmandel: I was being consistent with all the other utility functions as previously written (they all used `context.Background()`. My suggestion would e that if we want to make that change it should probably be in a separate PR.

v2/protocol/http/utility.go
markmandel: Not quite sure what you mean here. The line is 116 chars (I usually wrap at 120), is that what you are referring to?

v2/protocol/http/utility.go
markmandel: The only way to send batch is in a JSON encoded format, so there is no need for much of that logic.

markmandel:
Okay! I think this is good to re-revue. Implements everything that I didn't have questions for.

As requested, coverage report. I ran it in my IDE and plugged all the holes I could find, which also corresponded to fixing several of the review comments, as well as some other paths as well.

```shell
❯ go test -cover  ./...
?       github.com/cloudevents/sdk-go/v2        [no test files]
ok      github.com/cloudevents/sdk-go/v2/binding        0.028s  coverage: 64.8% of statements
ok      github.com/cloudevents/sdk-go/v2/binding/buffering      0.023s  coverage: 81.0% of statements
ok      github.com/cloudevents/sdk-go/v2/binding/format 0.021s  coverage: 87.0% of statements
ok      github.com/cloudevents/sdk-go/v2/binding/spec   0.014s  coverage: 79.7% of statements
?       github.com/cloudevents/sdk-go/v2/binding/test   [no test files]
ok      github.com/cloudevents/sdk-go/v2/binding/transformer    0.025s  coverage: 85.4% of statements
ok      github.com/cloudevents/sdk-go/v2/binding/utils  0.017s  coverage: 77.8% of statements
ok      github.com/cloudevents/sdk-go/v2/client 6.033s  coverage: 72.5% of statements
ok      github.com/cloudevents/sdk-go/v2/client/test    0.023s  coverage: 67.4% of statements
ok      github.com/cloudevents/sdk-go/v2/context        0.020s  coverage: 65.4% of statements
ok      github.com/cloudevents/sdk-go/v2/event  0.011s  coverage: 89.3% of statements
ok      github.com/cloudevents/sdk-go/v2/event/datacodec        0.019s  coverage: 100.0% of statements
ok      github.com/cloudevents/sdk-go/v2/event/datacodec/json   0.019s  coverage: 100.0% of statements
ok      github.com/cloudevents/sdk-go/v2/event/datacodec/text   0.013s  coverage: 88.9% of statements
ok      github.com/cloudevents/sdk-go/v2/event/datacodec/xml    0.011s  coverage: 100.0% of statements
ok      github.com/cloudevents/sdk-go/v2/extensions     0.015s  coverage: 70.3% of statements
?       github.com/cloudevents/sdk-go/v2/observability  [no test files]
ok      github.com/cloudevents/sdk-go/v2/protocol       0.027s  coverage: 61.5% of statements
ok      github.com/cloudevents/sdk-go/v2/protocol/gochan        0.051s  coverage: 45.0% of statements
ok      github.com/cloudevents/sdk-go/v2/protocol/http  6.046s  coverage: 69.3% of statements
ok      github.com/cloudevents/sdk-go/v2/protocol/test  0.040s  coverage: 54.9% of statements
ok      github.com/cloudevents/sdk-go/v2/test   0.012s  coverage: 47.6% of statements
ok      github.com/cloudevents/sdk-go/v2/types  0.008s  coverage: 89.1% of statements
```

v2/protocol/http/utility.go
markmandel: 🤔 What is there to transform? From my reading of the batch spec, it's JSON in, JSON out.

duglin: LOL I dunno. I just noticed that for the single event it has that transform stuff and I didn't want to miss anything. @lionelvillard any idea what this transform stuff is for and why we need it for single events but not batch?

markmandel: I figure at least we don't need it for the utility functions, since none of the existing functions in `utility.go` don't pass in transformers.

The good thing is, if there was a need to add an optional list of transformers to the method signature isn't going to break.

markmandel:
Just gently bumping the PR, to see if there are any additional changes you would like made?

cncfcloudevents:
LGTM but I defer to @lionelvillard for the final decison

v2/binding/to_event.go
lionelvillard: Ideally `MessageReader` should define `ReadBatch`, even though currently batching mode is only defined by HTTP protocol binding spec. I'm fine with not adding `ReadBatch` but I wonder if it makes sense to match the `ToEvent` signature and do the type switch in the function body. Or have `ToEvents` take a `*Message` instead of `MessageReader`. 

I think it's fine to not initially supporting `transformers` on batched events until someone ask for it.

lionelvillard:
I added a small comment.

v2/binding/to_event.go
markmandel: > Ideally `MessageReader` should define `ReadBatch`, even though currently batching mode is only defined by HTTP protocol binding spec.

I started playing with this, and as soon as that interface changes, it requires changes across lots of codebase, so I'm hesitant to make that change. Or maybe that change should be a subsequent PR for a larger architectural change?

> I'm fine with not adding `ReadBatch` but I wonder if it makes sense to match the `ToEvent` signature and do the type switch in the function body. Or have `ToEvents` take a `*Message` instead of `MessageReader`.

Not sure I'm 100% sure I'm following here. Are you saying that there is a way I could get the `io.Reader` from the `MessageReader`? If so, I'd love some direction - it wasn't immediately obvious to me. 

> 
> I think it's fine to not initially supporting `transformers` on batched events until someone ask for it.

👍🏻

lionelvillard: > Not sure I'm 100% sure I'm following here. Are you saying that there is a way I could get the io.Reader from the MessageReader? If so, I'd love some direction - it wasn't immediately obvious to me.

You can do something like this:

```
switch mt := message.(type) {
case *http.Message:
  body := mt.BodyReader
  // Since Format doesn't support batch Marshalling, and we know it's structured batch json, we'll go direct to the
  // json.UnMarshall(), since that is the best way to support batch operations for now.
  var events []event.Event
  return events, json.NewDecoder(body).Decode(&events)
}
```

Then return an error for all other cases.

Does that make sense?

markmandel: 🤔 interesting. Curious - why have the system return an error at runtime, when you could do the check at compile time?

lionelvillard: `ToEvents` takes a `MessageReader`, which is more general than `http.Message`. In your case you have only one (http) caller but in theory it does not have to (be the only one).

You can either make `ToEvents` to handle only http message, in which case just use `*http.Message` as the second argument type and remove `body`, or make it a bit more generally applicable and consistent with `ToEvent` and pass MessageReader only.

markmandel: I see, I see 🤔 that makes sense -- we could set ourselves up so if more implementations of Batch come down the pipe, we could handle them here in this place, and expand the switch. SGTM. I'll make the change to `Message` and do the type switch 👍🏻

markmandel: Hit a pretty good wrinkle on making this change. I can't reference the concrete type at all within ToEvents -- because I get into a cyclic dependency issue.

```shell
package github.com/cloudevents/sdk-go/v2
        imports github.com/cloudevents/sdk-go/v2/binding
        imports github.com/cloudevents/sdk-go/v2/protocol/http
        imports github.com/cloudevents/sdk-go/v2/binding: import cycle not allowed
```

So my suggestion - leave this as is, and then as there is a need to create more implementation on Batch processing, implement a `BatchEvents` on `MessageReader` and do the work to implement that across the code base. I figure no point in doing all that work, if the demand may not be there.

ghstahl:
@markmandel 

Can we extend this to the ```type Client interface```.

we have
```go
// Send will transmit the given event over the client's configured transport.
Send(ctx context.Context, event event.Event) protocol.Result
```
would like to have the following added.
```go
// Send will transmit the given event over the client's configured transport.
SendEvents(ctx context.Context, events []event.Event) protocol.Result
```
I build out an http protocol where I add ApiKey, basicauth, bearer token stuff and pass it to. 
```go
	c, err := cloudevents_sdk_go.NewClient(p, cloudevents_sdk_go.WithTimeNow(),
		cloudevents_sdk_go.WithUUIDs(),
	)
```

PR #863, opened 2023-02-28
upgrade the sarama version to v1.38.1

yanmxa:
Signed-off-by: myan [myan@redhat.com](mailto:myan@redhat.com)

Upgrade the sarama to v1.38.1, enable to commit the offsets manually, ref: https://github.com/Shopify/sarama/blob/397cee4974f6dfb4387fd701a33881f5c9a2a2b3/offset_manager.go#L25


lionelvillard:
@yanmxa seems like you need to run `go mod tidy`

duglin:
Can you check on the test failures?

PR #865, opened 2023-03-15
claim.Messages() seems returning nil sometime, avoid nil pointer

nbajaj90:
claim.Messages() seems returning nil sometime, avoid nil pointer


rahulchheda (approved):
LGTM

duglin:
Can you please sign your commit and check on the test failures?

duglin:
Actually, it looks like the error may not be PR specific. So just signing your commits should be enough

/LGTM

@lionelvillard  ?

nbajaj90:
@lionelvillard PTAL

PR #737, opened 2021-11-15
pubsub subscription filtering support

irejanodev:
Signed-off-by: Iker Rejano <irejanodev@gmail.com>

PR #866, opened 2023-03-18
Publish a PubSub message WithAttributes

andreas-lindfalk:
This is a complement to #737, so it's possible to publish a PubSub message `WithAttributes` that the filter then can be applied to. Otherwise we have to abuse the extension feature + filter on these attributes with the ”ce-”-prefix which feels like a hack.

### DIFF
--- a/protocol/pubsub/v2/protocol_test.go
+++ b/protocol/pubsub/v2/protocol_test.go
@@ -20,9 +20,9 @@ type testPubsubClient struct {
 	conn *grpc.ClientConn
 }
 
-func (pc *testPubsubClient) NewWithOrderInterceptor(ctx context.Context, projectID, orderingKey string) (*pubsub.Client, error) {
+func (pc *testPubsubClient) NewWithInterceptor(ctx context.Context, projectID string, interceptor grpc.UnaryClientInterceptor) (*pubsub.Client, error) {
 	pc.srv = pstest.NewServer()
-	conn, err := grpc.Dial(pc.srv.Addr, grpc.WithInsecure(), grpc.WithUnaryInterceptor(orderingKeyInterceptor(orderingKey)))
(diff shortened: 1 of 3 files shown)


protocol/pubsub/v2/protocol_test.go
```diff
@@ -73,3 +73,44 @@ func TestPublishMessageHasOrderingKey(t *testing.T) {
 	err = prot.Send(WithOrderingKey(ctx, orderingKey), test.FullMessage())
 	require.NoError(err)
 }
+
+func attributesInterceptor(attributes map[string]string) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		if method == "/google.pubsub.v1.Publisher/Publish" {
+			pr, _ := req.(*pb.PublishRequest)
+			for _, m := range pr.Messages {
```
duglin, line 81: Should we check to make sure the count of the incoming attributes matches the count of the expected attributes to make sure we don't also have extra ones added by mistake?

lionelvillard:
LGTM

protocol/pubsub/v2/protocol_test.go
duglin: @andreas-lindfalk ping

andreas-lindfalk: We have choosen not to use cloudevents and will not ”need this” anymore so do what you want with this pr so to say :)

duglin: ok thanks for the update

duglin:
Closing for now since the author has moved on. We can reopen if someone else sees a need for it.

PR #876, opened 2023-04-06
Add possibility to add a single header field dynamically

creydr:
Currently there is no direct option to add a single header to the context on a "per-request" basis. To add for example a Header before we send a cloud event, we have to do the following:

```
headers := http.HeaderFrom(ctx)
headers.Set("Foo", "bar")
ctx = http.WithCustomHeader(ctx, headers)

client.Send(ctx, ....)
```

This PR addresses it, and adds the possibility to add a single header, so we end up with the following:
```
http.WithCustomHeaderField(ctx, "Foo", "bar")
client.Send(ctx, ....)
```

**Hints for reviewers**
* We have already the possibility to add a single header, when we create the client (`WithHeader()` option). However it would be good to have something more fine grained, which can be set on a "per-request" basis

### DIFF
--- a/v2/protocol/http/headers.go
+++ b/v2/protocol/http/headers.go
@@ -7,11 +7,12 @@ package http
 
 import (
 	"context"
-	"github.com/cloudevents/sdk-go/v2/binding"
 	"net/http"
 	"net/textproto"
 	"strings"
 	"unicode"
+
+	"github.com/cloudevents/sdk-go/v2/binding"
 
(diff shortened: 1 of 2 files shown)


creydr:
/hold
want to do some more testing

duglin:
@creydr any update on this?

duglin:
@creydr did you want to complete this one? If we don't hear from you we'll assume we should close it.

PR #889, opened 2023-04-11
Trying to fix a race warning

duglin:


duglin:
Merging - can't hurt so let's see if it helps with main's tests

PR #892, opened 2023-04-24
fix: Fixing syntax errors and add some test feedback

flc1125:


duglin:
can you please sign your commits?

duglin:
yes but you'll need to do it for all commits. It might be easier to squash it down to one commit and then just sign that one.

duglin:
LGTM
ping @lionelvillard

PR #894, opened 2023-05-08
Convert ECONNRESET error to a normal close to avoid an endless loop in the go routine

nishant-parekh:
This PR is intended to fix an issue we observed on our lab setup where:
    When the connection was reset by a peer the receiver got stuck in this go routine:
    https://github.com/cloudevents/sdk-go/blob/ff0a142752d3253c3b974246c07f450421714f23/v2/client/client.go#L229
    and our logs were flooded by the this warning log..
    https://github.com/cloudevents/sdk-go/blob/ff0a142752d3253c3b974246c07f450421714f23/v2/client/client.go#L248
    
   To get around this per @duglin's suggestion we are converting the `connection reset` error in the respond function to a normal close `EOF` so the go routine would exit normally.

### DIFF
--- a/v2/protocol/http/protocol.go
+++ b/v2/protocol/http/protocol.go
@@ -16,6 +16,7 @@ import (
 	"strconv"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/cloudevents/sdk-go/v2/binding"
@@ -265,7 +266,7 @@ func (p *Protocol) Receive(ctx context.Context) (binding.Message, error) {
 // Respond receives the next incoming HTTP request as a CloudEvent and waits
 // for the response callback to invoked before continuing.
 // Returns non-nil error if the incoming HTTP request fails to parse as a CloudEvent
-// Returns io.EOF if the receiver is closed.
+// Returns io.EOF if the receiver is closed and on ECONNRESET.
 func (p *Protocol) Respond(ctx context.Context) (binding.Message, protocol.ResponseFn, error) {
 	if ctx == nil {
 		return nil, nil, fmt.Errorf("nil Context")
@@ -276,7 +277,9 @@ func (p *Protocol) Respond(ctx context.Context) (binding.Message, protocol.Respo
 		if !ok {
 			return nil, nil, io.EOF
 		}
-
+		if in.err != nil && errors.Is(in.err, syscall.ECONNRESET) {
+			in.err = io.EOF
+		}
 		if in.msg == nil {
 			return nil, in.respFn, in.err
 		}
--- a/v2/client/client.go
+++ b/v2/client/client.go
@@ -210,7 +210,7 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 		return fmt.Errorf("mismatched receiver callback without protocol.Receiver supported by protocol")
 	}
 	if invoker.IsResponder() && c.responder == nil {
-		return fmt.Errorf("mismatched receiver callback without protocol.Responder supported by protocol")
+		return fmt.Errorf("mismatched responder callback without protocol.Responder supported by protocol")
 	}
 	c.invoker = invoker
 
@@ -243,7 +243,6 @@ func (c *ceClient) StartReceiver(ctx context.Context, fn interface{}) error {
 				if err == io.EOF { // Normal close
 					return
 				}
-
 				if err != nil {
 					cecontext.LoggerFrom(ctx).Warn("Error while receiving a message: ", err)
 					continue


duglin:
I wonder if we need a check per protocol - and just do http for now. So, when I look at this:
https://github.com/cloudevents/sdk-go/blob/722f1722eff556978f526badd956bd0ce320a193/v2/protocol/http/protocol.go#L280-L283
it seems like it could just be:
```
	return in.msg, in.respFn, in.err
```
and we can add a check before the "return" for `"connection reset by peer"` in in.err and if there change in.err to be io.EOF.
WDYT?

nishant-parekh:
> I wonder if we need a check per protocol - and just do http for now. So, when I look at this:
> 
> https://github.com/cloudevents/sdk-go/blob/722f1722eff556978f526badd956bd0ce320a193/v2/protocol/http/protocol.go#L280-L283
> 
> 
> it seems like it could just be:
> ```
> 	return in.msg, in.respFn, in.err
> ```
> 
> and we can add a check before the "return" for `"connection reset by peer"` in in.err and if there change in.err to be io.EOF. WDYT?

That would work.. just thinking from a code documentation stand point. I think we need to update the method definition to reflect this error masking for traceability?
https://github.com/cloudevents/sdk-go/blob/ff0a142752d3253c3b974246c07f450421714f23/v2/protocol/http/protocol.go#LL268C1-L268C1

duglin:
@nishant-parekh i assume this fixes your issue, right?

duglin:
> > I wonder if we need a check per protocol - and just do http for now. So, when I look at this:
> > https://github.com/cloudevents/sdk-go/blob/722f1722eff556978f526badd956bd0ce320a193/v2/protocol/http/protocol.go#L280-L283
> > 
> > it seems like it could just be:
> > ```
> > 	return in.msg, in.respFn, in.err
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and we can add a check before the "return" for `"connection reset by peer"` in in.err and if there change in.err to be io.EOF. WDYT?
> 
> That would work.. just thinking from a code documentation stand point. I think we need to update the method definition to reflect this error masking for traceability? https://github.com/cloudevents/sdk-go/blob/ff0a142752d3253c3b974246c07f450421714f23/v2/protocol/http/protocol.go#LL268C1-L268C1

Yup, I think updating the comment makes sense

v2/protocol/http/protocol.go
```diff
@@ -276,9 +278,8 @@ func (p *Protocol) Respond(ctx context.Context) (binding.Message, protocol.Respo
 		if !ok {
 			return nil, nil, io.EOF
 		}
-
-		if in.msg == nil {
-			return nil, in.respFn, in.err
+		if strings.Contains(in.err.Error(), connectionReset) {
```
duglin, line 281: Check if in,err is nil first

v2/protocol/http/protocol.go
```diff
@@ -267,6 +268,7 @@ func (p *Protocol) Receive(ctx context.Context) (binding.Message, error) {
 // Returns non-nil error if the incoming HTTP request fails to parse as a CloudEvent
 // Returns io.EOF if the receiver is closed.
 func (p *Protocol) Respond(ctx context.Context) (binding.Message, protocol.ResponseFn, error) {
+	const connectionReset = "connection reset by peer"
```
embano1, line 271: You can use a typed error here e.g., using `errors.Is()` based on `ECONNRESET` from https://pkg.go.dev/syscall#pkg-constants

embano1:
@nishant-parekh is there an issue describing what this PR is fixing?

v2/protocol/http/protocol.go
duglin: the tests are failing because of a NPE

v2/protocol/http/protocol.go
```diff
@@ -276,7 +279,9 @@ func (p *Protocol) Respond(ctx context.Context) (binding.Message, protocol.Respo
 		if !ok {
 			return nil, nil, io.EOF
 		}
-
+		if &in != nil && in.err != nil && strings.Contains(in.err.Error(), connectionReset) {
```
duglin, line 282: just `in` not `&in`

nishant-parekh: that gives me a type mismatch error as `in` is a struct of `msgErr` I can change it to comparing to an empty struct like `in != msgErr{}`

nishant-parekh:
> @nishant-parekh is there an issue describing what this PR is fixing?

I will add a description soon sorry about that.

nishant-parekh:
> > @nishant-parekh is there an issue describing what this PR is fixing?
> 
> I will add a description soon sorry about that.

Description added

nishant-parekh:
> For some reason Github is not saving my suggestion on the doc string, so here we go:
> 
> ```go
> // Returns io.EOF if the receiver is closed and on ECONNRESET.
> ```

updated

duglin:
LGTM
thanks

duglin:
ping @lionelvillard @dan-j @markpeek for review

dan-j:
I could be missing something, but @nishant-parekh could you confirm if the `ECONNRESET` is coming from the call to the rate-limit service? I can't see any other place where `in.err` could be set, we only write into `p.incoming` in three places and the other two have either no error, or a hardcoded `binding.ErrUnknownEncoding`.

I don't think returning `io.EOF` is going to work because that's going to stop the go-routine, reducing capacity on the receiver. Then if it happens `c.pollGoroutines` times the call to `ceClient.StartReceiver` is going to return. Furthermore it will return with a `nil` error since it's simply waiting on `wg.Wait()` so the user is going to have no idea why the server has stopped.

Can someone confirm if my thoughts are valid?

duglin:
@nishant-parekh i assume this fixes your issue, right? Have you tested it with your usecase?

dan-j:
> @nishant-parekh i assume this fixes your issue, right? Have you tested it with your usecase?

I'd still like to confirm if a `RateLimiter` is being used, I don't see how an `ECONNRESET` could be propagated to where the OP is stating it is doing otherwise (`client.go:248`). If I'm right one of 2 options is happening:

1. The cloudevent request is being closed - We pass `req.Context()` to the `RateLimiter` so if the CE connection is reset by the peer, and the `RateLimiter` is returning `ctx.Err()`
2. The `RateLimiter` is making it's own HTTP request to an external service and it's being closed by the peer.

In either case, I think the solution is to just not publish a `msgErr` to the channel on https://github.com/cloudevents/sdk-go/blob/ff0a142752d3253c3b974246c07f450421714f23/v2/protocol/http/protocol.go#L296. 

If `err == ctx.Err()` (option 1) then we immediately return, otherwise we should respond 500 as we already do.

Something like:

```go
// always apply limiter first using req context
ok, reset, err := p.limiter.Allow(req.Context(), req)
if err != nil {
  if req.Context().Err() !=  {
    // request closed, nothing to do
    return
  }

  rw.WriteHeader(http.StatusInternalServerError)
  return
}
```

If I'm wrong and a `RateLimiter` isn't in use, I'd want to know the source of `ECONNRESET` before approving because I think it's a bug somewhere else.

dan-j:
> @nishant-parekh i assume this fixes your issue, right? Have you tested it with your usecase?

I'd still like to confirm if a `RateLimiter` is being used, I don't see how an `ECONNRESET` could be propagated to where the OP is stating it is doing otherwise (`client.go:248`). If I'm right one of 2 options is happening:

1. The cloudevent request is being closed - We pass `req.Context()` to the `RateLimiter` so if the CE connection is reset by the peer, and the `RateLimiter` is returning `ctx.Err()`
2. The `RateLimiter` is making it's own HTTP request to an external service and it's being closed by the peer.

In either case, I think the solution is to just not publish a `msgErr` to the channel on https://github.com/cloudevents/sdk-go/blob/ff0a142752d3253c3b974246c07f450421714f23/v2/protocol/http/protocol.go#L296

If `err == ctx.Err()` (option 1) then we immediately return, otherwise we should respond 500 as we already do.

Something like:

```go
// always apply limiter first using req context
ok, reset, err := p.limiter.Allow(req.Context(), req)
if err != nil {
  if req.Context().Err() !=  {
    // request closed, nothing to do
    return
  }

  rw.WriteHeader(http.StatusInternalServerError)
  return
}
```

If I'm wrong and a `RateLimiter` isn't in use, I'd want to know the source of `ECONNRESET` before approving because I think it's a bug somewhere else.

nishant-parekh:
> > @nishant-parekh i assume this fixes your issue, right? Have you tested it with your usecase?
> 
> I'd still like to confirm if a `RateLimiter` is being used, I don't see how an `ECONNRESET` could be propagated to where the OP is stating it is doing otherwise (`client.go:248`). If I'm right one of 2 options is happening:
> 
> 1. The cloudevent request is being closed - We pass `req.Context()` to the `RateLimiter` so if the CE connection is reset by the peer, and the `RateLimiter` is returning `ctx.Err()`
> 2. The `RateLimiter` is making it's own HTTP request to an external service and it's being closed by the peer.
> 
> In either case, I think the solution is to just not publish a `msgErr` to the channel on
> 
> https://github.com/cloudevents/sdk-go/blob/ff0a142752d3253c3b974246c07f450421714f23/v2/protocol/http/protocol.go#L296
> 
> If `err == ctx.Err()` (option 1) then we immediately return, otherwise we should respond 500 as we already do.
> 
> Something like:
> 
> ```go
> // always apply limiter first using req context
> ok, reset, err := p.limiter.Allow(req.Context(), req)
> if err != nil {
>   if req.Context().Err() !=  {
>     // request closed, nothing to do
>     return
>   }
> 
>   rw.WriteHeader(http.StatusInternalServerError)
>   return
> }
> ```
> 
> If I'm wrong and a `RateLimiter` isn't in use, I'd want to know the source of `ECONNRESET` before approving because I think it's a bug somewhere else.

So turns out the issue we were tracking was in the AMQ interface and not the http interface..sorry about the confusion. I think I am going to close this PR and track the issue through the AMQ interface code.

Issue #898, opened 2023-05-22
Question: can we support short-circuit logic ?


chengjingtao:
Hi， If I have a key that not exists in cloudevent, Evaluate will return `missing attribute` error.
But can we support short-circuit logic to avoid filling all needed fields empty in cloudevent ?

``` golang

import (
    	cesql "github.com/cloudevents/sdk-go/sql/v2/parser"
        cloudevents "github.com/cloudevents/sdk-go/v2"
)


func TestA(t *testing.T) {

	sql := "(EXISTS revisiontype AND revisiontype = Branch) OR (branch = master)"

	expression, err := cesql.Parse(string(sql))
	evt := cloudevents.NewEvent("1.0")
	evt.SetID("evt-1")
	evt.SetType("what-ever")
	evt.SetSource("/event")
	evt.SetExtension("branch", "master")
	val, err := expression.Evaluate(evt)

	if err != nil {
		t.Errorf("err should be nil: %s", err.Error())
	} else {
		if !val.(bool) {
			t.Errorf("should be true ,but :%s", val)
		}
	}
}


=== RUN   TestShortCircuit
    filters_test.go:49: err should be nil: missing attribute 'revisiontype'
--- FAIL: TestShortCircuit (0.00s)

```

Expected: `expression.Evaluate(evt)` should return true, because "branch = master" matched
Actual: errors happened: missing attribute 'revisiontype'

duglin:
See: https://github.com/cloudevents/sdk-go/pull/899

chengjingtao:
Really a great job 👍🏻

And by the way when we plan to publish a new release ? 
If so, I could use it as a released version.

@duglin @lionelvillard

duglin:
there are a couple of PRs that are really close to being merged - so maybe after those go in we can look at cutting a new release.

chengjingtao:
> 

thanks for your reply. I get it .

PR #899, opened 2023-05-22
Short-circuit AND expressions

duglin:
Fixes: #898

We can short-circuit AND and OR cases based on the left-expression sometimes.


duglin:
ping @lionelvillard @embano1

PR #904, opened 2023-06-02
chore: update samples go.mod

embano1:
Fixes Dependabot issues when upgrading dependencies using Generics or `any`.

Ref: #903


embano1:
@duglin

embano1:
@lionelvillard for opinion

duglin:
@lionelvillard any thoughts on upgrading golang? Or anyone else from the community? Does anyone know what might break by moving to 1.18 from 1.17?

embano1:
Sorry, I probably wasn't too clear on the change/ask here. The change is in the samples, bumping `go.mod` version to `1.18` to keep up with dependencies. The `v2` `go.mod` remains at `1.17`. However, I had to upgrade the CI because it runs across the various `go.mod` files and broke with the sample `1.18` (since an older version was used to compile a breaking release using `any`). 

The README of the SDK states we support `1.17`. My concern was that we're only testing with `1.18` now, i.e., we don't see regressions with `1.17` (unless we add fancy foo to CI which I'm not keen to touch tbh).

We could update the README as well and keep `v2` `go.mod` at `1.17` to not cause downstream impact (on SDK users).

duglin:
rebase needed

embano1:
@duglin going to close this draft PR until we see issues with Dependabot again.

PR #908, opened 2023-07-03
chore: use t.Errorf(...) instead of t.Error(fmt.Sprintf(...))

testwill:


embano1:
@testwill thx for your contribution. Is there a concrete reason for this change since I can't find an issue? If it's about performance improvements, please also file a test benchmark.

duglin:
@testwill please see comment: https://github.com/cloudevents/sdk-go/pull/908#issuecomment-1618167395
and a rebase is needed

Issue #901, opened 2023-05-26
Whether to consider adding MQTT protocol binding


yanmxa:
I found the [cloudevetns/spec](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/mqtt-protocol-binding.md) which describes the MQTT binding, Do we consider adding the implementation for the sdk-go?

duglin:
@yanmxa I don't see it on anyone's TODO list, so if you're offering to submit a PR to add it I'm sure it would be welcome :-)

yanmxa:
@duglin Okay~
Let me add it to my to-do list.

davidjungermann:
This would be great for our use case!

PR #910, opened 2023-07-05
mqtt protocol binding

yanmxa:
Initial implementation of MQTT binding, using https://github.com/eclipse/paho.golang which supports MQTT V5

Note: This is just a basic implementation; some new features which are brought by MQTT 5.0 and other enhancements will be done in future PRs.

resolved: https://github.com/cloudevents/sdk-go/issues/901


duglin:
@yanmxa thanks.  Quick scan looks nice - will review more...

ping @lionelvillard and @embano1 for their review too.

samples/mqtt/README.md
```diff
@@ -0,0 +1,11 @@
+MQTT samples
+
+To run the samples, you need a running MQTT broker.
+
+To run a sample MQTT broker using docker:
+
+```bash
+echo "listener 1883 0.0.0.0
+allow_anonymous true" > samples/mosquitto.conf
```
embano1, line 9: nit: remove echo newline here

samples/mqtt/README.md
```diff
@@ -0,0 +1,11 @@
...
+docker run --rm --name mosquitto -p 1883:1883 -v "$(pwd)/samples/mosquitto.conf:/mosquitto/config/mosquitto.conf" eclipse-mosquitto
```
embano1, line 10: looks like `mosquitto.conf` is not part of the samples here? where is this command supposed to be run (since it refs `pwd`)?

samples/mqtt/sender/main.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		log.Fatalf("failed to create client, %v", err)
+	}
+
+	for i := 0; i < count; i++ {
+		e := cloudevents.NewEvent()
+		e.SetID(uuid.New().String())
+		e.SetType("com.cloudevents.sample.sent")
+		e.SetSource("https://github.com/cloudevents/sdk-go/samples/mqtt/sender")
+		_ = e.SetData(cloudevents.ApplicationJSON, map[string]interface{}{
```
embano1, line 50: nit: you check `err` always, just not here?

samples/mqtt/sender/main.go
```diff
@@ -0,0 +1,65 @@
...
+	clientConfig := &paho.ClientConfig{
+		ClientID: "sender-client-id",
+		Conn:     conn,
+	}
+	cp := &paho.Connect{
+		KeepAlive:  30,
+		CleanStart: true,
+	}
+
+	p, err := cemqtt.New(context.TODO(), clientConfig, cp, "test-topic", nil, 0, false)
```
embano1, line 34: nit: create a `ctx` and reuse instead of TODO

samples/mqtt/receiver/main.go
```diff
@@ -0,0 +1,50 @@
...
+		log.Fatalf("failed to create protocol: %s", err.Error())
+	}
+	defer p.Close(context.Background())
+
+	c, err := cloudevents.NewClient(p)
+	if err != nil {
+		log.Fatalf("failed to create client, %v", err)
+	}
+
+	log.Printf("will listen consuming topic test-topic\n")
```
embano1, line 39: nit: wording?

samples/mqtt/receiver/main.go
```diff
@@ -0,0 +1,50 @@
...
+	clientConfig := &paho.ClientConfig{
+		ClientID: "receiver-client-id",
+		Conn:     conn,
+	}
+	cp := &paho.Connect{
+		KeepAlive:  30,
+		CleanStart: true,
+	}
+
+	p, err := cemqtt.New(context.TODO(), clientConfig, cp, "test-topic", []string{"test-topic"}, 0, false)
```
embano1, line 28: question: is sender topic here required, or would `""` be ok?

yanmxa: The senderTopic isn't necessary for the receiver, `""` is ok.

samples/mqtt/README.md
yanmxa: Done!

samples/mqtt/README.md
yanmxa: @embano1 Sorry for the confusion. I just updated the doc with a simpler command, so that it can run any path.

samples/mqtt/sender/main.go
yanmxa: Done!

samples/mqtt/receiver/main.go
yanmxa: Done!

protocol/mqtt_paho/v2/message_test.go
```diff
@@ -0,0 +1,67 @@
...
+			}
+		})
+	}
+}
+
+func TestReadBinary(t *testing.T) {
+	// msg := &paho.Publish{
+	// 	Payload:    []byte("{hello:world}"),
+	// 	Properties: &paho.PublishProperties{ContentType: event.ApplicationCloudEventsJSON},
+	// }
```
embano1, line 48: remove?

protocol/mqtt_paho/v2/protocol.go
```diff
@@ -0,0 +1,134 @@
+package mqtt_paho
```
embano1, line 1: This file has no tests, can we get some coverage here e.g., using integration tests?

protocol/mqtt_paho/v2/write_message.go
```diff
@@ -0,0 +1,120 @@
...
+
+	_, err := binding.Write(
+		ctx,
+		m,
+		structuredWriter,
+		binaryWriter,
+		transformers...,
+	)
+	return err
+}
```
embano1, line 29: ideally we have tests for this one

embano1:
Overall looking good @yanmxa We just need some more tests, especially around attribute/extension handling (verification) and writing messages.

protocol/mqtt_paho/v2/message.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Check if .Message implements binding.Message
+var (
+	_ binding.Message               = (*Message)(nil)
+	_ binding.MessageMetadataReader = (*Message)(nil)
+)
+
+func NewMessage(msg *paho.Publish) *Message {
+	var f format.Format = nil
+	var v spec.Version = nil
```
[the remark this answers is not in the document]
yanmxa: Done!

protocol/mqtt_paho/v2/message.go
```diff
@@ -0,0 +1,122 @@
...
+	if msg.Properties != nil {
+		// Use Content-type attr to determine if message is structured and
+		// set format.
+		if s := msg.Properties.ContentType; format.IsFormat(s) {
+			f = format.Lookup(s)
+		}
+		// Binary v0.3:
```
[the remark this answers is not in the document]
yanmxa: This is a legacy comment without much meaning. I just deleted it from here~

protocol/mqtt_paho/v2/message_test.go
yanmxa: Done.

protocol/mqtt_paho/v2/write_message.go
yanmxa: As above.

protocol/mqtt_paho/v2/protocol.go
yanmxa: I had planned to add an integration test in future PR, which might need to set some environment in the workflow.
If you think it's better to put it here, I will and these tests tomorrow~

protocol/mqtt_paho/v2/protocol.go
```diff
@@ -0,0 +1,134 @@
...
+
+	topic := cecontext.TopicFrom(ctx)
+	if topic == "" {
+		topic = t.senderTopic
+	}
+
+	msg := &paho.Publish{
+		QoS:    t.qos,
+		Retain: t.retained,
+		Topic:  topic,
```
skeeey, line 74: can we allow to add properties here?

protocol/mqtt_paho/v2/protocol.go
```diff
@@ -0,0 +1,134 @@
...
+
+	_, err = t.client.Publish(ctx, msg)
+	if err != nil {
+		return err
+	}
+	return err
+}
+
+func (t *Protocol) OpenInbound(ctx context.Context) error {
+	t.openerMutex.Lock()
```
skeeey, line 89: why need lock here?

protocol/mqtt_paho/v2/protocol.go
yanmxa: We can add properties in the event extensions and it will be propagated to the `paho.Pulish.Properties`.

protocol/mqtt_paho/v2/protocol.go
yanmxa: This method is to bootstrap a `Receiver/Responder`, it's a blocking call and ctx is used to stop the Inbound message Receiver/Responder. I think that can only be started once, that's the reason a lock is placed here. 

Do I understand correctly? @duglin  If not, please help me to clarify that. Thanks!

embano1: Yes, other protocol implementations (besides AMQP IIUC) implement 

```go
// Opener is the common interface for things that need to be opened.
type Opener interface {
	// OpenInbound is a blocking call and ctx is used to stop the Inbound message Receiver/Responder.
	// Closing the context won't close the Receiver/Responder, aka it won't invoke Close(ctx).
	OpenInbound(ctx context.Context) error
}
```

in the same way using a mutex. It's a bit confusing as blocking can mean many things, in this project it seems to actually mimic a singleton.

protocol/mqtt_paho/v2/protocol.go
yanmxa: @embano1 I added some integration tests today, PTAL~

yanmxa:
> Overall looking good @yanmxa We just need some more tests, especially around attribute/extension handling (verification) and writing messages.

Sure! I added some unit tests and integration test for protocol and writing messages today. Thanks for your review!

protocol/mqtt_paho/v2/write_message.go
```diff
@@ -0,0 +1,126 @@
...
+func (b *pubMessageWriter) SetAttribute(attribute spec.Attribute, value interface{}) error {
+	if attribute.Kind() == spec.DataContentType {
+		if value == nil {
+			b.removeProperty(contentType)
+		}
+		s, err := types.Format(value)
+		if err != nil {
+			return err
+		}
+		b.addProperty(contentType, s)
```
embano1, line 89: unchecked `error` (IIUC, this is copy and paste from other implementations, but perhaps worth checking it here)

embano1:
Looks good @yanmxa , just one final minor nit

protocol/mqtt_paho/v2/write_message.go
yanmxa: @embano1 Nice Catch! Done~

embano1 (approved):
GREAT JOB!

morvencao (commented):
looks good, thanks @yanmxa

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,99 @@
...
+			})
+	})
+}
+
+func testSenderReceiver(t testing.TB, ctx context.Context) (bindings.Sender, bindings.Receiver, func()) {
+	topicName := "test-ce-client-" + uuid.New().String()
+
+	broker := os.Getenv("TEST_MQTT_BROKER")
+	if broker == "" {
+		broker = "127.0.0.1:1883"
```
embano1, line 67: nit: make const

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+				AssertEventEquals(t, eventIn, ConvertEventExtensionsToString(t, eventOut))
+			})
+	})
+}
+
+func testSenderReceiver(t testing.TB, ctx context.Context) (bindings.Sender, bindings.Receiver, func()) {
+	topicName := "test-ce-client-" + uuid.New().String()
+
+	broker := os.Getenv("TEST_MQTT_BROKER")
```
embano1, line 65: nit: make const

embano1: also, where does this env come from?

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,99 @@
...
+	receiverConn, err := net.Dial("tcp", broker)
+	require.NoError(t, err)
+	r, err := mqtt_paho.New(ctx, &paho.ClientConfig{
+		Conn: receiverConn,
+	}, cp, topicName, []string{topicName}, 0, false)
+	require.NoError(t, err)
+	require.NotNil(t, r)
+
+	go func() {
+		require.NoError(t, r.OpenInbound(context.TODO()))
```
embano1, line 91: failing in a separate gorouine is not allowed: https://pkg.go.dev/testing#T.FailNow

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}()
+
+	time.Sleep(5 * time.Millisecond) // let the receiver ready
```
embano1, line 94: you can avoid this (anti-pattern) and the issue with failing in a goroutine above using channels and a ctx with timeout. please rewrite accordingly 🙏🏻

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,99 @@
...
+				eventOut := MustToEvent(t, ctx, out)
+				assert.Equal(t, binding.EncodingBinary, out.ReadEncoding())
+				eventOut = ConvertEventExtensionsToString(t, eventOut)
+
+				AssertEventEquals(t, eventIn, ConvertEventExtensionsToString(t, eventOut))
+			})
+	})
+}
+
+func testSenderReceiver(t testing.TB, ctx context.Context) (bindings.Sender, bindings.Receiver, func()) {
```
embano1, line 62: lower case test?

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: Done~

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: I used the environment for local tests, and now remove it from here.

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: Nice Catch! Remove the assert from the goroutine.

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: Sorry for the unclear naming. It's only for getting the sender and receiver, so change it to `getSenderReceiver`.

yanmxa:
@embano1 Thank you very much for checking the code so carefully, which allows me to avoid similar problems in the future.

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/cloudevents/sdk-go/v2/binding"
+	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/cloudevents/sdk-go/v2/protocol"
+
+	. "github.com/cloudevents/sdk-go/v2/binding/test"
+	. "github.com/cloudevents/sdk-go/v2/test"
+)
+
+const (
+	Broker = "127.0.0.1:1883"
```
embano1, line 25: nit: no need to export IMHO

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,140 @@
...
+		select {
+		case <-ctx.Done():
+			require.Fail(t, "timeout waiting for event")
+			return
+		case receiveMsg := <-receiveChan:
+			require.NoError(t, receiveMsg.err)
+			outAssert(receiveMsg.msg)
+			return
+		case <-time.After(2 * time.Second):
+			// the receiver mightn't be ready before the sender send the message, so we retry
```
embano1, line 91: nit: this implementation wasn't obvious. consider adding moving the comment below and add a `continue` statement so it's clear you want to retry (loop)

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,140 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			require.Fail(t, "timeout waiting for event")
+			return
+		case receiveMsg := <-receiveChan:
+			require.NoError(t, receiveMsg.err)
+			outAssert(receiveMsg.msg)
+			return
+		case <-time.After(2 * time.Second):
```
embano1, line 90: nit: not an issue in this test, but time.After can leak (consider using Ticker). check google for leaks ,)

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,140 @@
...
+			func(out binding.Message) {
+				assert.Equal(t, binding.EncodingStructured, out.ReadEncoding())
+				eventOut := MustToEvent(t, ctx, out)
+				AssertEventEquals(t, eventIn, ConvertEventExtensionsToString(t, eventOut))
+			})
+	})
+}
+
+func TestSendBinaryMessageToBinary(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
```
embano1, line 53: nit: defer cancel missing

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,140 @@
...
+	s, r, c, err := getSenderReceiver(ctx, cancel)
```
embano1, line 54: nit: generally I find this code hard to read, not blocking you here though

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,140 @@
...
+	require.NoError(t, err)
+	defer c()
```
embano1, line 56: especially this one is just ???

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	EachEvent(t, Events(), func(t *testing.T, eventIn event.Event) {
+		eventIn = ConvertEventExtensionsToString(t, eventIn)
+		in := MustCreateMockStructuredMessage(t, eventIn)
+
+		sendReceive(t, binding.WithPreferredEventEncoding(ctx, binding.EncodingStructured), s, r, in,
```
embano1, line 43: do we really need that many levels of indirection? can't we keep the whole code within the test function? I have a feeling we're overcomplicating the test

test/integration/mqtt_paho_binding/mqtt_test.go
embano1: also, since you retry anyways, do we need all these sleeps?

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -0,0 +1,140 @@
...
+			return
+		case <-time.After(2 * time.Second):
+			// the receiver mightn't be ready before the sender send the message, so we retry
+			result := sender.Send(ctx, in)
+			require.NoError(t, result)
+		}
+	}
+}
+
+func getSenderReceiver(ctx context.Context, cancel context.CancelFunc) (
```
embano1, line 98: that whole code is really hard to follow (and thus review)

embano1: may be want to go back to the drawing board, thinking about what you're trying to accomplish and then simplify within one Test function?

embano1:
> @embano1 Thank you very much for checking the code so carefully, which allows me to avoid similar problems in the future.

No worries, that's my duty here ;) You're doing great work! Just try to keep the code simple and clean - your future self (and other contributors) will appreciate it ;)

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: Yes, I think we need to wait here, otherwise, it will throw the error message like this:
```bash
{"level":"error","ts":1688780919.884921,"logger":"fallback","caller":"client/invoker.go:77","msg":"call to Invoker.Invoke(...) has panicked: send on closed channel",
```

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: Thank you for the reminder @embano1! 
It really has a race condition to cause the memory leak when using `time.After(2 * time.Second) `in such a scenario.

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: Okay, to make it clear, I modified it to avoid the shorthand。

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: I try to reuse `sendReceive` in `TestSendStructuredMessageToStructured` and `TestSendBinaryMessageToBinary`. 

Of course, we can expand these codes and put them here. In this case, there will be some redundant codes in these two tests.
If you prefer to expand it, please reply directly to me and I can remove this layer from here.

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: Done.

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: Done!

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: Done!

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: Done!

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: As mentioned above, I have now changed it to try to avoid using this seemingly more complicated function.

yanmxa:
@embano1 Thanks! It was indeed too complicated to write tests with reference to other tests before, but it should be simplified a lot now compared to before.

test/integration/mqtt_paho_binding/mqtt_test.go
```diff
@@ -2,139 +2,130 @@ package mqtt_paho_binding
...
+						require.Fail(t, "timeout waiting for event")
+						return
+					case receiveMsg := <-receiveChan:
+						require.NoError(t, receiveMsg.err)
+						outMessage := receiveMsg.msg
+						assert.Equal(t, tt.expectedEncoding, outMessage.ReadEncoding())
+						outEvent := MustToEvent(t, ctx, outMessage)
+						AssertEventEquals(t, inEvent, ConvertEventExtensionsToString(t, outEvent))
+						return
+					default:
```
embano1, line 85: nit: just be aware that this is a hot loop now, ideally should also be a ticker, right?

embano1 (approved):
Approving, just one minor nit if you feel it's worth the change

test/integration/mqtt_paho_binding/mqtt_test.go
yanmxa: @embano1 Yes, It's a hot loop! 
Since the `default` case causes a bit of confusion here, I turned it into an explicit ticker.

embano1 (approved):
LGTM

Feel free to squash and we should be good to go cc/ @duglin

lionelvillard:
Overall it looks good, though I'm not an MQTT expert so there is so much I can review. Please add the copyright header to all files and then I think we can merge this PR.

lionelvillard:
looks good. @duglin, ok with you?

lionelvillard:
@embano1 approved it
LGTM

PR #916, opened 2023-07-10
fix: Close response body between retries

taehioum:
This PR aims to address a memory leak issue our team have observed, which was caused by the response body of messages not being closed between retries.

The consumer of binding.Message usually closes the response body by calling msg.Finish().
During retries, only the last message is returned to the consumer; so I think we should close previous message's response body via Finish() before proceeding with the next retry.

### DIFF
--- a/v2/protocol/http/protocol_retry.go
+++ b/v2/protocol/http/protocol_retry.go
@@ -12,7 +12,6 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
-	"net/url"
 	"time"
 
 	"go.uber.org/zap"
@@ -52,7 +51,7 @@ func (p *Protocol) doOnce(req *http.Request) (binding.Message, protocol.Result) 
 }
 
 func (p *Protocol) doWithRetry(ctx context.Context, params *cecontext.RetryParams, req *http.Request) (binding.Message, error) {
-	then := time.Now()
+	start := time.Now()
 	retry := 0
 	results := make([]protocol.Result, 0)
 
@@ -79,51 +78,34 @@ func (p *Protocol) doWithRetry(ctx context.Context, params *cecontext.RetryParam
 
 		// Fast track common case.
 		if protocol.IsACK(result) {
-			return msg, NewRetriesResult(result, retry, then, results)
+			return msg, NewRetriesResult(result, retry, start, results)
 		}
 
-		// Try again?
-		//
-		// Make sure the error was something we should retry.
-
-		{
-			var uErr *url.Error
-			if errors.As(result, &uErr) {
-				goto DoBackoff
+		var httpResult *Result
+		if errors.As(result, &httpResult) {
+			sc := httpResult.StatusCode
+			if !p.isRetriableFunc(sc) {
+				cecontext.LoggerFrom(ctx).Debugw("status code not retryable, will not try again",
+					zap.Error(httpResult),
+					zap.Int("statusCode", sc))
+				return msg, NewRetriesResult(result, retry, start, results)
 			}
 		}
-
-		{
-			var httpResult *Result
-			if errors.As(result, &httpResult) {
-				sc := httpResult.StatusCode
-				if p.isRetriableFunc(sc) {
-					// retry!
-					goto DoBackoff
-				} else {
-					// Permanent error
-					cecontext.LoggerFrom(ctx).Debugw("status code not retryable, will not try again",
-						zap.Error(httpResult),
-						zap.Int("statusCode", sc))
-					return msg, NewRetriesResult(result, retry, then, results)
-				}
-			}
-		}
-
-	DoBackoff:
-		resetBody(req, body)
-
-		// Wait for the correct amount of backoff time.
 
 		// total tries = retry + 1
-		if err := params.Backoff(ctx, retry+1); err != nil {
+		if err = params.Backoff(ctx, retry+1); err != nil {
 			// do not try again.
 			cecontext.LoggerFrom(ctx).Debugw("backoff error, will not try again", zap.Error(err))
-			return msg, NewRetriesResult(result, retry, then, results)
+			return msg, NewRetriesResult(result, retry, start, results)
 		}
 
 		retry++
+		resetBody(req, body)
 		results = append(results, result)
+		if msg != nil {
+			// avoid leak, forget message, ignore error
+			_ = msg.Finish(nil)
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


embano1:
@taehioum thx for this PR! I need to think more about the implications of this small change (which sounds reasonable), because we're dealing with a lot of pointers and body readers/closers internally which I need to get my head around again.

embano1:
I went through the code an IMHO the change is fine. However, I'm suggesting improving the retry logic since we're on this code to simplify it as I had a hard time getting my head around:

```patch
diff --git a/v2/protocol/http/protocol_retry.go b/v2/protocol/http/protocol_retry.go
index 71e7346..3a09c31 100644
--- a/v2/protocol/http/protocol_retry.go
+++ b/v2/protocol/http/protocol_retry.go
@@ -12,7 +12,6 @@
 	"io"
 	"io/ioutil"
 	"net/http"
-	"net/url"
 	"time"
 
 	"go.uber.org/zap"
@@ -52,7 +51,7 @@ func (p *Protocol) doOnce(req *http.Request) (binding.Message, protocol.Result)
 }
 
 func (p *Protocol) doWithRetry(ctx context.Context, params *cecontext.RetryParams, req *http.Request) (binding.Message, error) {
-	then := time.Now()
+	start := time.Now()
 	retry := 0
 	results := make([]protocol.Result, 0)
 
@@ -79,51 +78,34 @@ func (p *Protocol) doWithRetry(ctx context.Context, params *cecontext.RetryParam
 
 		// Fast track common case.
 		if protocol.IsACK(result) {
-			return msg, NewRetriesResult(result, retry, then, results)
+			return msg, NewRetriesResult(result, retry, start, results)
 		}
 
-		// Try again?
-		//
-		// Make sure the error was something we should retry.
-
-		{
-			var uErr *url.Error
-			if errors.As(result, &uErr) {
-				goto DoBackoff
+		var httpResult *Result
+		if errors.As(result, &httpResult) {
+			sc := httpResult.StatusCode
+			if !p.isRetriableFunc(sc) {
+				cecontext.LoggerFrom(ctx).Debugw("status code not retryable, will not try again",
+					zap.Error(httpResult),
+					zap.Int("statusCode", sc))
+				return msg, NewRetriesResult(result, retry, start, results)
 			}
 		}
 
-		{
-			var httpResult *Result
-			if errors.As(result, &httpResult) {
-				sc := httpResult.StatusCode
-				if p.isRetriableFunc(sc) {
-					// retry!
-					goto DoBackoff
-				} else {
-					// Permanent error
-					cecontext.LoggerFrom(ctx).Debugw("status code not retryable, will not try again",
-						zap.Error(httpResult),
-						zap.Int("statusCode", sc))
-					return msg, NewRetriesResult(result, retry, then, results)
-				}
-			}
-		}
-
-	DoBackoff:
-		resetBody(req, body)
-
-		// Wait for the correct amount of backoff time.
-
 		// total tries = retry + 1
-		if err := params.Backoff(ctx, retry+1); err != nil {
+		if err = params.Backoff(ctx, retry+1); err != nil {
 			// do not try again.
 			cecontext.LoggerFrom(ctx).Debugw("backoff error, will not try again", zap.Error(err))
-			return msg, NewRetriesResult(result, retry, then, results)
+			return msg, NewRetriesResult(result, retry, start, results)
 		}
 
 		retry++
+		resetBody(req, body)
 		results = append(results, result)
+		if msg != nil {
+			// avoid leak, forget message, ignore error
+			_ = msg.Finish(nil)
+		}
 	}
 }
 
diff --git a/v2/protocol/http/protocol_retry_test.go b/v2/protocol/http/protocol_retry_test.go
index 425b98a..45612ee 100644
--- a/v2/protocol/http/protocol_retry_test.go
+++ b/v2/protocol/http/protocol_retry_test.go
@@ -14,6 +14,7 @@
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zaptest"
 
 	"github.com/cloudevents/sdk-go/v2/binding"
 	cecontext "github.com/cloudevents/sdk-go/v2/context"
@@ -192,6 +193,7 @@ func TestRequestWithRetries_linear(t *testing.T) {
 			}
 
 			ctx := cecontext.WithTarget(context.Background(), srv.URL)
+			ctx = cecontext.WithLogger(ctx, zaptest.NewLogger(t).Sugar())
 			ctxWithRetries := cecontext.WithRetriesLinearBackoff(ctx, tc.delay, tc.retries)
 
 			dummyMsg := binding.ToMessage(&tc.event)
```

taehioum:
@embano1 Thanks for the suggestion! I will include the changes in this pr as well.

embano1:
@taehioum thx

Feel free to take the patch as is (includes your change too)

taehioum:
@embano1 I have applied your changes, and also added a small change : replaced deprecated ioutil package to io.

taehioum:
@embano1  I have applied your changes.

embano1:
@duglin to approve workflow runs/tests

embano1 (approved):

v2/protocol/http/protocol_retry.go
```diff
@@ -79,51 +78,34 @@ func (p *Protocol) doWithRetry(ctx context.Context, params *cecontext.RetryParam
...
-		if err := params.Backoff(ctx, retry+1); err != nil {
+		if err = params.Backoff(ctx, retry+1); err != nil {
 			// do not try again.
 			cecontext.LoggerFrom(ctx).Debugw("backoff error, will not try again", zap.Error(err))
-			return msg, NewRetriesResult(result, retry, then, results)
+			return msg, NewRetriesResult(result, retry, start, results)
 		}
 
 		retry++
+		resetBody(req, body)
```
duglin, line 103: Overall this looks ok to me but I have a question.... more of a golang question... below in `resetBody` we'll reset the req.Body appropriately, but then only set req.GetBody once, if ever, based on whether it's already set or not. Now, it seems to me that the Reader created on line 124 (new numbering) is never reset - so that would then imply that GetBody() is never actually called during a retry. Do I have the right? If so, do we need lines 121-126 at all?

embano1: Tbh, I don't fully recall why I had to add these lines as part of https://github.com/cloudevents/sdk-go/issues/773 but IIRC it was because of a `nil` pointer panic during tests with `httptest` using `GetBody`. Looks like our code base is not using `GetBody` at all, just setting it in `v2/protocol/http/write_request.go`, with a `default` case not setting `GetBody`. 

IMHO it's correct to only set it once (ever) since it's caller e.g., `httptest`, then must handle the case of re-reading a previously read body. But I could be totally wrong :)

embano1:
Going to merge, unless you object @duglin

duglin:
go for it

PR #924, opened 2023-07-20
provide publish/subscribe options for MQTT protocol

yanmxa:

Provide more flexible parameter configurations when using the MQTT protocol to send events.

### DIFF
--- a/protocol/mqtt_paho/v2/option.go
+++ b/protocol/mqtt_paho/v2/option.go
@@ -1,0 +1,48 @@
+/*
+ Copyright 2023 The CloudEvents Authors
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+package mqtt_paho
+
+import (
+	"fmt"
+
+	"github.com/eclipse/paho.golang/paho"
+)
+
+// Option is the function signature required to be considered an mqtt_paho.Option.
+type Option func(*Protocol) error
+
+// WithConnect sets the paho.Connect configuration for the client. This option is not required.
+func WithConnect(connOpt *paho.Connect) Option {
+	return func(p *Protocol) error {
+		if connOpt == nil {
+			return fmt.Errorf("the paho.Connect option must not be nil")
+		}
+		p.connOption = connOpt
+		return nil
+	}
+}
+
+// WithPublish sets the paho.Publish configuration for the client. This option is required if you want to send messages.
+func WithPublish(publishOpt *paho.Publish) Option {
+	return func(p *Protocol) error {
+		if publishOpt == nil {
+			return fmt.Errorf("the paho.Publish option must not be nil")
+		}
+		p.publishOption = publishOpt
+		return nil
+	}
+}
+
+// WithSubscribe sets the paho.Subscribe configuration for the client. This option is required if you want to receive messages.
+func WithSubscribe(subscribeOpt *paho.Subscribe) Option {
+	return func(p *Protocol) error {
+		if subscribeOpt == nil {
+			return fmt.Errorf("the paho.Subscribe option must not be nil")
+		}
+		p.subscribeOption = subscribeOpt
+		return nil
+	}
+}
--- a/protocol/mqtt_paho/v2/protocol.go
+++ b/protocol/mqtt_paho/v2/protocol.go
@@ -19,12 +19,11 @@ import (
 )
 
 type Protocol struct {
-	client         *paho.Client
-	connConfig     *paho.Connect
-	senderTopic    string
-	receiverTopics []string
-	qos            byte
-	retained       bool
+	client          *paho.Client
+	config          *paho.ClientConfig
+	connOption      *paho.Connect
+	publishOption   *paho.Publish
+	subscribeOption *paho.Subscribe
 
 	// receiver
 	incoming chan *paho.Publish
@@ -41,80 +40,89 @@ var (
 	_ protocol.Closer   = (*Protocol)(nil)
 )
 
-func New(ctx context.Context, clientConfig *paho.ClientConfig, connConfig *paho.Connect, SenderTopic string,
-	ReceiverTopics []string, qos byte, retained bool,
-) (*Protocol, error) {
-	client := paho.NewClient(*clientConfig)
-	ca, err := client.Connect(ctx, connConfig)
+func New(ctx context.Context, config *paho.ClientConfig, opts ...Option) (*Protocol, error) {
+	if config == nil {
+		return nil, fmt.Errorf("the paho.ClientConfig must not be nil")
+	}
+
+	p := &Protocol{
+		client: paho.NewClient(*config),
+		// default connect option
+		connOption: &paho.Connect{
+			KeepAlive:  30,
+			CleanStart: true,
+		},
+		incoming:  make(chan *paho.Publish),
+		closeChan: make(chan struct{}),
+	}
+	if err := p.applyOptions(opts...); err != nil {
+		return nil, err
+	}
+
+	// Connect to the MQTT broker
+	connAck, err := p.client.Connect(ctx, p.connOption)
 	if err != nil {
 		return nil, err
 	}
-	if ca.ReasonCode != 0 {
-		return nil, fmt.Errorf("failed to connect to %s : %d - %s", client.Conn.RemoteAddr(), ca.ReasonCode,
-			ca.Properties.ReasonString)
+	if connAck.ReasonCode != 0 {
+		return nil, fmt.Errorf("failed to connect to %q : %d - %q", p.client.Conn.RemoteAddr(), connAck.ReasonCode,
+			connAck.Properties.ReasonString)
 	}
 
-	return &Protocol{
-		client:         client,
-		connConfig:     connConfig,
-		senderTopic:    SenderTopic,
-		receiverTopics: ReceiverTopics,
-		qos:            qos,
-		retained:       retained,
-		incoming:       make(chan *paho.Publish),
-		openerMutex:    sync.Mutex{},
-		closeChan:      make(chan struct{}),
-	}, nil
+	return p, nil
 }
 
-func (t *Protocol) Send(ctx context.Context, m binding.Message, transformers ...binding.Transformer) error {
+func (p *Protocol) applyOptions(opts ...Option) error {
+	for _, fn := range opts {
+		if err := fn(p); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (p *Protocol) Send(ctx context.Context, m binding.Message, transformers ...binding.Transformer) error {
+	if p.publishOption == nil {
+		return fmt.Errorf("the paho.Publish option must not be nil")
+	}
+
 	var err error
 	defer m.Finish(err)
 
-	topic := cecontext.TopicFrom(ctx)
-	if topic == "" {
-		topic = t.senderTopic
+	msg := p.publishOption
+	if cecontext.TopicFrom(ctx) != "" {
+		msg.Topic = cecontext.TopicFrom(ctx)
+		cecontext.WithTopic(ctx, "")
 	}
 
-	msg := &paho.Publish{
-		QoS:    t.qos,
-		Retain: t.retained,
-		Topic:  topic,
-	}
 	err = WritePubMessage(ctx, m, msg, transformers...)
 	if err != nil {
 		return err
 	}
 
-	_, err = t.client.Publish(ctx, msg)
+	_, err = p.client.Publish(ctx, msg)
 	if err != nil {
 		return err
 	}
 	return err
 }
 
-func (t *Protocol) OpenInbound(ctx context.Context) error {
-	t.openerMutex.Lock()
-	defer t.openerMutex.Unlock()
+func (p *Protocol) OpenInbound(ctx context.Context) error {
+	if p.subscribeOption == nil {
+		return fmt.Errorf("the paho.Subscribe option must not be nil")
+	}
+
+	p.openerMutex.Lock()
+	defer p.openerMutex.Unlock()
 
 	logger := cecontext.LoggerFrom(ctx)
 
-	t.client.Router = paho.NewSingleHandlerRouter(func(m *paho.Publish) {
-		t.incoming <- m
+	p.client.Router = paho.NewSingleHandlerRouter(func(m *paho.Publish) {
+		p.incoming <- m
 	})
 
-	subs := make(map[string]paho.SubscribeOptions)
-	for _, topic := range t.receiverTopics {
-		subs[topic] = paho.SubscribeOptions{
-			QoS:               t.qos,
-			RetainAsPublished: t.retained,
-		}
-	}
-
-	logger.Infof("subscribe to topics: %v", t.receiverTopics)
-	_, err := t.client.Subscribe(ctx, &paho.Subscribe{
-		Subscriptions: subs,
-	})
+	logger.Infof("subscribing to topics: %v", p.subscribeOption.Subscriptions)
+	_, err := p.client.Subscribe(ctx, p.subscribeOption)
 	if err != nil {
 		return err
 	}
@@ -122,15 +130,15 @@ func (t *Protocol) OpenInbound(ctx context.Context) error {
 	// Wait until external or internal context done
 	select {
 	case <-ctx.Done():
-	case <-t.closeChan:
+	case <-p.closeChan:
 	}
-	return t.client.Disconnect(&paho.Disconnect{ReasonCode: 0})
+	return p.client.Disconnect(&paho.Disconnect{ReasonCode: 0})
 }
 
 // Receive implements Receiver.Receive
-func (t *Protocol) Receive(ctx context.Context) (binding.Message, error) {
+func (p *Protocol) Receive(ctx context.Context) (binding.Message, error) {
 	select {
-	case m, ok := <-t.incoming:
+	case m, ok := <-p.incoming:
 		if !ok {
 			return nil, io.EOF
 		}
--- a/samples/mqtt/sender/main.go
+++ b/samples/mqtt/sender/main.go
@@ -15,7 +15,7 @@ import (
 	"github.com/eclipse/paho.golang/paho"
 	"github.com/google/uuid"
 
-	cemqtt "github.com/cloudevents/sdk-go/protocol/mqtt_paho/v2"
+	mqtt_paho "github.com/cloudevents/sdk-go/protocol/mqtt_paho/v2"
 	cecontext "github.com/cloudevents/sdk-go/v2/context"
 )
 
@@ -29,18 +29,19 @@ func main() {
 	if err != nil {
 		log.Fatalf("failed to connect to mqtt broker: %s", err.Error())
 	}
-	clientConfig := &paho.ClientConfig{
+	config := &paho.ClientConfig{
 		ClientID: "sender-client-id",
 		Conn:     conn,
 	}
-	cp := &paho.Connect{
+	// optional connect option
+	connOpt := &paho.Connect{
 		KeepAlive:  30,
 		CleanStart: true,
 	}
 	// set a default topic with test-topic1
-	p, err := cemqtt.New(ctx, clientConfig, cp, "test-topic1", nil, 0, false)
+	p, err := mqtt_paho.New(ctx, config, mqtt_paho.WithPublish(&paho.Publish{Topic: "test-topic1"}), mqtt_paho.WithConnect(connOpt))
 	if err != nil {
-		log.Fatalf("failed to create protocol: %s", err.Error())
+		log.Fatalf("failed to create protocol: %v", err)
 	}
 	defer p.Close(ctx)
 
--- a/samples/mqtt/receiver/main.go
+++ b/samples/mqtt/receiver/main.go
@@ -11,7 +11,7 @@ import (
 	"log"
 	"net"
 
-	cemqtt "github.com/cloudevents/sdk-go/protocol/mqtt_paho/v2"
+	mqtt_paho "github.com/cloudevents/sdk-go/protocol/mqtt_paho/v2"
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"github.com/eclipse/paho.golang/paho"
 )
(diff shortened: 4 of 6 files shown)


protocol/mqtt_paho/v2/context.go
```diff
@@ -0,0 +1,43 @@
...
+var protocolContextKey = protocolContextKeyType{}
+
+// WithProtocolContext return a context with the given ProtocolContext into the provided context object.
+func WithProtocolContext(ctx context.Context, pctx ProtocolContext) context.Context {
+	return context.WithValue(ctx, protocolContextKey, pctx)
+}
+
+// ProtocolContextFrom pulls a ProtocolContext out of a context. Always
+// returns a non-nil object.
+func ProtocolContextFrom(ctx context.Context) ProtocolContext {
```
duglin, line 32: Just wondering if it would be better to return nil if there isn't one and then we can avoid the deepEqual below and just check against nil?

protocol/mqtt_paho/v2/protocol.go
```diff
@@ -71,14 +72,20 @@ func (t *Protocol) Send(ctx context.Context, m binding.Message, transformers ...
...
 
-	topic := cecontext.TopicFrom(ctx)
-	if topic == "" {
-		topic = t.senderTopic
+	topic := t.senderTopic
+	retain := t.retained
+	qos := t.qos
+
+	protocolCtx := ProtocolContextFrom(ctx)
+	if !reflect.DeepEqual(protocolCtx, ProtocolContext{}) {
```
duglin, line 80: Just want to make sure that you meant for this to be an all-or-nothing step. Meaning, the context can't set just one of he three without clearing the other two values. Was that intended ?

embano1: Agree, let's return nil if the value cannot be found (follows Go ctx conventions IMHO) and avoid reflect here by simply using a `if protocolCtx != nil` check

embano1: Also, can we get some coverage for these changes?

protocol/mqtt_paho/v2/protocol.go
```diff
@@ -71,14 +72,20 @@ func (t *Protocol) Send(ctx context.Context, m binding.Message, transformers ...
...
+		topic = protocolCtx.Topic
+		retain = protocolCtx.Retained
+		qos = protocolCtx.QoS
 	}
 
 	msg := &paho.Publish{
-		QoS:    t.qos,
-		Retain: t.retained,
+		QoS:    qos,
+		Retain: retain,
```
embano1, line 88: nit: please reorder, following topic/retain/qos flow from above

protocol/mqtt_paho/v2/context.go
```diff
@@ -0,0 +1,43 @@
...
+	pctx := ctx.Value(protocolContextKey)
+	if pctx != nil {
+		if tx, ok := pctx.(ProtocolContext); ok {
+			return tx
+		}
+		if tx, ok := pctx.(*ProtocolContext); ok {
```
embano1, line 38: would that be ever a scenario?

protocol/mqtt_paho/v2/context.go
embano1: +1 make pointer type and return &val

protocol/mqtt_paho/v2/context.go
```diff
@@ -0,0 +1,43 @@
...
+	return context.WithValue(ctx, protocolContextKey, pctx)
+}
+
+// ProtocolContextFrom pulls a ProtocolContext out of a context. Always
+// returns a non-nil object.
+func ProtocolContextFrom(ctx context.Context) ProtocolContext {
+	pctx := ctx.Value(protocolContextKey)
+	if pctx != nil {
+		if tx, ok := pctx.(ProtocolContext); ok {
+			return tx
```
embano1, line 36: nit: rename to val

protocol/mqtt_paho/v2/context.go
yanmxa: @duglin @embano1 Thanks for the reminder, it's better to use a pointer!

protocol/mqtt_paho/v2/protocol.go
yanmxa: @duglin Yes. 
If we just set the `Topic` in `ProtocolContext`, then `QoS` and `Retained` will use the default `0 `and `false`, it seems that we can't distinguish whether they are set by the user or used by the default. 
So I deal with it like this here. If you use ProtocolContext, I think you have set these three values. do you have a better suggestion?

protocol/mqtt_paho/v2/context.go
yanmxa: You are right, we should only fix to one type.

protocol/mqtt_paho/v2/protocol.go
yanmxa: @embano1 I used the `ProtocolContext` in the integration test to verify that it works as expected.

protocol/mqtt_paho/v2/protocol.go
yanmxa: Done.

protocol/mqtt_paho/v2/context.go
yanmxa: Done.

protocol/mqtt_paho/v2/protocol.go
duglin: For topic/string we can look for a non-empty string to know if they set it in the context. But for `qos` and `retained` since the default values for those types are also valid mqtt values I guess we're stuck.

One option I can think of would be to split them out from the struct and create separate contexts for each - but that seems like it might be overkill.

Another would is to define them as pointers (e.g. `*bool` so if it's `nil` we know it's not set. That'll work but it is more typing for the user and only something we should consider if we think a popular usecase will be for people to only set a subset of the 3 flags. I don't have enough experience with MQTT to know if that's the case or not.

protocol/mqtt_paho/v2/context.go
```diff
@@ -0,0 +1,37 @@
+package mqtt_paho
+
+import "context"
+
+// ProtocolContext allows a Receiver to understand the context of a request.
+type ProtocolContext struct {
+	Topic    string
+	QoS      byte
+	Retained bool
+}
```
skeeey, line 10: would we also consider to support `packet id` and `packet properties`?

https://mosquitto.org/man/mosquitto-8.html

yanmxa: Thanks @skeeey! Append these two parameters.

protocol/mqtt_paho/v2/protocol.go
yanmxa: Thanks @duglin! I think it's a better idea to define them as pointers. This practice also seems to be common in [paho mqtt](https://github.com/eclipse/paho.golang/blob/86185b4d2a6c372afe12535368e2dcf805278911/paho/cp_publish.go#L27-L30). We can also provide [a constructor](https://github.com/cloudevents/sdk-go/pull/924/files#diff-c9bfbe8b1a4e23643b340cf417da6912def99e374fab7bb4d95486f87da8fd8dR18-R25) to reduce user typing for these pointer variables.

protocol/mqtt_paho/v2/context.go
```diff
@@ -0,0 +1,44 @@
...
+// ProtocolContext allows a Receiver to understand the context of a request.
+type ProtocolContext struct {
+	Topic             string
+	QoS               *byte
+	Retained          *bool
+	PacketID          *uint16
+	PublishProperties *paho.PublishProperties
+}
+
+// for more commonly used parameters, provide a constructor
```
embano1, line 18: nit: incorrect doc string

yanmxa: @embano1 How about `provide a constructor to initialize frequently used parameters?`

yanmxa: Done.

embano1 (commented):
Digging more into the implementation, I have a general question @yanmxa on this PR: it seems that we're replicating functionality in the protocol `context` which is already (partially) available in the protocol constructor `New()`. Can you provide an example why someone would use the `context` vs `New()` to configure these settings? I understand that we do this to overwrite behavior for certain, but not all, `Send()` calls in protocols. However, this proposal seems to create inconsistencies between the options available in the constructor `New()` and what can be defined on `context`. Also, I'm afraid we have to keep up with field level changes, i.e., add more fields, with the current proposal, so I was wondering whether there's a generic MQTT options/property struct we should allow (use) to influence all parameters in `Send()` instead of handpicking a couple of options?

samples/mqtt/sender/main.go
```diff
@@ -62,7 +61,7 @@ func main() {
 			log.Printf("failed to set data: %v", err)
 		}
 		if result := c.Send(
-			cecontext.WithTopic(ctx, "test-topic"),
+			cemqtt.WithProtocolContext(ctx, cemqtt.NewProtocolContext("test-topic", 0, false)),
```
embano1, line 64: I can see this list of parameters getting quiet out of hand vs using functional options or a config struct instead.

protocol/mqtt_paho/v2/protocol.go
```diff
@@ -71,16 +71,35 @@ func (t *Protocol) Send(ctx context.Context, m binding.Message, transformers ...
...
+		}
+		if protocolCtx.QoS != nil {
+			msg.QoS = *protocolCtx.QoS
+		}
+		if protocolCtx.PacketID != nil {
+			msg.PacketID = *protocolCtx.PacketID
+		}
+		if protocolCtx.PublishProperties != nil {
+			msg.Properties = protocolCtx.PublishProperties
+		}
```
embano1, line 100: Blocking: these are not available in the `New()` constructor call (inconsistent)

yanmxa:
> 
**1. Why someone would use the `context` vs. `New()` to configure these settings?**

When using MQTT sends a message, you can configure many parameters according to your needs. We provide the `ProtocolConext` here to support users to pass [all the parameters](https://github.com/eclipse/paho.golang/blob/master/paho/cp_publish.go) to MQTT publisher. For example, let's create a `ProtocolContext` and then add it to the ctx.
  ```golang
  qos := 0
  retained := false
  packetId := uint16(123)
  properties := &paho.PublishProperties{
	  TopicAlias: paho.Uint16(3),
  }
  pctx := cemqtt.ProtocolContext{
       Topic: "test",
        QoS : &qos,
        Retained: &retained,
        PacketID: &packetId,
        PublishProperties: properties
  }
  ctx := cemqtt.WithProtocolContext(ctx, pctx)
  ```
  But in most cases, users only specify `topic/qos/retain` when publishing messages, and do not set packet-related parameters, so I provide a `New` method here to deal with such cases
  ```golang
  if result := c.Send(
	  cemqtt.WithProtocolContext(ctx, cemqtt.NewProtocolContext("test-topic", 0, false)),
	  e,
  );
  ```
  
**2. Whether there generic MQTT options/property struct we should allow (use) to influence all parameters in Send() instead of handpicking a couple of options?**

Yes. There is a generic MQTT option struct when sending the message, But in its implementation, both payload and properties are placed [in the struct](https://github.com/eclipse/paho.golang/blob/86185b4d2a6c372afe12535368e2dcf805278911/paho/cp_publish.go#L12-L32). That's the reason I created the `ProtocolContext` to map all the properties, instead of using its own option struct.

protocol/mqtt_paho/v2/protocol.go
yanmxa: Yes. A partial constructor is used to handle most cases, and I can add another constructor with full arguments if you think it is necessary.

samples/mqtt/sender/main.go
yanmxa: To keep the code short, such a `New()` constructor is used here. There is indeed some confusion when the parameter is not identified.

But cause we define several pointers in the `ProtocolContext`, using a config struct will significantly increase the amount of code(typing). 
Using functional options to pass multiple parameters will also increase the number of lines of configuration code.

I'm sure which one is better, and which of them do you suggest?

embano1: We already use functional options and short (w/ default) constructors in other places so that IMHO would remove the need for `ProtocolContext` completely. WDYT?

protocol/mqtt_paho/v2/protocol.go
embano1: See above

embano1:
> **1. Why someone would use the `context` vs. `New()` to configure these settings?**
> 
> When using MQTT sends a message, you can configure many parameters according to your needs. We provide the `ProtocolConext` here to support users to pass [all the parameters](https://github.com/eclipse/paho.golang/blob/master/paho/cp_publish.go) to MQTT publisher. For example, let's create a `ProtocolContext` and then add it to the ctx.
> But in most cases, users only specify `topic/qos/retain` when publishing messages, and do not set packet-related parameters, so I provide a `New` method here to deal with such cases

I don't think this is the right use case for `Context` here. We generally use a specific `Context` to overwrite certain behavior for individual `Send` calls. In your case, I'd suggest using a default `New` constructor with functional options. I'm not 100% clear on whether the `topic` value will always come from `cecontext` or must be passed to `New()`. WDYT?

> **2. Whether there generic MQTT options/property struct we should allow (use) to influence all parameters in Send() instead of handpicking a couple of options?**
> 
> Yes. There is a generic MQTT option struct when sending the message, But in its implementation, both payload and properties are placed [in the struct](https://github.com/eclipse/paho.golang/blob/86185b4d2a6c372afe12535368e2dcf805278911/paho/cp_publish.go#L12-L32). That's the reason I created the `ProtocolContext` to map all the properties, instead of using its own option struct.

I just checked our implementation and IMHO it would be fine to have the `Context` use the whole `struct` assuming a user would know how to use it (since he's intentionally overwriting behavior) **and** would not pass a `payload` (which would be overwritten anyways during `SetData`.

samples/mqtt/sender/main.go
yanmxa: The short constructor is used to initialize the `ProtocolContext`. But I think you have given a good option that we can make use of the [struct option](https://github.com/eclipse/paho.golang/blob/86185b4d2a6c372afe12535368e2dcf805278911/paho/cp_publish.go#L12-L32) directly instead of `ProtocolContext`

yanmxa:
> > **1. Why someone would use the `context` vs. `New()` to configure these settings?**
> > When using MQTT sends a message, you can configure many parameters according to your needs. We provide the `ProtocolConext` here to support users to pass [all the parameters](https://github.com/eclipse/paho.golang/blob/master/paho/cp_publish.go) to MQTT publisher. For example, let's create a `ProtocolContext` and then add it to the ctx.
> > But in most cases, users only specify `topic/qos/retain` when publishing messages, and do not set packet-related parameters, so I provide a `New` method here to deal with such cases
> 
> I don't think this is the right use case for `Context` here. We generally use a specific `Context` to overwrite certain behavior for individual `Send` calls. In your case, I'd suggest using a default `New` constructor with functional options. I'm not 100% clear on whether the `topic` value will always come from `cecontext` or must be passed to `New()`. WDYT?
> 
> > **2. Whether there generic MQTT options/property struct we should allow (use) to influence all parameters in Send() instead of handpicking a couple of options?**
> > Yes. There is a generic MQTT option struct when sending the message, But in its implementation, both payload and properties are placed [in the struct](https://github.com/eclipse/paho.golang/blob/86185b4d2a6c372afe12535368e2dcf805278911/paho/cp_publish.go#L12-L32). That's the reason I created the `ProtocolContext` to map all the properties, instead of using its own option struct.
> 
> I just checked our implementation and IMHO it would be fine to have the `Context` use the whole `struct` assuming a user would know how to use it (since he's intentionally overwriting behavior) **and** would not pass a `payload` (which would anyways be overwritten during `SetData`).

Agree with that. In order to be consistent, we can keep using function options. At the same time, discard `protocolContext` and use the struct option as the value.

yanmxa:
@embano1 If I append the mqtt function option in the file `v1/context/context.go`, then we have to update so many `go.mod` and `go.sum` files just like this PR. Another way is that start a new PR and add this MQTT function option to the directory `protocol/mqtt_phao/v2`, this only requires updating a few files. 
So which one do you think is better?

embano1:
Sorry, I think I may have confused you somehow. I did not suggest using cecontext for this, but to implement everything in func New() of the protocol API e.g., using functional options and defaults. See http protocol implementation for how this is done there.

embano1:
@yanmxa looks like you reverted to the mqtt context approach instead of what I meant following other protocol implementations e.g., http. Can you please review https://github.com/cloudevents/sdk-go/blob/3dfc0335ba8a9ed559fda3d32ea1b2a18b17f5da/v2/protocol/http/protocol.go#L95 and see if this approach would work for your proposal? I know it's a breaking change but we haven't cut a release with mqtt anyways so strictly it's non-breaking.

yanmxa:
> 

Yes, I think we can support setting MQTT options in both the`New()` protocol and `Send()`(just like the current change, cause the settings sent each time may be different from the settings at initialization) at the same time. What do you think?

embano1:
> > 
> 
> 
> 
> Thanks @embano1! I misunderstood before. 
> 
> Also, I think we can support setting the options in both the`New()` protocol and `Send()`(cause the settings sent each time may be different from the ones at initialization) at the same time. What do you think?

Send is an interface implementation. For those cases context is typically used. Can we start with New() in this PR and create a follow up PR if needed?

embano1:
@yanmxa what do you think of using functional options in Mew() like other protocols do to reduce API surface, allow for API evolution, set defaults, etc.?

embano1:
Great! Excited!

samples/mqtt/sender/main.go
```diff
@@ -29,18 +29,17 @@ func main() {
...
 		KeepAlive:  30,
 		CleanStart: true,
 	}
 	// set a default topic with test-topic1
-	p, err := mqtt_paho.New(ctx, clientConfig, connConfig, &paho.Publish{
-		Topic: "test-topic1", QoS: 0,
-	}, nil)
+	p, err := mqtt_paho.New(config, mqtt_paho.WithPublish(&paho.Publish{Topic: "test-topic1"}), mqtt_paho.WithConnect(connOpt))
 	if err != nil {
 		log.Fatalf("failed to create protocol: %s", err.Error())
```
embano1, line 44: nit: you can use `%v` and just `err` to reduce typing :)

protocol/mqtt_paho/v2/protocol.go
```diff
@@ -39,35 +41,47 @@ var (
...
 }
 
-func (t *Protocol) Send(ctx context.Context, m binding.Message, transformers ...binding.Transformer) error {
-	var err error
+func (p *Protocol) Send(ctx context.Context, m binding.Message, transformers ...binding.Transformer) error {
+	if p.publishOption == nil {
+		return fmt.Errorf("the paho.Publish option should not be nil")
+	}
+
+	err := p.initClient(ctx)
```
embano1, line 77: question: this is called in `Send` and `OpenInbound` which might be confusing/inconsistent across usage. Should `initClient` be part of `New()`?

protocol/mqtt_paho/v2/protocol.go
```diff
@@ -78,41 +92,50 @@ func (t *Protocol) Send(ctx context.Context, m binding.Message, transformers ...
...
 
-	t.client.Router = paho.NewSingleHandlerRouter(func(m *paho.Publish) {
-		t.incoming <- m
+	p.client.Router = paho.NewSingleHandlerRouter(func(m *paho.Publish) {
+		p.incoming <- m
 	})
 
-	logger.Infof("subscribe to topics: %v", t.subscribeOption.Subscriptions)
-	_, err := t.client.Subscribe(ctx, t.subscribeOption)
+	logger.Infof("subscribe to topics: %v", p.subscribeOption.Subscriptions)
```
embano1, line 121:
```suggestion
	logger.Infof("subscribing to topics: %v", p.subscribeOption.Subscriptions)
```

protocol/mqtt_paho/v2/protocol.go
```diff
@@ -41,96 +41,101 @@ var (
 	_ protocol.Closer   = (*Protocol)(nil)
 )
 
-func New(ctx context.Context, clientConfig *paho.ClientConfig, connConfig *paho.Connect, SenderTopic string,
-	ReceiverTopics []string, qos byte, retained bool,
-) (*Protocol, error) {
-	client := paho.NewClient(*clientConfig)
-	ca, err := client.Connect(ctx, connConfig)
-	if err != nil {
+func New(config *paho.ClientConfig, opts ...Option) (*Protocol, error) {
```
embano1, line 44: questions:

- should we keep `ctx` as first arg?
- can we supply defaults to `config` so it can also become an option? I see it's mostly used to set a client `id` and `conn` which can be defaulted IMHO

embano1 (changes requested):

protocol/mqtt_paho/v2/protocol.go
yanmxa: That's okay. Just if we put `initClient` on `New()`, then the `ctx` must be added as the parameter of the `New()` constructor.

embano1: I don't think this is an issue and makes the protocol even easier to use/harder to misuse

protocol/mqtt_paho/v2/protocol.go
yanmxa: - We can keep the `ctx` as the first args. So I will init the client on the `New()`. 
- Of course, we can set the `id` by default. but how can we set a default `conn` from the `config`?  If we don't specify the address of the conn, it seems there is no way to establish a connection. and throw the error message: `client connection is nil`

protocol/mqtt_paho/v2/option.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// Option is the function signature required to be considered an mqtt_paho.Option.
+type Option func(*Protocol) error
+
+// WithConnect sets an paho.Connect for the client. This option is not required.
+func WithConnect(connOpt *paho.Connect) Option {
+	return func(p *Protocol) error {
+		if connOpt == nil {
+			return fmt.Errorf("the paho.Connect option should not be nil")
```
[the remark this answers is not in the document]
yanmxa: Done.

protocol/mqtt_paho/v2/protocol.go
```diff
@@ -39,35 +41,47 @@ var (
...
+	p := &Protocol{
+		config: config,
+		// default connect option
+		connOption: &paho.Connect{
+			KeepAlive:  30,
+			CleanStart: true,
+		},
+		incoming:    make(chan *paho.Publish),
+		openerMutex: sync.Mutex{},
+		clientMutex: sync.Mutex{},
```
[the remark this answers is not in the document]
yanmxa: Done.

samples/mqtt/sender/main.go
yanmxa: Done.

embano1 (commented):
Great work! Can you please squash the commits?

embano1:
Are there any doc-related changes needed for this?

yanmxa:
> Great work! Can you please squash the commits?

Done!

yanmxa:
> Are there any doc-related changes needed for this?

The protocol here has not changed, and it seems that there is no need to change the document now.

embano1 (approved):
GREAT WORK!

LGTM

PR #932, opened 2023-08-16
update ci to use go matrix to test using 1.17/18/19/20/21

cpanato:
- update ci to use go matrix to test using 1.17/18/19/20/21

if want we can go straight to 1.21
also we might want to bump the go modules as well to set the go version to maybe 1.19

this will require an update in the github status check


embano1:
@cpanato quick question, is there a reason for this CI/mod change? Asking bc we haven't yet discussed up to which Go version we want to test/"support". README IIRC still says 1.17. 

cc/ @lionelvillard

cpanato:
@embano1 from my point of view go 1.17 is deprecated and we should test/build this project with latest go releases.
We can add a go matrix and test it with go 1.17/18/19/20.... if that is what the maintainers want, happy to help here.

otherwise we should aim to test/build with latest go version.

will wait feedback to change the code
thanks!

embano1:
+1 on the matrix since we currently state in the README

> Note: Supported go version: 1.17+

Regarding bumping `go 1.17` I'm not sure about the benefits here? Can you elaborate more?

cpanato:
> +1 on the matrix since we currently state in the README
> 
> > Note: Supported go version: 1.17+
> 
> Regarding bumping `go 1.17` I'm not sure about the benefits here? Can you elaborate more?

That one we can leave, but that tells what the minimum go version is to use the lib, but also, the newer go version has better management of the go modules, so it is good to bring that up to a newer go release.
but again it is a decision from the maintainers

embano1:
> That one we can leave, but that tells what the minimum go version is to use the lib, but also, the newer go version has better management of the go modules

Yes, main concern from my end is that this will encourage usage of Go 1.18 APIs/features (Generics for example). This is not a bad thing per se, but requires users to also be on that Go version to compile. Hence my caution.

Feel free to start with the Go version matrix in CI.

cpanato:
closing this looks like it is done in https://github.com/cloudevents/sdk-go/pull/958

PR #935, opened 2023-09-13
Clean up the previous user properties from MQTT protocol  

yanmxa:
Since we use the MQTT `Properties.User` to load the extension of the cloudevents. It can't contain the previous extension.


yanmxa:
@embano1 /cc

embano1 (approved):
LGTM

PR #936, opened 2023-09-13
context.Done() may never reach if waiting on r.incoming <- msgErr

nbajaj90:
While analysing an issue, we observed that even after context is cancelled, the code was blocked in ConsumeClaim. And the reason is if the code if blocked at r.incoming <- msgErr, it may fail to honour or may skip the outside session.Context().Done().
Reference - https://stackoverflow.com/questions/67608032/context-ctx-done-not-being-executed-even-though-context-was-passed-to-the-functi

Solution is to add session.Context().Done() with  r.incoming <- msgErr in select clause.

### DIFF
--- a/protocol/kafka_sarama/v2/receiver.go
+++ b/protocol/kafka_sarama/v2/receiver.go
@@ -66,13 +66,21 @@ func (r *Receiver) ConsumeClaim(session sarama.ConsumerGroupSession, claim saram
 				return nil
 			}
 			m := NewMessageFromConsumerMessage(msg)
-
-			r.incoming <- msgErr{
+			msgErrObj := msgErr{
 				msg: binding.WithFinish(m, func(err error) {
 					if protocol.IsACK(err) {
 						session.MarkMessage(msg, "")
 					}
 				}),
+			}
+
+			// Need to use select clause here, otherwise r.incoming <- msgErrObj can become a blocking operation,
+			// resulting in never reaching outside block's case <-session.Context().Done()
+			select {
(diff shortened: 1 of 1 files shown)


protocol/kafka_sarama/v2/receiver.go
```diff
@@ -66,15 +66,23 @@ func (r *Receiver) ConsumeClaim(session sarama.ConsumerGroupSession, claim saram
...
 					}
 				}),
 			}
 
+			// Need to use select clause here, otherwise r.incoming <- msgErrObj can become a blocking operation,
+			// resulting in never reaching outside block's case <-session.Context().Done()
+			select {
+			case r.incoming <- msgErrObj:
+				// do nothing
+			case <-session.Context().Done():
```
embano1, line 82: spot on! usually we'd return `session.Context.Err()` to let the caller know why the function returned. But this would break the existing behavior (returning `nil`). A user could always use `context.Cause(ctx)`, but that might not be intuitive from the `nil` behavior.

Anyways, IMHO with this change you can remove the code in line 95?

nbajaj90: Code on line 95 is still required. If code on line no. 64 `case msg, ok := <-claim.Messages():` doesn't get hit, that code on line no. 95 would exit the flow.

embano1: True, was under the assumption that `ok` would be false due to propagated channel somewhere into Sarama...but seems that's not the case.

PR #958, opened 2023-10-18
Move to go 1.18.

duglin:
Upgrade to go v1.18
Had to run gofmt on a few files too.
Added more `time.Sleep` calls to try to fix the race condition, again.


.github/workflows/conformance.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     name: CloudEvents
     strategy:
       matrix:
-        go-version: [1.17.x]
+        go-version: [1.18.x]
```
embano1, line 15: nit: you can use the latest available go version in the workflows

duglin: ok I changed them all to use "stable".

duglin: I think one downside to this is that the first person to do a PR after a new release of golang might be forced to do things like run "gofmt" on everything if the format rules change. Just ran into that. But let's see how it goes...

embano1: Yeah, not a big fan of these abstract version identifiers, especially when it comes to troubleshooting and you don't see immediately which version we are using in CI. Feel free to revert to 1.21 for now.

duglin: ok back to 1.21

duglin:
with the race condition behind us, this one is now ready for final review and merge.
I had to add a few more `time.Sleep` calls to avoid another race condition. I'm not happy about it but it seems to help.... at least for now.
@embano1 @lionelvillard

v2/types/value.go
```diff
@@ -86,7 +86,8 @@ func Format(v interface{}) (string, error) {
 }
 
 // Validate v is a valid CloudEvents attribute value, convert it to one of:
-//     bool, int32, string, []byte, types.URI, types.URIRef, types.Timestamp
+//
```
embano1, line 89: question: is this newline intentional?

test/integration/kafka_sarama_binding/kafka_test.go
```diff
@@ -104,7 +104,7 @@ func testSenderReceiver(t testing.TB) (func(), bindings.Sender, bindings.Receive
 
 	// Not perfect but we need to give OpenInbound() as chance to start
 	// as it's a race condition. I couldn't find something on 'p' to wait for
-	time.Sleep(5 * time.Second)
+	time.Sleep(15 * time.Second)
```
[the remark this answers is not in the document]
duglin: once i added the other sleeps I saw this race condition appear again. So I increased this one to be more than the other 2 sleep values combined. Not 100% sure if I'm just getting lucky or not but my local testing seems happier now

v2/client/test/test.go
```diff
@@ -29,6 +30,9 @@ func SendReceive(t *testing.T, protocolFactory func() interface{}, in event.Even
 	wg := sync.WaitGroup{}
 	wg.Add(2)
 
+	// Give time for Kafka client protocol to get setup
+	time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
duglin: Because I had this PR open :-) 
I can create a separate PR but in the end does it really matter as long as it all gets merged?

v2/protocol/http/context.go
```diff
@@ -24,7 +24,8 @@ type RequestData struct {
 }
 
 // WithRequestDataAtContext uses the http.Request to add RequestData
-//  information to the Context.
+//
+//	information to the Context.
```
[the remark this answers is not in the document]
duglin: same here

embano1 (approved):
LGTM

embano1 (approved):

Issue #964, opened 2023-10-20
Pass message to RecordReceivedMalformedEvent when validation failed


raza-matillion:
We have encountered a scenario where we want to send all messages (that failed validation) to dead letter topic to be reviewed later. Currently I can't find a way to get the message if it fails validation. I think it will be useful if we pass message
Request to pass the `binding.Message` to ObservabilityService method RecordReceivedMalformedEvent which will enable us to log or send message to dead letter topic.

[https://github.com/cloudevents/sdk-go/blob/70abff67deec3c82098aca9beb391ee4d8ae0c67/v2/client/invoker.go#L68](url)
and 
https://github.com/cloudevents/sdk-go/blob/70abff67deec3c82098aca9beb391ee4d8ae0c67/v2/client/invoker.go#L74

embano1:
Quick question: since `invoker.go` returns an error in such cases, why not send that event to a configured DLQ (this SDK is not opinionated on DLQs, so you can use whatever SDK/logic you want in your app). I have not seen many cases where the observability service is responsible for DLQ-ing, but I could be wrong here.

cc/ @duglin for thoughts

duglin:
I like the idea of a DLQ that isn't tied to the ObservabilityService.

raza-matillion:
I agree with the idea. My plan is to use custom error handler in ObservabilityService that will handle the error as we intended. So the DLQ logic will not live in ObservabilityService. 

`
func (o Observability) RecordReceivedMalformedEvent(ctx context.Context, m binding.Message, err error) {
	o.ErrorHandler.Handle(ctx, m, err)
}
`

What I understand is, the `invoker.go` does not return error when we use `client.StartReceive`. Correct me if I am wrong.

raza-matillion:
The message is also passed as `nil` to the `respFn` otherwise I could use `client.Responder` instead of `client.StartReceive`

https://github.com/cloudevents/sdk-go/blob/70abff67deec3c82098aca9beb391ee4d8ae0c67/v2/client/invoker.go#L75C24-L75C24

embano1:
So you don't know what the `CloudEvent` is upfront? I was assuming you know the to be sent `CloudEvent` and together with the validation error returned you can construct a DLQ message? Can you describe a common flow of how you/your users are using the SDK?

raza-matillion:
[...]

PR #965, opened 2023-10-20
Passing message to record malformed event

raza-matillion:
Issue: https://github.com/cloudevents/sdk-go/issues/964


duglin:
still testing, but I think the race condition will be fixed by https://github.com/cloudevents/sdk-go/pull/958

duglin:
please rebase

raza-matillion:
@duglin Rebase done!

raza-matillion:
@duglin Resolved conflicts and updated PR. Sneaked one more tiny change in `invoker.go`.

raza-matillion:
@duglin test seems flakey. Its passing locally on my machine so probably need a rerun.

duglin:
rebase again and let's see how it goes...

raza-matillion:
@duglin done!

duglin:
I'm hoping https://github.com/cloudevents/sdk-go/pull/970 fixes the errors you're seeing....

duglin:
I'm confused as to why this is associated with the Observability service

raza-matillion:
@duglin For logging purpose. Basically I want to get malformed message and log it. If there is better approach i.e. if we can pass error handler as an extra client option, I am ok with that. At the moment, there is no way I can get hold of malformed event so that I can investigate it further.

duglin:
I shouldn't admit this, but I'm not as up to speed on the SDK as I should be :-)  but I just can't help but wonder why we're not doing something like extending the `receiveInvoker` struct with some kind of DLQ/handler/fn/thingy. And then also send events to it if the receiverFn returns an error, maybe?

@embano1 @lionelvillard any thoughts on whether extending the Obs. service is the best approach?

embano1:
Yeah, we should think about how we want to give users an option to consume events/requests throughout the various stages of event handling logic in the SDK. We can log, use callbacks, change signatures, or pass objects instead of nil. So we should discuss this a bit more. Perhaps we also need to distinguish for the implementation/PR(s) between error and success cases.

duglin:
Doing some digging/learning and I have a few questions for @embano1 and @lionelvillard ...

It appears that people are supposed to setup a ceClient and then call StartReceiver() with the call-back Fn. In StartReceiver() it'll setup a receiveInvoker (https://github.com/cloudevents/sdk-go/blob/main/v2/client/client.go#L206) with all of the various options for processing the incoming event. However, I'm wondering why StarrtReceiver() takes the call-back Fn as a parameter instead of it being one of the ceClient options? Is this because there's an assumption that there can be more than one StartReceiver() running at the same time and each might need a different event processor? This line: https://github.com/cloudevents/sdk-go/blob/main/v2/client/client.go#L202 would seem to indicate "no", it's a singleton.

It's probably not a big deal, I'm just wondering why this one option (the call-back Fn) is special from all of the other ceClient options (like Decorators) and not setup the same way? And, this popped in my head because I was trying to think about how we might want to setup a DLQ - to me a DLQ isn't much different from the call-back Fn so for consistency I thought we should follow the same pattern. But, I'm not sure I'd want to extend the StartReceiver() signature with another (optional) arg.

duglin:
Or maybe it's special because it's the only required param? I could buy that.

So, either way, what do you guys think about adding a DQL field in receiveInvoker?

duglin:
@embano1 @lionelvillard since this one has lingered for so long w/o anyone really complaining or asking for this... any objection to closing it?  @raza-matillion do you think this is still needed?

embano1:
Ok to close for me

duglin:
Closing due to lack of interest. We can reopen if someone wants to drive it.
@raza-matillion if you want to see this one added please see the above comments.

Issue #831, opened 2023-01-19
No logs when nacking PubSub msgs


FSMatias:
The PubSub's Finish func is swallowing the error making it hard to identify issues.

In our case, we noticed that the number of unack messages in the subscription was increasing, however there were no warning or error logs to indicate why. After investigating the issue, we found out that some of the messages sent to the PubSub topic did not have all the required attributes, leading to an error while converting the message to an event. However, because the Finish func always returns nil, there were no logs in our system.

I am opening this issue to suggest a change on the [Finish func](https://github.com/cloudevents/sdk-go/blob/1f844bade36c2da5d36a085daa42d05991e898bb/protocol/pubsub/v2/message.go#L125) in the PubSub protocol so it could return the original error, instead of always retuning nil. By doing that, the error would be properly [logged](https://github.com/cloudevents/sdk-go/blob/1f844bade36c2da5d36a085daa42d05991e898bb/v2/client/client.go#L254) making it easier to identify issues. 

Below you can see the proposed change. [Here](https://github.com/cloudevents/sdk-go/blob/1f844bade36c2da5d36a085daa42d05991e898bb/protocol/pubsub/v2/message.go#L125) is the current code.

```
func (m *Message) Finish(err error) error {
	if err != nil {
		m.internal.Nack()
		return err
	} 

	m.internal.Ack()
	return nil
}
```

PR #968, opened 2023-10-24
returning the error when doing a nack in the message

cpanato:
Fixes: #831 

other protocol return the error, doing similar thing here as well

### DIFF
--- a/protocol/pubsub/v2/message_test.go
+++ b/protocol/pubsub/v2/message_test.go
@@ -7,6 +7,7 @@ package pubsub
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"cloud.google.com/go/pubsub"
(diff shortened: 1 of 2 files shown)


duglin:
Seems reasonable. Is there a test we can add for this?

duglin:
rebase is needed

cpanato:
done @duglin, thanks for the review, PTAL

duglin:
Is it possible to test that the error makes its way back to the logger?

cpanato:
> Is it possible to test that the error makes its way back to the logger?
did some manual test and saw that showing up in the logs.

but not sure how to do that here in the tests, do you have a tests that setup the pubsub? sorry for asking that, i am too new to this codebase

duglin:
I looked around and I didn't see a good example of us testing that yet so perhaps I'm asking too much.
Unless @embano1 can think of an easy way to test it, this
/LGTM

duglin (dismissed):

embano1:
> Is it possible to test that the error makes its way back to the logger?

Are you asking for the caller here, i.e., integration tests?

duglin:
> > Is it possible to test that the error makes its way back to the logger?
> 
> Are you asking for the caller here, i.e., integration tests?

Yea, I guess an integration test since the testcase added (while valid) is kind of silly/trivial since it's basically just testing one if-statement, and the real point of this PR was to ensure that the error bubbled up to the proper spot (like the logger). So, in essence not just test the if-statement but also the Finish() logic in conjunction with our logging system. But if we don't have an existing test that can be easily hooked-into I'm not sure it's fair to ask @cpanato to add one just for this trivial PR.

cpanato:
maybe we can open an issue for that, I think that will require some more complex things to setup

duglin:
> maybe we can open an issue for that, I think that will require some more complex things to setup

Done: https://github.com/cloudevents/sdk-go/issues/979

I'm still ok with this PR. @embano1 for the final review/merge.

protocol/pubsub/v2/message_test.go
```diff
@@ -45,3 +46,39 @@ func TestReadStructured(t *testing.T) {
...
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			msg := NewMessage(tc.pm)
+			err := msg.Finish(tc.err)
+			if tc.wantErr {
+				if err != tc.err {
+					t.Errorf("Error mismatch. got: %v, want: %v", err, tc.err)
+				}
+			}
```
embano1, line 81: nit: logic is partially broken IMHO if `tc.wantErr` is `false` but`err != nil`, i.e., the second test case does not seem to have any effect here?

Usually you'd do something like

```suggestion
			if tc.wantErr && err != nil {
				if err != tc.err {
					t.Errorf("Error mismatch. got: %v, want: %v", err, tc.err)
				}
			} else { // assert `tc.wantErr` is `false` and `err == nil` }
```

cpanato: yep, makes sense

cpanato: done, thanks for the review

embano1 (approved):
LGTM

PR #971, opened 2023-10-25
Try to fix race again - don't reuse clients for sender/receiver

duglin:
from: https://pkg.go.dev/github.com/shopify/sarama#NewConsumerGroupFromClient
- - -
NewConsumerGroupFromClient creates a new consumer group using the given client. It is still necessary to call Close() on the underlying client when shutting down this consumer. PLEASE NOTE: consumer groups can only re-use but not share clients. 
- - -
I'm not 100% sure if using the same client for sending and receiving qualifies for this restriction or not but let's see if this works w/o the "sleeps". Seems to work locally for me.

So, this PR just creates 2 clients (one for sending and one for receiving) instead of trying to reuse the same one for both.

### DIFF
--- a/test/integration/kafka_sarama_binding/kafka_test.go
+++ b/test/integration/kafka_sarama_binding/kafka_test.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/IBM/sarama"
 	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


duglin:
@embano1 I ran the tests several times via the "checks" and they seem happy. I did see some weirdness in the mqtt stuff but we can look at that separately - haven't seen a kafka complaint yet. Thoughts?

test/integration/kafka_sarama_binding/kafka_test.go
```diff
@@ -91,25 +90,28 @@ func testClient(t testing.TB) sarama.Client {
 }
 
 func testSenderReceiver(t testing.TB) (func(), bindings.Sender, bindings.Receiver, string) {
-	client := testClient(t)
-
 	topicName := "test-ce-client-" + uuid.New().String()
-	p, err := kafka_sarama.NewProtocolFromClient(client, topicName, topicName, kafka_sarama.WithReceiverGroupId(TEST_GROUP_ID))
+
+	clientR := testClient(t)
```
embano1, line 95: nit: please add a code comment or better variable name to indicate what the `R` and `S` suffixes on `client` and `p` mean so our future-selves have an easier time understanding the change :)

embano1 (commented):
Minor nit.

And IMHO this should fix the races as my understanding from prior code digging was the race due to sharing the same Kafka client in the test racing against the broker with concurrent function calls during producer/consumer initialization.

test/integration/kafka_sarama_binding/kafka_test.go
duglin: Done - I can barely remember why/what I did yesterday so this is a good "nit" :-)

embano1 (approved):

PR #976, opened 2023-10-27
 Add WithCustomAttributes for PubSub

AndreasBergmeier6176:
Enables to pre-fill PubSub Message Attributes independent of CloudEvent spec.
Thus this allows for setting arbitrary Message Attributes (given that they are not later overwritten by CloudEvent Attributes).

### DIFF
--- a/protocol/pubsub/v2/attributes.go
+++ b/protocol/pubsub/v2/attributes.go
@@ -1,0 +1,25 @@
+/*
+ Copyright 2021 The CloudEvents Authors
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+package pubsub
+
+import (
+	"context"
+
+	"github.com/cloudevents/sdk-go/v2/binding"
+)
+
+type withCustomAttributes struct{}
+
+func AttributesFrom(ctx context.Context) map[string]string {
+	return binding.GetOrDefaultFromCtx(ctx, withCustomAttributes{}, make(map[string]string)).(map[string]string)
+}
+
+// WithCustomAttributes sets Message Attributes without any CloudEvent logic.
+// Note that this function is not intended for CloudEvent Extensions or any `ce-`-prefixed Attributes.
+// For these please see `Event` and `Event.SetExtension`.
+func WithCustomAttributes(ctx context.Context, attrs map[string]string) context.Context {
+	return context.WithValue(ctx, withCustomAttributes{}, attrs)
+}
--- a/protocol/pubsub/v2/protocol.go
+++ b/protocol/pubsub/v2/protocol.go
@@ -110,7 +110,9 @@ func (t *Protocol) Send(ctx context.Context, in binding.Message, transformers ..
 
 	conn := t.getOrCreateConnection(ctx, topic, "", "")
 
-	msg := &pubsub.Message{}
+	msg := &pubsub.Message{
+		Attributes: AttributesFrom(ctx),
+	}
 
 	if key, ok := ctx.Value(withOrderingKey{}).(string); ok {
 		if !t.MessageOrdering {
--- a/protocol/pubsub/v2/protocol_test.go
+++ b/protocol/pubsub/v2/protocol_test.go
@@ -3,14 +3,17 @@ package pubsub
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"testing"
 
 	"cloud.google.com/go/pubsub"
+	"cloud.google.com/go/pubsub/apiv1/pubsubpb"
 	"cloud.google.com/go/pubsub/pstest"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/api/option"
 	pb "google.golang.org/genproto/googleapis/pubsub/v1"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
 
 	"github.com/cloudevents/sdk-go/v2/test"
 )
@@ -18,6 +21,32 @@ import (
 type testPubsubClient struct {
 	srv  *pstest.Server
 	conn *grpc.ClientConn
+}
+
+func (pc *testPubsubClient) NewWithAttributesInterceptor(ctx context.Context, projectID, orderingKey string) (*pubsub.Client, error) {
+	pc.srv = pstest.NewServer()
+	conn, err := grpc.Dial(pc.srv.Addr, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithUnaryInterceptor(customAttributesInterceptor(map[string]string{
+		"Content-Type":        "text/json",
+		"ce-dataschema":       "http://example.com/schema",
+		"ce-exbinary":         "AAECAw==",
+		"ce-exbool":           "true",
+		"ce-exint":            "42",
+		"ce-exstring":         "exstring",
+		"ce-extime":           "2020-03-21T12:34:56.78Z",
+		"ce-exurl":            "http://example.com/source",
+		"ce-id":               "full-event",
+		"ce-source":           "http://example.com/source",
+		"ce-specversion":      "1.0",
+		"ce-subject":          "topic",
+		"ce-time":             "2020-03-21T12:34:56.78Z",
+		"ce-type":             "com.example.FullEvent",
+		"Proxy-Authorization": "YWxhZGRpbjpvcGVuc2VzYW1l",
+	})))
+	if err != nil {
+		return nil, err
+	}
+	pc.conn = conn
+	return pubsub.NewClient(ctx, projectID, option.WithGRPCConn(conn))
 }
 
 func (pc *testPubsubClient) NewWithOrderInterceptor(ctx context.Context, projectID, orderingKey string) (*pubsub.Client, error) {
@@ -35,6 +64,20 @@ func (pc *testPubsubClient) Close() {
 	pc.conn.Close()
 }
 
+func customAttributesInterceptor(attrs map[string]string) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		if method == "/google.pubsub.v1.Publisher/Publish" {
+			pr, _ := req.(*pubsubpb.PublishRequest)
+			for _, m := range pr.Messages {
+				if !reflect.DeepEqual(m.Attributes, attrs) {
+					return fmt.Errorf("expecting Attributes %q, got %q", attrs, m.Attributes)
+				}
+			}
+		}
+		return invoker(ctx, method, req, reply, cc, opts...)
+	}
+}
+
 func orderingKeyInterceptor(orderingKey string) grpc.UnaryClientInterceptor {
 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 		if method == "/google.pubsub.v1.Publisher/Publish" {
@@ -47,6 +90,32 @@ func orderingKeyInterceptor(orderingKey string) grpc.UnaryClientInterceptor {
 		}
 		return invoker(ctx, method, req, reply, cc, opts...)
 	}
+}
+
+func TestPublishMessageHasCustomAttributes(t *testing.T) {
+	require := require.New(t)
+	ctx := context.Background()
+	pc := &testPubsubClient{}
+	defer pc.Close()
+
+	projectID, topicID, orderingKey := "test-project", "test-topic", "foobar"
+
+	client, err := pc.NewWithAttributesInterceptor(ctx, projectID, orderingKey)
+	require.NoError(err, "create pubsub client")
+	defer client.Close()
+
+	prot, err := New(ctx,
+		WithClient(client),
+		WithProjectID(projectID),
+		WithTopicID(topicID),
+		AllowCreateTopic(true),
+	)
+	require.NoError(err, "create protocol")
+
+	err = prot.Send(WithCustomAttributes(ctx, map[string]string{
+		"Proxy-Authorization": "YWxhZGRpbjpvcGVuc2VzYW1l",
+	}), test.FullMessage())
+	require.NoError(err)
 }
 
 func TestPublishMessageHasOrderingKey(t *testing.T) {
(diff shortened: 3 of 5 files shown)


protocol/pubsub/v2/protocol.go
```diff
@@ -233,8 +233,14 @@ var _ protocol.Sender = (*Protocol)(nil)
 var _ protocol.Receiver = (*Protocol)(nil)
 var _ protocol.Closer = (*Protocol)(nil)
 
+type withInitialMessageAttributes struct{}
 type withOrderingKey struct{}
 
+// WithMessageAttributes sets the initial Message Attributes for publishing events.
```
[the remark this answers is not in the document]
embano1: There's also no way to overwrite/delete them with `SetAttribute` since that assumes the `ce-` prefix.

protocol/pubsub/v2/protocol.go
```diff
@@ -233,8 +233,14 @@ var _ protocol.Sender = (*Protocol)(nil)
 var _ protocol.Receiver = (*Protocol)(nil)
 var _ protocol.Closer = (*Protocol)(nil)
 
+type withInitialMessageAttributes struct{}
```
duglin, line 236: More of a golang question... what's the advantage of making the context value's key a "struct" instead of just a "string" with a value of `initialMessageAttributes` ?

protocol/pubsub/v2/protocol.go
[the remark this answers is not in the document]
duglin: > There's also no way to overwrite/delete them with `SetAttribute` since that assumes the `ce-` prefix.

I assume you're talking about https://github.com/cloudevents/sdk-go/blob/24f0eb47c891bbc5955ed1dd3bd2f4582f6270af/protocol/pubsub/v2/write_pubsub_message.go#L70 right?  I'm probably missing it, but how would someone add a non-CE attribute in general ?

duglin:
rebase needed

protocol/pubsub/v2/protocol.go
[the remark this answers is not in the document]
embano1: Oops, spot on @duglin ! I thought those were for "arbitrary" attributes (not just CE) sorry 🤦 
So, my question would rather be related to extension attributes here because currently we hardcode-prefix them with `ce-` which is not strictly required as per the spec - making it impossible to provide custom extension attributes as discussed in this PR (hence the approach taken in this PR).

protocol/pubsub/v2/protocol.go
embano1: > More of a golang question... what's the advantage of making the context value's key a "struct" instead of just a "string" with a value of `initialMessageAttributes` ?

See: https://pkg.go.dev/context#WithValue

> The provided key must be comparable and should not be of type string or any other built-in type to avoid collisions between packages using context. Users of WithValue should define their own types for keys. To avoid allocating when assigning to an interface{}, context keys often have concrete type struct{}. Alternatively, exported context key variables' static type should be a pointer or interface.

duglin: Ah, thanks. That's interesting, sort of a "namespacing" thingy.

protocol/pubsub/v2/protocol.go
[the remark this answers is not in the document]
duglin: Well, I'm a bit confused... my normal state these days :-)  but when I look at the test in this PR it appears to be adding non-CE attributes like `Proxy-Authorization`, which seems like a reasonable thing to do. However, w/o this PR how would someone add that attribute in general? I'm trying to make sure we're consistent between how we allow people to define "initial attributes" and "non-initial attributes" for both CE and non-CE attributes. If we only allow people to set non-CE attributes as "initial" ones, that would feel a bit odd to me.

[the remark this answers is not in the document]
embano1: Rewording my question: why would we ever prefix (custom) extension attributes with `ce-`?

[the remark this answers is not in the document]
duglin: The CE spec doesn't differentiate between spec-defined and extension attributes w.r.t. how they are serialized:
```
Extension attributes are always serialized according to binding rules like standard attributes.
```
So all CE context attributes are prefixed with `ce-` if that's what the protocol bindings defines.

protocol/pubsub/v2/protocol_test.go
```diff
@@ -20,6 +23,32 @@ type testPubsubClient struct {
 	conn *grpc.ClientConn
 }
 
+func (pc *testPubsubClient) NewWithAttributesInterceptor(ctx context.Context, projectID, orderingKey string) (*pubsub.Client, error) {
```
duglin, line 26: I'm confused by the design of the PubsubClient... this PR defines `NewWithAttributesInterceptor` and below I see `NewWithOrderInterceptor`, does this really mean a user can't have Attributes and Order enabled for the same client? I must be missing something.

embano1: That's a test file.

protocol/pubsub/v2/protocol.go
embano1, line 236: Hm, something that worries me is that we're adding a new context attribute, but it's taking effect in a separate part of that protocol.

If you see where/how `withOrderingKey` is used [here](https://github.com/cloudevents/sdk-go/blob/af1a06c50c614e4455e902b5e05d562bfc5fabd1/protocol/pubsub/v2/protocol.go#L115-L120), I'm wondering if there's some opportunity for us to clean up/consolidate **or** move that attributes logic to the file where it's actually used? Having a hard time keeping track of the implementation of this protocol tbh (not to blame you btw!)

protocol/pubsub/v2/protocol.go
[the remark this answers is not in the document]
embano1: > Well, I'm a bit confused... my normal state these days :-) but when I look at the test in this PR it appears to be adding non-CE attributes like `Proxy-Authorization`, which seems like a reasonable thing to do. However, w/o this PR how would someone add that attribute in general? I'm trying to make sure we're consistent between how we allow people to define "initial attributes" and "non-initial attributes" for both CE and non-CE attributes. If we only allow people to set non-CE attributes as "initial" ones, that would feel a bit odd to me.

@AndreasBergmeier6176 thoughts?

protocol/pubsub/v2/protocol_test.go
duglin: doi! sorry

duglin:
(from issue)
> I think from a clarity sense I would prefer to always know the serialization beforehand. To this end it might make sense to also introduce different types, say:
> 
>     * `Attributes`, current type which is specific to CloudEvents - thus in the PubSub case get a `ce-` prefix
> 
>     * `ExternalAttributes`, new type which are attributes not covered by the CloudEvent spec - thus in the PubSub case get no implicit prefix.

Using HTTP as an example, I see this `Option`: https://github.com/cloudevents/sdk-go/blob/af1a06c50c614e4455e902b5e05d562bfc5fabd1/v2/protocol/http/options.go#L53  which (I believe) is a way to distinguish between a CE attribute and a Protocol attribute (http header). Meaning, use this option for HTTP headers and then SetAttribute for CE attributes. Then the user doesn't need to know about prefixes, that's a problem for the SDK to deal with.

So, yes I think we should introduce two types of attributes but I'm not sure we need two Options. Is PubSub different enough from HTTP that we couldn't just have your new Options deal with non-CE attributes and then then use SetAttribute for the CE ones?

Or, if you're suggesting that it would be nice to also have a way to define CE attributes for all Events flowing over this PubSub client (ie. "initial" CE attributes) then I think we should consider that for all protocols for consistency.

AndreasBergmeier6176:
> Using HTTP as an example, I see this `Option`:
> 
> https://github.com/cloudevents/sdk-go/blob/af1a06c50c614e4455e902b5e05d562bfc5fabd1/v2/protocol/http/options.go#L53
> which (I believe) is a way to distinguish between a CE attribute and a Protocol attribute (http header). Meaning, use this option for HTTP headers and then SetAttribute for CE attributes. Then the user doesn't need to know about prefixes, that's a problem for the SDK to deal with.
> 
> So, yes I think we should introduce two types of attributes but I'm not sure we need two Options. Is PubSub different enough from HTTP that we couldn't just have your new Options deal with non-CE attributes and then then use SetAttribute for the CE ones?

The way PubSub works is that the official client sends (Message publish) Requests via REST/GRPC to Google APIs. So e.g. the PubSub Message Attributes are then serialized into the HTTP/Protobuf body. I have seen people requesting for ways of modifying the actual REST/GRPC communication - and thus would be careful about reusing HTTP Options. Technically using PubSub may use HTTP and thus it might be confusing for HTTP Options to have a different effect. Also, looking at the `WithHeader` code, it has some very specific HTTP code.

All that being said - `Option`s are mostly for reusable configuration.
In our case, we might have a different value on every Message. Thus "solving" that by use of `Option` would force to recreate a protocol on every message which seems like a bit too much clutter.

Thus I personally would prefer for this to be implemented in a `Transformer`. To be able to pass the Transformer from `Client.Send` it is then necessary to extend the signature by being able to pass `Transformer`s.

> Or, if you're suggesting that it would be nice to also have a way to define CE attributes for all Events flowing over this PubSub client (ie. "initial" CE attributes) then I think we should consider that for all protocols for consistency.

I only want PubSub Message Attributes eventually set. Don't really care when.

AndreasBergmeier6176:
> Using HTTP as an example, I see this `Option`:
> 
> https://github.com/cloudevents/sdk-go/blob/af1a06c50c614e4455e902b5e05d562bfc5fabd1/v2/protocol/http/options.go#L53
> which (I believe) is a way to distinguish between a CE attribute and a Protocol attribute (http header). Meaning, use this option for HTTP headers and then SetAttribute for CE attributes. Then the user doesn't need to know about prefixes, that's a problem for the SDK to deal with.

Is this really the only way of modifying HTTP Headers? Thus you would have to recreate the whole `Protocol` if you want to set a Non-ce Header, that is unique for every message (e.g. TraceIds)?

duglin:
I'll look over your latest change, but wanted to comment on one bit:
> and thus would be careful about reusing HTTP Options.

I didn't mention HTTP to suggest we use HTTP's options, rather I was just looking at patterns from other protocols to see what we should learn, or use, for consistency.

protocol/pubsub/v2/attributes.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"context"
+
+	"github.com/cloudevents/sdk-go/v2/binding"
+)
+
+type withCustomAttributes struct{}
+
+func AttributesFrom(ctx context.Context) map[string]string {
```
embano1, line 19: nit: can be private, right?

protocol/pubsub/v2/attributes.go
```diff
@@ -0,0 +1,26 @@
...
+	return binding.GetOrDefaultFromCtx(ctx, withCustomAttributes{}, make(map[string]string)).(map[string]string)
+}
+
+// WithCustomAttributes sets Message Attributes without any Cloud Event logic.
+func WithCustomAttributes(ctx context.Context, attrs map[string]string) context.Context {
```
embano1, line 24: question: do we want this feature per event or as a global behavior on the PubSub protocol implementation (client)?

AndreasBergmeier6176: Per event

AndreasBergmeier6176:
To simplify the code I now moved setting the Attributes to where the PubSub Message gets built.

AndreasBergmeier6176:
> I'll look over your latest change, but wanted to comment on one bit:
> 
> > and thus would be careful about reusing HTTP Options.
> 
> I didn't mention HTTP to suggest we use HTTP's options, rather I was just looking at patterns from other protocols to see what we should learn, or use, for consistency.

Ah, misunderstood you there.
Anyway - now this PR should be pretty consistent with HTTP protocol.

protocol/pubsub/v2/protocol_test.go
```diff
@@ -20,6 +23,32 @@ type testPubsubClient struct {
...
+		"ce-exstring":         "exstring",
+		"ce-extime":           "2020-03-21T12:34:56.78Z",
+		"ce-exurl":            "http://example.com/source",
+		"ce-id":               "full-event",
+		"ce-source":           "http://example.com/source",
+		"ce-specversion":      "1.0",
+		"ce-subject":          "topic",
+		"ce-time":             "2020-03-21T12:34:56.78Z",
+		"ce-type":             "com.example.FullEvent",
+		"Proxy-Authorization": "YWxhZGRpbjpvcGVuc2VzYW1l",
```
duglin, line 43: So, just to be clear... you're expecting people to use `WithCustomAttributes` to set non-CE extension attributes and then `event.SetExtension` to set

AndreasBergmeier6176: Yeah, similar to people currently using `WithCustomHeader` to set non-CE extension attributes and then `event.SetExtension` to set CE extension attributes (for HTTP).

duglin: ok thanks. In that case, would it be possible to add this to our docs someplace? That might not be a fair ask since I think the docs are kind of lacking, but if you know of a good spot for us to describe this new Option then it would be nice to add it there. The main reason I'm asking is because any example associated with the docs should only show attribute w/o `ce-` unlike your test. I'm not keen on people thinking this is the right way to set CE attributes.

Maybe at a minimum, add some more text here: https://github.com/cloudevents/sdk-go/pull/976/files#diff-42e7fdd259e37dfd3740b150824fabc99fc50cbcb2d934ed1ca675d733121719R23 with something about how this is for protocol specific attributes and NOT for CE extensions, which should be set via the SetExtention func. Maybe?

AndreasBergmeier6176: Done.

duglin: thanks!

duglin:
thanks for your patience.

duglin (approved):

duglin:
ping @embano1 and @lionelvillard for review

AndreasBergmeier6176:
We are working with a fork now: https://github.com/abergmeier/cloudevents-sdk-go

embano1 (approved):
LGTM

@AndreasBergmeier6176 would you say there's any breaking/changed behavior for existing users or is the implementation (as I understand it) consistent with the current behavior unless a user uses the new attribute functions?

AndreasBergmeier6176:
> @AndreasBergmeier6176 would you say there's any breaking/changed behavior for existing users or is the implementation (as I understand it) consistent with the current behavior unless a user uses the new attribute functions?

The latter. Note that no current test had to be modified.

PR #978, opened 2023-10-29
Make a few comments a bit clearer

duglin:
Minor clean-up following https://github.com/cloudevents/sdk-go/pull/977

### DIFF
--- a/test/integration/mqtt_paho/mqtt_test.go
+++ b/test/integration/mqtt_paho/mqtt_test.go
@@ -38,8 +38,9 @@ func TestSendEvent(t *testing.T) {
 		eventChan := make(chan receiveEvent)
 		defer close(eventChan)
 
-		// Used to try to make sure the receiver is ready before we start to
-		// send events
+		// 'wait' is used to ensure that we at least wait until the Receiver
+		// thread starts. We'll then use a 'sleep' (hoping) that waits until
+		// the Receiver itself is ready
 		wait := make(chan bool)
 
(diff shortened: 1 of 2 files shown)


embano1 (approved):
Thx, realized my review comments didn't really make it during the review of the other PR. Appreciate you taking the time
to fix.

embano1:
Btw: with your fixes and reducing flaky integration tests, should we make them required to pass in CI before merging?

duglin:
> Btw: with your fixes and reducing flaky integration tests, should we make them required to pass in CI before merging?

I don't have a strong opinion on that since I assume we won't merge any PRs that have failing checks regardless of whether they're one of the 'required' ones or not.

embano1:
Reason I ask is that I can approve a change and use auto-merge once CI passes :) Enabling now.

duglin:
I was wondering if that was the reason... I have to then ask... why not require all checks to be required? Would we ever want to merge stuff that had a failing check? If so, I'm not sure why we have a check we don't really care about.

embano1:
> I was wondering if that was the reason... I have to then ask... why not require all checks to be required? Would we ever want to merge stuff that had a failing check? If so, I'm not sure why we have a check we don't really care about.

Most of the checks run super fast, i.e., before a maintainer does a review or when changes are pushed afterwards. But you raise a fair point. Right now, Github seems to have issues though: can't add the integration tests to required checks, so 🤷

PR #981, opened 2023-11-09
Bump the paho mqtt to v0.12

yanmxa:
Reference: https://github.com/eclipse/paho.golang/releases/tag/v0.12.0


duglin:
All mqtt users will to upgrade too, right?

yanmxa:
Do you mean whether this version is stable? 

It was released last month and contains small fixes/improvements.

embano1 (approved):
LGTM

duglin:
What's I'm wondering about is this: https://github.com/cloudevents/sdk-go/pull/981/files#diff-1a09a46967bd262e499762fc039a474889783193cc58172b29b3f643afae9511R148
While I understand that paho is only v0.12 (not a v1.0) so things can change, but this breaks existing SDK/mqtt users no?

It still may be ok, but I'd like to make sure we understand the ramifications.... if I'm correct.

yanmxa:
> What's I'm wondering about is this: https://github.com/cloudevents/sdk-go/pull/981/files#diff-1a09a46967bd262e499762fc039a474889783193cc58172b29b3f643afae9511R148 While I understand that paho is only v0.12 (not a v1.0) so things can change, but this breaks existing SDK/mqtt users no?
> 
> It still may be ok, but I'd like to make sure we understand the ramifications.... if I'm correct.

Yes! Users must modify a small amount of "cloudevent-client" initialization code to use the latest MQTT(0.12) SDK.

embano1:
Good 👀 @duglin !

Since it's been a while that we had a release, I also lost track of any potential breaking change we're introducing with the upcoming release. Usually this is mitigated with PR notes and change-log parsing/generation, which we haven't set up in the repo though.

duglin:
LGTM
one last check... @embano1 you're still ok with this, right?

duglin (approved):

duglin:
ping @embano1 and @lionelvillard for final review

embano1 (approved):
Someone just needs to keep in mind to add a `BREAKING` notice to the release notes :)

Issue #980, opened 2023-11-07
gRPC protocol implementation


morvencao:
The cloudevents sdk-go currently offers support for various [protocol implementations](https://github.com/cloudevents/sdk-go/tree/main/protocol/), except for **gRPC**.
It is essential to add gRPC protocol support since it is widely used.

While the [cloudevent proto ](https://github.com/cloudevents/sdk-go/blob/main/binding/format/protobuf/v2/pb/cloudevent.proto) is presented in the repository, it lacks the RPC service definition necessary for implementing the gRPC protocol in cloud event sdk-go.

To enable gRPC functionality, defining the RPC service and method is imperative. This ensures cloudevents sdk-go to send and receive through the generated golang gRPC client.

Can we incorporate the RPC service and method into the [cloudevent proto](https://github.com/cloudevents/sdk-go/blob/main/binding/format/protobuf/v2/pb/cloudevent.proto) in the following manner:

![image](https://github.com/cloudevents/sdk-go/assets/5468682/07f68f5e-dd7e-44b8-b2a0-caab668e8a1d)

Next, utilize `protoc` to generate the RPC client and server code, which can be employed to implement the gRPC protocol as follows:

```golang
# cat protocol/grpc/protocol.go
package grpc

import (
	"github.com/cloudevents/sdk-go/binding/format/protobuf/v2/pb"
)

type Protocol struct {
	// grpc client to publish and subscribe message
	grpcClient pb.CloudEventServiceClient
	// closeChan is used to close the receiver
	closeChan chan struct{}
}
...
```

/cc @embano1 @yanmxa

morvencao:
kindly ping @duglin

yanmxa:
Reference: https://github.com/cloudevents/sdk-go/issues/761

embano1:
@yanmxa sorry for my delayed response on your PR, been super busy lately which impacts my availability for this SDK :/

[...]

PR #984, opened 2023-11-09
add grpc protocol.

morvencao:
resolve: https://github.com/cloudevents/sdk-go/issues/980


protocol/grpc/write_message.go
```diff
@@ -0,0 +1,208 @@
...
+		if !ok {
+			return fmt.Errorf("invalid Type type, expected string got %T", value)
+		}
+		b.Type = val
+		return nil
+	case spec.DataContentType:
+		if value == nil {
+			delete(b.Attributes, contenttype)
+		}
+		b.Attributes[contenttype], _ = attributeFor(value)
```
duglin, line 129: Why do we delete the item from the map just to then re-add it here with `nil` ? Seems like this should be under an `else` or just remove the if+delete on lines 126-128. Same for below

protocol/grpc/message.go
```diff
@@ -0,0 +1,211 @@
...
+						return err
+					}
+				}
+			} else {
+				err = encoder.SetExtension(strings.TrimPrefix(name, prefix), v)
+				if err != nil {
+					return err
+				}
+			}
+		} else if name == contenttype {
```
duglin, line 137: what if there's both 'contenttype' and 'ce-datacontenttype' ?

protocol/grpc/write_message.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func (b *pbEventWriter) SetExtension(name string, value interface{}) (err error) {
+	if value == nil {
+		delete(b.Attributes, prefix+name)
+	}
+
+	b.Attributes[prefix+name], err = attributeFor(value)
+
+	return
```
duglin, line 166: Let's define 'err' in the func instead of in the return arg - just for consistency

duglin:
I see we have a sample, but can we get some testcases too?

duglin:
unit tests are failing

yanmxa:
From my humble understanding, protobuffer is a data transfer format. Directly defining a cloudevent implantation on it seems a bit of deviation. Instead, Can we just define a conversion between cloudevents and protobuffer format? like this PR: https://github.com/cloudevents/sdk-go/pull/793

protocol/grpc/write_message.go
morvencao: good catch, added `else` to handle non-nil case.

protocol/grpc/message.go
morvencao: 'ce-datacontenttype' will be handled by: https://github.com/cloudevents/sdk-go/blob/ab0722849ab42679f5dee3f08eb85cd6c5e7078f/protocol/grpc/message.go#L122-L128

protocol/grpc/write_message.go
```diff
@@ -0,0 +1,208 @@
...
+			delete(b.Attributes, prefix+subject)
+		}
+		b.Attributes[prefix+subject], _ = attributeFor(value)
+		return nil
+	case spec.Time:
+		if value == nil {
+			delete(b.Attributes, prefix+time)
+		}
+		b.Attributes[prefix+time], _ = attributeFor(value)
+		return nil
```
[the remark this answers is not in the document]
morvencao: thanks, done.

protocol/grpc/write_message.go
morvencao: done.

morvencao (commented):
Added some test cases, PTAL @duglin

duglin:
@morvencao can you take a look at the failing checks?

morvencao:
@duglin I've resolved the issues with testing and successfully executed it on my local environment, passing all tests.
I kindly request your approval to proceed with running the tests once more.

binding/format/protobuf/v2/pb/cloudevent.pb.go
```diff
@@ -7,7 +7,8 @@
 package pb
 
 import (
-	any "github.com/golang/protobuf/ptypes/any"
+	any1 "github.com/golang/protobuf/ptypes/any"
```
yanmxa, line 10: Why replace `any` with `any1`? Is this intentional？

binding/format/protobuf/v2/pb/cloudevent_grpc.pb.go
```diff
@@ -0,0 +1,175 @@
...
+
+const (
+	CloudEventService_Publish_FullMethodName   = "/pb.CloudEventService/Publish"
+	CloudEventService_Subscribe_FullMethodName = "/pb.CloudEventService/Subscribe"
+)
+
+// CloudEventServiceClient is the client API for CloudEventService service.
+//
+// For semantics around ctx use and closing/ending streaming RPCs, please refer to https://pkg.go.dev/google.golang.org/grpc/?tab=doc#ClientConn.NewStream.
+type CloudEventServiceClient interface {
```
yanmxa, line 30: The Publish-subscribe model has the following characteristics:
1. The publisher sends events without specifically targeting any recipient. These messages are broadcasted to all interested subscribers.
2. Subscribers express interest in certain types of messages and receive notifications or updates when messages of interest are published. This allows for a loosely coupled communication model where publishers and subscribers are decoupled from each other.

In my opinion, the grpc implementation seems more suitable for the request-response model. 
The requester initiates communication with a specific recipient, and the recipient responds directly to the requester's query or request. So I suggest naming the interface `Publish/Subscribe` to `Request/Response`

binding/format/protobuf/v2/pb/cloudevent.pb.go
morvencao: This code is generated by proto compiler, not manually altered.

protocol/grpc/message.go
duglin: yes but which one wins? Should we require them to be the same (if both are there)?

protocol/kafka_sarama/v2/coverage.tmp
```diff
@@ -1 +0,0 @@
-mode: atomic
```
duglin, line 1: can we delete this file?

morvencao: Uncertain, but based on the UT script at https://github.com/cloudevents/sdk-go/blob/main/hack/unit-test.sh#L20, it seems that `coverage.tmp` is produced as intermediate output by UT. Deleting it should pose no harm.

morvencao: I can restore it if there are additional uses that I overlooked.

protocol/grpc/message.go
morvencao: When both 'ce-datacontenttype' and 'contenttype' are set, 'ce-datacontenttype' takes precedence. In practical scenarios, only 'contenttype' is necessary. Check how the 'contenttype' attribute is configured here:
https://github.com/cloudevents/sdk-go/blob/65239604c031066dab2a25bde0719c898566c3db/protocol/grpc/write_message.go#L121-L130

This follow the similar approach that the Kafka protocol and PubSub protocol take, refer to the implementations in the CloudEvents SDK for Kafka (https://github.com/cloudevents/sdk-go/blob/main/protocol/kafka_sarama/v2/write_producer_message.go#L103-L114) and PubSub (https://github.com/cloudevents/sdk-go/blob/main/protocol/pubsub/v2/write_pubsub_message.go#L70-L80).

morvencao:
rebased the code.

morvencao:
fixed the testing and rebased the code.

duglin:
rebase is needed

duglin:
ping @morvencao

Issue #918, opened 2023-07-11
confluent kafka protocol binding


yanmxa:
Given the popularity of this repos: https://github.com/confluentinc/confluent-kafka-go, Are we also considering integrating it?

We now have the implementation of [sarama](https://github.com/Shopify/sarama), but there are still many people who are optimistic about the stability of [confluent](https://github.com/confluentinc/confluent-kafka-go). Also the confluent provides a high-level consumer, which is more convenient for asynchronous message confirmation([commit partition/offset](https://github.com/cloudevents/sdk-go/issues/846)) than sarma.

Of course, both clients have their own [advantages](https://www.reddit.com/r/golang/comments/olrd34/confluentkafkago_or_shopifysarama/), just considering that there are still many people who like confluent. If you consider adding the confluent protocol, I can bring its implementation.

refers: https://github.com/cloudevents/sdk-go/issues/846, https://github.com/cloudevents/sdk-go/issues/298, https://github.com/cloudevents/sdk-go/pull/183

Issue #846, opened 2023-02-17
How to submit message offset asynchronously based on kafka protocol


yanmxa:
**There is a scenario base on Kafka protocol binding and I'm not sure how to use the sdk to achieve it:**

When the cloudevent client receives an event, it returns the `NACK` response to tell Kafka not to commit the message offset. And also forwards the event to the backend program, when the backend program has completely consumed the event, then submits the offset of the message partition. 

If I can't use cloudevents client to submit message offset directly, how can I get the message partition and offset from the cloud event? so that the other progress can use the kafka client to submit this offset once the message is consumed successfully.

yanmxa:
Can we add the `sarama.ConsumerGroupSession` instance to the extension context of cloudevents, So we can use method `sess.MarkOffset(msg.Topic, msg.Partition, msg.Offset+1, metadata)` to update the offset asynchronously according to the result of the background goroutine for event processing.

embano1:
@yanmxa is this still an issue/open question?

yanmxa:
Yes. 

If it is based on this [saram](https://github.com/IBM/sarama) protocol, it seems a bit difficult to achieve asynchronous message confirmation. So I referred using the confluent protocol, https://github.com/cloudevents/sdk-go/issues/918, which can be combined with the cloudevents client to achieve asynchronous message confirmation.


https://github.com/cloudevents/sdk-go/issues/918

embano1:
Another popular one is https://github.com/twmb/franz-go.

@duglin how are we thinking about different implementations of open protocols, such as Kafka? It might be a bit of maintenance headache on the long-term to support multiple implementations of the same transport API. But I can understand the reasons as outlined by @yanmxa. Do we want to just provide one "reference" implementation, add multiple, or switch?

yanmxa:
Resolved by this PR: https://github.com/cloudevents/sdk-go/pull/988

PR #988, opened 2023-12-07
Confluent kafka binding

yanmxa:

Adding the Kafka binding with https://github.com/confluentinc/confluent-kafka-go. This implementation will support the following features compared to the [sarama](https://github.com/IBM/sarama) binding.

- The binding provides the ability to submit offset asynchronously. ref: https://github.com/cloudevents/sdk-go/issues/846
- Be able to subscribe to batch topics with a regular pattern. ref: https://github.com/confluentinc/confluent-kafka-go/issues/96
- Support to consume the event from a specific position(topic-partition-offset)
- Support to invoke the customized rebalance callback function for the consumer group

Note: In the confluent samples, with configuration [WithPollGoroutines(1)](https://github.com/cloudevents/sdk-go/blob/ba92a5c9a1e2508d855930049999ccd75636f0cb/v2/client/options.go#L85C6-L85C11) to ensure the order of the messages on the partition. 

   Since Kafka already natively utilizes consumer groups to enhance consumption efficiency, there is no need to use multiple goroutines to boost consumption speed. 

Resolve: https://github.com/cloudevents/sdk-go/issues/918


test/integration/kafka_confluent/kafka_test.go
```diff
@@ -0,0 +1,111 @@
...
+			test.HasData(eventIn.Data()),
+			test.HasExtensionKeys([]string{kafka_confluent.KafkaPartitionKey, kafka_confluent.KafkaOffsetKey}),
+			test.HasExtension(kafka_confluent.KafkaTopicKey, topicName),
+		)(eventOut)
+		require.NoError(t, err)
+	})
+}
+
+// To start a local environment for testing:
+// docker run --rm --net=host -e ADV_HOST=localhost -e SAMPLEDATA=0 -p 9091:9091 -p 9092:9092 lensesio/fast-data-dev
```
embano1, line 90: question: should we use the `confluent` docker images for integration testing instead? Assuming many will use this binding with `confluent` Kafka (which is Kafka, I understand, however might have some differences when it comes to defaults, connection/auth handling, etc.)

v2/context/context.go
```diff
@@ -57,6 +57,48 @@ func TopicFrom(ctx context.Context) string {
 	return ""
 }
 
+// Opaque key type used to store topic partition
+type topicPartitionKeyType struct{}
```
embano1, line 61: question: compared to the other types/functions in this package, this seems very Kafka specific. I'm assuming you have this in this package to influence per message handling vs in the protocol? IIRC, we have other protocol implementations allowing per-message send/receive properties in the protocol w/out changing this package. Can you please double-check?

test/integration/kafka_confluent/kafka_test.go
yanmxa: I tested both `lensesio/fast-data-dev` and `confluentinc/confluent-local` images. I didn't find any differences in their usage during the test. 

I understand your concerns. However, running the `confluentinc/confluent-local` in the integration workflow might cause issues with the `lensesio/fast-data-dev`. I can start another PR focus on these issues. What do you think of it?

v2/context/context.go
yanmxa: My initial reason for putting these properties in this package is that the sarama binding might leverage them. Like you said it's only for the kafka, to avoid unnecessary reference confusion, I have removed them from here.

yanmxa:
@embano1 Thanks for your review!

embano1:
Thy @yanmxa and sorry for my slow responses. I'm under the water at the moment at work and can't give you a concrete date when I'll be able to review. Perhaps @duglin and @lionelvillard can take another look?

lionelvillard:
@yanmxa can you update the documentation: https://github.com/cloudevents/sdk-go/blob/main/docs/protocol_implementations.md?

Also I wonder if it makes sense to indicate (maybe in the documentation) this is a new implementation and it hasn't been widely tested in the wild. WDYT?

yanmxa:
> @yanmxa can you update the documentation: https://github.com/cloudevents/sdk-go/blob/main/docs/protocol_implementations.md?
> 
> Also I wonder if it makes sense to indicate (maybe in the documentation) this is a new implementation and it hasn't been widely tested in the wild. WDYT?

Thanks @lionelvillard! I update the protocol binding for this implementation. PTAL~

yanmxa:
Thanks @duglin !

yanmxa:
@embano1 @lionelvillard PTAL~

protocol/kafka_confluent/v2/go.mod
```diff
@@ -0,0 +1,25 @@
+module github.com/cloudevents/sdk-go/protocol/kafka_confluent/v2
+
+go 1.18
+
+replace github.com/cloudevents/sdk-go/v2 => ../../../v2
+
+require (
+	github.com/cloudevents/sdk-go/v2 v2.0.0-00010101000000-000000000000
```
embano1, line 8: bump to latest?

protocol/kafka_confluent/v2/message.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/cloudevents/sdk-go/v2/binding"
+	"github.com/cloudevents/sdk-go/v2/binding/format"
+	"github.com/cloudevents/sdk-go/v2/binding/spec"
+	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
+)
+
+const (
+	prefix         = "ce-"
+	contentTypeKey = "Content-Type"
```
embano1, line 22: question: why not make it lower here to directly?

protocol/kafka_confluent/v2/message.go
```diff
@@ -0,0 +1,143 @@
...
+func NewMessage(msg *kafka.Message) *Message {
+	var contentType, contentVersion string
+	properties := make(map[string][]byte, len(msg.Headers)+3)
+	for _, header := range msg.Headers {
+		k := strings.ToLower(string(header.Key))
+		if k == strings.ToLower(contentTypeKey) {
+			contentType = string(header.Value)
+		}
+		if k == specs.PrefixedSpecVersionName() {
+			contentVersion = string(header.Value)
```
embano1, line 58: question: what if `header.Value` is `1.0.2` but MUST be `1.0` according to 

> Compliant event producers MUST use a value of 1.0 when referring to this version of the specification
https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#specversion

protocol/kafka_confluent/v2/message.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// Check if Message implements binding.Message
+var (
+	_ binding.Message               = (*Message)(nil)
+	_ binding.MessageMetadataReader = (*Message)(nil)
+)
+
+func NewMessage(msg *kafka.Message) *Message {
+	var contentType, contentVersion string
```
embano1, line 50: nit: want to add `nil` check for `msg` and `msg.TopicPartition.Topic` and `panic()` immediately if someone uses the function wrong?

protocol/kafka_confluent/v2/option.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil
+	}
+}
+
+// Opaque key type used to store offsets: assgin offset from ctx, commit offset from context
+type commitOffsetType struct{}
+
+var offsetKey = commitOffsetType{}
+
+// WithOffsetCtx will return the topic partitions to commit offsets for.
```
embano1, line 102: nit: this is a modify operation, so the comment seems a bit off

protocol/kafka_confluent/v2/option.go
```diff
@@ -0,0 +1,158 @@
...
+	return nil
+}
+
+// Opaque key type used to store topic partition
+type topicPartitionKeyType struct{}
+
+var topicPartitionKey = topicPartitionKeyType{}
+
+// WithTopicPartition returns back a new context with the given partition.
+func WithTopicPartition(ctx context.Context, partition int32) context.Context {
```
embano1, line 124: this comment is more clear than the above

protocol/kafka_confluent/v2/option.go
```diff
@@ -0,0 +1,158 @@
...
+	"context"
+	"fmt"
+
+	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
+)
+
+// Option is the function signature required to be considered an kafka_confluent.Option.
+type Option func(*Protocol) error
+
+func WithConfigMap(config *kafka.ConfigMap) Option {
```
embano1, line 18: nit: comment missing

protocol/kafka_confluent/v2/option.go
embano1: also, I found the name to be a bit misleading:

1) remove the `Ctx` suffix to follow option style
2) There's no offset parameter in the signature (hidden deep inside topicPartitions)

IIUC what this does it sets the offsets to start consuming from, per topic/partition. If so, rename parameter to `topicPartitionOffsets`. Also, some validations might make sense e.g., checking if all three properties are set in each slice value.

embano1 (commented):
I haven't checked the whole PR because for some parts especially structured/binary handling, I'd like to see more tests for non-happy path scenarios and mis-use if possible. Could you please also add Confluent to the Github integration tests so we run this against Confluent images?

protocol/kafka_confluent/v2/message.go
yanmxa: I‘m not sure how to handle the specversion with value 1.0.2. Maybe just validate the 'major' and 'minor' versions whether compliant with "1.0"? and if not, then print a warning message?

I checked the [other protocol](https://github.com/cloudevents/sdk-go/blob/0a3095dbbd1db5c6bbd6a039c927dcac876d4f1b/protocol/kafka_sarama/v2/message.go#L74). Then all just assert whether the specversion key exists and ignore the value. 

Do you have any other advice?

protocol/kafka_confluent/v2/go.mod
yanmxa: Done.

protocol/kafka_confluent/v2/message.go
yanmxa: No particular reason to use the capital. I thought the `contety-type` might be a special property to decide the event format.

To align it with the sarama protocol. I changed it to lowercase.

protocol/kafka_confluent/v2/option.go
yanmxa: Done

protocol/kafka_confluent/v2/message.go
yanmxa: Done.

yanmxa:
> I haven't checked the whole PR because for some parts especially structured/binary handling, I'd like to see more tests for non-happy path scenarios and mis-use if possible. Could you please also add Confluent to the Github integration tests so we run this against Confluent images?

Thank @embano1! 

I added some tests for the misuse cases when the consumers try to initialize the client. If others come up with scenarios beyond this, we can continually improve it in the future.

Also, I updated the GitHub integration test workflow to add the kafka_confluent service(running on port 9192, the sarama running on 9092) with the confluent image.  So the current confluent kafka protocol uses it for the tests.

protocol/kafka_confluent/v2/message.go
embano1: Yeah, IMHO this is a shortcoming in the other implementations. I'm fine leaving this here as is then given it's consistent with our other implementations. cc/ @duglin

.github/workflows/integration.yaml
```diff
@@ -27,6 +27,17 @@ jobs:
           - 9091:9091
           - 9092:9092
 
+      kafka_confluent:
+        image: confluentinc/confluent-local:latest
```
embano1, line 31: note to myself: Ideally we should always use tags (or better) SHAs. But this is not done in the other workflows either yet, making troubleshooting harder.

protocol/kafka_confluent/v2/message.go
embano1: nit: just noticed code comment missing for exported function

protocol/kafka_confluent/v2/message.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		properties[k] = header.Value
+	}
+
+	// add the kafka message key, topic, partition and partition key to the properties
+	properties[prefix+KafkaOffsetKey] = []byte(strconv.FormatInt(int64(msg.TopicPartition.Offset), 10))
+	properties[prefix+KafkaPartitionKey] = []byte(strconv.FormatInt(int64(msg.TopicPartition.Partition), 10))
+	if msg.TopicPartition.Topic == nil {
+		panic("the kafka message topic shouldn't be nil")
+	}
```
embano1, line 68: note: the nil check should come at the beginning of the function and check whether:

- `msg` is nil
- `msg.TopicPartition.Topic` is nil

protocol/kafka_confluent/v2/message_test.go
```diff
@@ -0,0 +1,113 @@
...
+		Value:          []byte("hello world!"),
+		Headers: mustToKafkaHeaders(map[string]string{
+			"ce_type":            testEvent.Type(),
+			"ce_source":          testEvent.Source(),
+			"ce_id":              testEvent.ID(),
+			"ce_time":            test.Timestamp.String(),
+			"ce-specversion":     "1.0",
+			"ce_dataschema":      test.Schema.String(),
+			"ce_datacontenttype": "text/json",
+			"ce_subject":         "receiverTopic",
```
embano1, line 52: wondering why they're all using `_` instead of `-`?

protocol/kafka_confluent/v2/message_test.go
```diff
@@ -0,0 +1,113 @@
...
+			"ce_exta":            "someext",
```
embano1, line 53: question: for custom extensions, generally we don't use `ce` prefix. Is this supposed to test custom extension?

protocol/kafka_confluent/v2/message_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+			if tt.expectedEncoding == binding.EncodingBinary {
+				err = msg.ReadBinary(ctx, (*kafkaMessageWriter)(tt.consumerMessage))
+			}
+			require.Nil(t, err)
+		})
+	}
+}
+
+func mustToKafkaHeaders(m map[string]string) []kafka.Header {
```
embano1, line 105: nit: `must` functions usually panic

protocol/kafka_confluent/v2/option.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+}
+
+// Opaque key type used to store topicPartitionOffsets: assign them from ctx. This option is not required.
+type topicPartitionOffsetsType struct{}
+
+var offsetKey = topicPartitionOffsetsType{}
+
+// WithTopicPartitionOffsets will set the positions where the consumer starts consuming from. This option is not required.
+func WithTopicPartitionOffsets(ctx context.Context, topicPartitionOffsets []kafka.TopicPartition) context.Context {
```
embano1, line 109: nice!

protocol/kafka_confluent/v2/option.go
```diff
@@ -0,0 +1,178 @@
...
+			panic("the kafka topic cannot be nil or empty")
+		}
+		if offset.Partition < 0 || offset.Offset < 0 {
+			panic("the kafka partition/offset must be non-negative")
+		}
+	}
+	return context.WithValue(ctx, offsetKey, topicPartitionOffsets)
+}
+
+// TopicPartitionOffsetsFrom looks in the given context and returns `[]kafka.TopicPartition`
```
embano1, line 124:
```suggestion
// TopicPartitionOffsetsFrom looks in the given context and returns []kafka.TopicPartition or nil if not set
```

protocol/kafka_confluent/v2/option.go
```diff
@@ -0,0 +1,178 @@
...
+	return nil
+}
+
+// Opaque key type used to store topic partition
+type topicPartitionKeyType struct{}
+
+var topicPartitionKey = topicPartitionKeyType{}
+
+// WithTopicPartition returns back a new context with the given partition. This option is not required.
+func WithTopicPartition(ctx context.Context, partition int32) context.Context {
```
embano1, line 141: Wondering where/how this would be used? For example, there's also `WithTopicPartitionsOffsets` which I found similar and the latter one referencing a topic whereas this function just takes a partition (for which topic)?

samples/kafka_confluent/go.mod
```diff
@@ -0,0 +1,23 @@
+module github.com/cloudevents/sdk-go/samples/kafka_confluent
+
+go 1.18
+
+replace github.com/cloudevents/sdk-go/v2 => ../../v2
+
+replace github.com/cloudevents/sdk-go/protocol/kafka_confluent/v2 => ../../protocol/kafka_confluent/v2
```
embano1, line 7: needed?

protocol/kafka_confluent/v2/message.go
```diff
@@ -0,0 +1,146 @@
...
+			err = encoder.SetAttribute(m.version.AttributeFromKind(spec.DataContentType), string(v))
+		}
+		if err != nil {
+			return err
+		}
+	}
+
+	if m.internal.Value != nil {
+		err = encoder.SetData(bytes.NewBuffer(m.internal.Value))
+	}
```
embano1, line 128: btw: have you tested (E2E) with tombstone records, i.e., where value is `null` (`nil`)? Curious about the behavior.

.github/workflows/integration.yaml
```diff
@@ -27,6 +27,17 @@ jobs:
...
+        hostname: broker
+        container_name: broker
```
embano1, line 33: tests are failing due to this; needed?

.github/workflows/integration.yaml
yanmxa: Done with `latest` -> `7.6.0`.

protocol/kafka_confluent/v2/message.go
yanmxa: Done.

protocol/kafka_confluent/v2/message_test.go
yanmxa: Nice catch! 
It's a typo. I change the `_` into `-`, to align them with the `prefix` defined in `message.go`

samples/kafka_confluent/go.mod
yanmxa: yes, we need it

.github/workflows/integration.yaml
yanmxa: Yes. I tried to remove it but failed if we set a customized port(9192, the default is 9092) to expose the service on the workflow

protocol/kafka_confluent/v2/message.go
yanmxa: The value with a `nil` case is covered in the integration test. 
The input(sender) event:
```
Context Attributes,
  specversion: 1.0
  type: com.example.MinEvent
  source: http://example.com/source
  id: min-event-2
```

the output(receiver) event:
```bash
Context Attributes,
  specversion: 1.0
  type: com.example.MinEvent
  source: http://example.com/source
  id: min-event-2
Extensions,
  kafkaoffset: 0
  kafkapartition: 0
  kafkatopic: test-ce-confluent-56b0d978-1e0f-4aa3-aecd-993322685530
```

yanmxa:
@embano1 
Thanks for your patient reviews all along!
Cause several multi-cloud management projects in our team depend on this. If you could provide feedback and suggestions on this PR as soon as possible, it would be of great help to us!

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,212 @@
...
+			}
+			p.producer = producer
+		}
+		if p.producer == nil && p.consumer == nil {
+			return nil, fmt.Errorf("at least set one of the receiver or sender topic")
+		}
+	}
+
+	if p.kafkaConfigMap == nil && p.producer == nil && p.consumer == nil {
+		return nil, fmt.Errorf("at least one of the following to initialize the protocol: config, producer, or consumer")
```
embano1, line 76:
```suggestion
		return nil, fmt.Errorf("at least one of the following to initialize the protocol must be set: config, producer, or consumer")
```

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+		if p.producerDefaultTopic != "" && p.producer == nil {
+			producer, err := kafka.NewProducer(p.kafkaConfigMap)
+			if err != nil {
+				return nil, err
+			}
+			p.producer = producer
+		}
+		if p.producer == nil && p.consumer == nil {
+			return nil, fmt.Errorf("at least set one of the receiver or sender topic")
```
embano1, line 71:
```suggestion
			return nil, fmt.Errorf("at least receiver or sender topic must be set")
```

protocol/kafka_confluent/v2/option.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	return nil
+}
+
+// Opaque key type used to store message key
+type messageKeyType struct{}
+
+var keyForMessageKey = messageKeyType{}
+
+// WithMessageKey returns back a new context with the given messageKey. This option is not required.
```
embano1, line 137:
```suggestion
// WithMessageKey returns back a new context with the given messageKey.
```

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,212 @@
...
+		if err := fn(p); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (p *Protocol) Send(ctx context.Context, in binding.Message, transformers ...binding.Transformer) (err error) {
+	if p.producer == nil {
+		return fmt.Errorf("the producer client must not be nil")
```
embano1, line 95: nit (same for others:) 

```suggestion
		return fmt.Errorf("producer client must be set")
```

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,212 @@
...
+	return nil
+}
+
+func (p *Protocol) OpenInbound(ctx context.Context) error {
+	if p.consumer == nil {
+		return fmt.Errorf("the consumer client must not be nil")
+	}
+	if p.consumerTopics == nil {
+		return fmt.Errorf("the consumer topics must not be nil")
+	}
```
embano1, line 137: see comment above

samples/kafka_confluent/receiver-assign/main.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
+)
+
+var topic = "test-confluent-topic"
+
+func main() {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	receiver, err := kafka_confluent.New(kafka_confluent.WithConfigMap(&kafka.ConfigMap{
+		"bootstrap.servers": "127.0.0.1:9092",
```
embano1, line 25: will confluent bind to `9092` or `9091` by default?

.github/workflows/integration.yaml
embano1: Please fix, otherwise can't approve

samples/kafka_confluent/receiver-assign/main.go
yanmxa: The confluent binds to 9092 by default

.github/workflows/integration.yaml
yanmxa: @embano1 I have removed the `hostname` and `container_name` from here now.

samples/kafka_confluent/receiver-assign/main.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	kafka_confluent "github.com/cloudevents/sdk-go/protocol/kafka_confluent/v2"
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/cloudevents/sdk-go/v2/client"
+	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
```
embano1, line 16: nit: if you use import alias `confluent` code will be `confluent.New` and easier to read

samples/kafka_confluent/receiver-assign/main.go
```diff
@@ -0,0 +1,60 @@
...
+		// "auto.offset.reset":  "earliest",
+		"enable.auto.commit": "true",
+	}), kafka_confluent.WithReceiverTopics([]string{topic}))
+	if err != nil {
+		log.Fatalf("failed to create kafka protocol, %v", err)
+	}
+
+	defer receiver.Close(ctx)
+
+	c, err := cloudevents.NewClient(receiver, client.WithPollGoroutines(1))
```
embano1, line 36: question: reason you set goroutines to `1`?

samples/kafka_confluent/receiver-assign/main.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	log.Printf("will listen consuming topic %s\n", topic)
+	err = c.StartReceiver(kafka_confluent.WithTopicPartitionOffsets(ctx, offsetToStart), receive)
+	if err != nil {
+		log.Fatalf("failed to start receiver: %s", err)
+	} else {
+		log.Printf("receiver stopped\n")
+	}
+	cancel()
```
embano1, line 52: move up to ctx creation and use `defer cancel()`

embano1 (changes requested):
Some minor things left, sorry...had some time to dig a bit deeper. Should be easy to address. Then please squash your commits.

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,213 @@
...
+			if ev == nil {
+				continue
+			}
+			switch e := ev.(type) {
+			case *kafka.Message:
+				p.incoming <- e
+			case kafka.Error:
+				// Errors should generally be considered informational, the client will try to automatically recover.
+				// But in here, we choose to terminate the application if all brokers are down.
+				logger.Infof("Error %v: %v", e.Code(), e)
```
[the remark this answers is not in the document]
yanmxa: It has the `Errorf`. 
But I have read the [confluent official samples](https://github.com/confluentinc/confluent-kafka-go/blob/master/examples/producer_example/producer_example.go), then all treat this `kafka.Error` as an informational level. 
So I added an optional callback function named `ErrorHandler`. The customer can choose to proceed or abort the message consuming based on the error type.

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,213 @@
...
+			run = false
+		default:
+			ev := p.consumer.Poll(p.consumerPollTimeout)
+			if ev == nil {
+				continue
+			}
+			switch e := ev.(type) {
+			case *kafka.Message:
+				p.incoming <- e
+			case kafka.Error:
```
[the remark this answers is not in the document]
yanmxa: Yes. The `kafka.Error` shouldn't be a pointer. Otherwise, we will not catch the error message on the switch case.

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,213 @@
...
+	err := p.consumer.SubscribeTopics(p.consumerTopics, p.consumerRebalanceCb)
+	if err != nil {
+		return err
+	}
+
+	run := true
+	for run {
+		select {
+		case <-ctx.Done():
+			run = false
```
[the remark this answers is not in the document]
yanmxa: Done

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,213 @@
...
+				// But in here, we choose to terminate the application if all brokers are down.
+				logger.Infof("Error %v: %v", e.Code(), e)
+				if p.consumerErrorHandler != nil {
+					p.consumerErrorHandler(e)
+				}
+				if e.Code() == kafka.ErrAllBrokersDown {
+					run = false
+				}
+			default:
+				fmt.Printf("Ignored %v\n", e)
```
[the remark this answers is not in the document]
yanmxa: Done

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,213 @@
...
+			}
+		}
+	}
+
+	logger.Infof("Closing consumer %v", p.consumerTopics)
+	return p.consumer.Close()
```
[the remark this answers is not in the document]
yanmxa: Nice catch! 
1. It returns the error message in the for loop now.
2. If someone calls `Close()` concurrently. It will correctly handle this.

samples/kafka_confluent/receiver-assign/main.go
yanmxa: Done

samples/kafka_confluent/receiver-assign/main.go
yanmxa: I think that using multiple goroutines may break the order of messages in the Kafka partition, please correct me if I am wrong.

samples/kafka_confluent/receiver-assign/main.go
yanmxa: Done

yanmxa:
> Some minor things left, sorry...had some time to dig a bit deeper. Should be easy to address. Then please squash your commits.

Never mind! 
Done.

protocol/kafka_confluent/v2/protocol.go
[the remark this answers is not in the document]
embano1: Got it, wasn't sure whether they use pointers here as well.

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,225 @@
...
+)
+
+type Protocol struct {
+	kafkaConfigMap *kafka.ConfigMap
+
+	consumer             *kafka.Consumer
+	consumerTopics       []string
+	consumerRebalanceCb  kafka.RebalanceCb     // optional
+	consumerPollTimeout  int                   // optional
+	consumerErrorHandler func(err kafka.Error) //optional
```
embano1, line 35: regarding error handling discussion: let's add `ctx` here so users can shove in custom loggers without having to use closures?

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+		if p.producerDefaultTopic != "" && p.producer == nil {
+			producer, err := kafka.NewProducer(p.kafkaConfigMap)
+			if err != nil {
+				return nil, err
+			}
+			p.producer = producer
+		}
+		if p.producer == nil && p.consumer == nil {
+			return nil, fmt.Errorf("at least receiver or sender topic must be set")
```
embano1, line 71: sorry, just saw: any linter would flag this at some point: `Errorf` without variable placeholders (same with other errors)

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,225 @@
...
+				p.incoming <- e
+			case kafka.Error:
+				// Errors should generally be considered informational, the client will try to automatically recover.
+				// But in here, we choose to terminate the application if all brokers are down.
+				logger.Infof("Error %v: %v", e.Code(), e)
+				if p.consumerErrorHandler != nil {
+					p.consumerErrorHandler(e)
+				}
+				if e.Code() == kafka.ErrAllBrokersDown {
+					logger.Debugf("All broker connections are down!")
```
embano1, line 186: so let's make this `Errorf` then to not hide this error when debugging is off.

protocol/kafka_confluent/v2/protocol.go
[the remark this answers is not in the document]
embano1: OK

samples/kafka_confluent/receiver-assign/main.go
embano1: Hm, ideally you'd have `WithPollGoroutines == kafka_partitions`, right?

samples/kafka_confluent/receiver-assign/main.go
embano1: and please remove this here

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,225 @@
...
+					return e
+				}
+			default:
+				logger.Debugf("Ignored the kafka.Event: %v", e)
```
embano1, line 190: question: will this print the pointer or real value? What could `e` be as an example?

embano1: (asking perhaps we don't need `default` at all)

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+		msg := NewMessage(m)
+		return msg, nil
+	case <-ctx.Done():
+		return nil, io.EOF
+	}
+}
+
+func (p *Protocol) Close(ctx context.Context) error {
+	if p.consumer != nil && !p.consumer.IsClosed() {
```
embano1, line 212: Now you have a data race, if this is called from another goroutine. You need to protect with mutex here as well and be careful with deadlocking

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,225 @@
...
+func (p *Protocol) OpenInbound(ctx context.Context) error {
+	if p.consumer == nil {
+		return fmt.Errorf("the consumer client must be set")
+	}
+	if p.consumerTopics == nil {
+		return fmt.Errorf("the consumer topics must be set")
+	}
+
+	p.consumerMux.Lock()
+	defer p.consumerMux.Unlock()
```
embano1, line 140: can lead to deadlocks since you're holding on it for function lifetime. See comment below

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	logger.Infof("Subscribing to topics: %v", p.consumerTopics)
+	err := p.consumer.SubscribeTopics(p.consumerTopics, p.consumerRebalanceCb)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if !p.consumer.IsClosed() {
```
embano1, line 157: use mutex here when removing the "global" mutex above

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,225 @@
...
+			logger.Infof("Closing consumer %v", p.consumerTopics)
+			if err = p.consumer.Close(); err != nil {
+				logger.Errorf("failed to close the consumer: %v", err)
+			}
+		}
+	}()
+
+	for !p.consumer.IsClosed() {
```
embano1, line 165: simply drop this bool and make it a simple `for { select { } }`. We rely on `ctx` to terminate this loop. Just make sure to protext access to `isClosed` with mutex.

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,225 @@
...
+		select {
+		case <-ctx.Done():
+			logger.Debugf("The consumer context is canceled!")
+			return nil
```
embano1, line 169: nit: `return ctx.Err()` now.

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,225 @@
...
+				if e.Code() == kafka.ErrAllBrokersDown {
+					logger.Debugf("All broker connections are down!")
+					return e
+				}
+			default:
+				logger.Debugf("Ignored the kafka.Event: %v", e)
+			}
+		}
+	}
+	return fmt.Errorf("the consumer is closed")
```
embano1, line 194: remove

embano1 (changes requested):

samples/kafka_confluent/receiver-assign/main.go
yanmxa: Emm I don't think so. It's a little bit complicated.

Cause Kafka uses multiple goroutines(consumer group) to poll messages from the partitions on the broker. 
Ideally, the `max number of the consumer instance in the group` = the `number of the Kafka topic partitions`, and each of consumer will get a unique message from that assigned partition. This allows Kafka consumption throughput to reach its maximum while ensuring local ordering (partition). All these features all provided by Kafka itself. 

From my current understanding, Cloudevents also uses a multi-goroutines model to improve receiver consumption speed. However, if multiple `PollGoroutines` are used in conjunction with Kafka's consumer group feature, it will disrupt the intended purpose of the Kafka consumer group model because we cannot guarantee message orderliness under the message key partitioning mechanism.

So I set the `WithPollGoroutines` to `1` based on the above consideration

samples/kafka_confluent/receiver-assign/main.go
yanmxa: Done

protocol/kafka_confluent/v2/protocol.go
yanmxa: Done.

protocol/kafka_confluent/v2/protocol.go
yanmxa: Done.

protocol/kafka_confluent/v2/protocol.go
yanmxa: Done.

protocol/kafka_confluent/v2/protocol.go
yanmxa: Done

protocol/kafka_confluent/v2/protocol.go
yanmxa: Done

protocol/kafka_confluent/v2/protocol.go
yanmxa: Whether the mutex is intended to protect the `OpenInbound` can only be invoked once to avoid launching the consumer multiple times. Cause the saram implementation also uses this mutex for the "consumer loop" on the `OpenInbound`.

protocol/kafka_confluent/v2/protocol.go
yanmxa: Sure! Only keeping the `ctx` to terminate the loop is going to make the logic simpler. However, there is a scenario I'm not sure whether we need to consider:

1. Use the `kafka.Consumer` to initialize a `cloudevents.Client`.  
2. Start polling the message by invoking the `StartReceiver`(blocking) of the `cloudevents.Client`. 
3. If we call the `kafka.Consumer.Close()` outside, then the code will panic causing we can't use a closed consumer to poll message.

Since the `cloudevents.Client` hasn't exposed the `Close()` method, maybe we don't need to handle it? I'm not sure, need your confirm...

yanmxa: And I have removed the `p.consumer.IsClosed()` from the consumer for loop now

protocol/kafka_confluent/v2/protocol.go
yanmxa: Well, actually I haven't come across this default case. This code is written according to the [official samples](https://github.com/confluentinc/confluent-kafka-go/blob/f1230c0e9dd4b038986ecf26a901e945ec71ca23/examples/consumer_example/consumer_example.go#L117-L118). I am not sure if it contains some cases we unexpected. 

I agree with your idea and have removed the `default`. We can add it back if it is really needed in the future.

protocol/kafka_confluent/v2/protocol.go
yanmxa: I added another lock(mutex) to handle the data race between the `Opener` and `Closer`. It will make sure that the `p.consumer.Close()` and `p.consumer().IsClose()` will don't execute in parallel.

protocol/kafka_confluent/v2/protocol.go
yanmxa: See the [above](https://github.com/cloudevents/sdk-go/pull/988#discussion_r1527525977).

protocol/kafka_confluent/v2/protocol.go
embano1: Ah, got it yes.

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = p.producer.Produce(kafkaMsg, p.producerDeliveryChan)
+	if err != nil {
+		return err
+	}
+	e := <-p.producerDeliveryChan
+	m := e.(*kafka.Message)
```
embano1, line 127: question: will this always be this type or do we need assertion check?

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,232 @@
...
+	if m.TopicPartition.Error != nil {
```
embano1, line 128: is TopicPartition also pointer? nil check required as well?

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	logger.Infof("Subscribing to topics: %v", p.consumerTopics)
+	err := p.consumer.SubscribeTopics(p.consumerTopics, p.consumerRebalanceCb)
+	if err != nil {
+		return err
+	}
+
+	p.closerMux.Lock()
+	p.consumerCtx, p.consumerCancel = context.WithCancel(ctx)
```
embano1, line 160: also call defer p.consumerCancel() ?

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,232 @@
...
+
+func (p *Protocol) Close(ctx context.Context) error {
+	p.closerMux.Lock()
+	defer p.closerMux.Unlock()
+
+	if p.consumerCancel != nil {
+		p.consumerCancel()
+	}
+	if p.consumerIncoming != nil {
+		close(p.consumerIncoming)
```
embano1, line 222: this can cause a panic in openinbound loop when you're within the poll routine but concurrently close this channel
my recommendation is to not close channels here but handle closing those in send/openinbound and just call consumerCancel() here

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	if p.producer != nil && !p.producer.IsClosed() {
```
embano1, line 225: IMHO race with Send

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,232 @@
...
+		p.producer.Close()
+	}
+	if p.producerDeliveryChan != nil {
+		close(p.producerDeliveryChan)
```
embano1, line 229: can cause panic if concurrently in send, see comment above

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,232 @@
...
+	for _, fn := range opts {
+		if err := fn(p); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (p *Protocol) Send(ctx context.Context, in binding.Message, transformers ...binding.Transformer) (err error) {
+	if p.producer == nil {
```
embano1, line 97: if you add a producerMtx check here whether it's "closed" you can return early and avoid concurrency issues (see comments in Close())

embano1 (changes requested):
Looked closer at the Close/Send/OpenInbound implementations and there's still races/chances for panics. Left comments on how to simplify

protocol/kafka_confluent/v2/protocol.go
yanmxa: I think the answer is yes based on the [official example](https://github.com/confluentinc/confluent-kafka-go/blob/32f065e0ede50a539847a4e9bf2d8b9e8f8f5447/examples/producer_custom_channel_example/producer_custom_channel_example.go#L80-L101) they provide.

protocol/kafka_confluent/v2/protocol.go
yanmxa: Done!

protocol/kafka_confluent/v2/protocol.go
yanmxa: Done!

protocol/kafka_confluent/v2/protocol.go
yanmxa: Yes! I also leverage the `closerMux` to handle the race condition between `Sender` and `Closer`

protocol/kafka_confluent/v2/protocol.go
yanmxa: Done.

protocol/kafka_confluent/v2/protocol.go
yanmxa: The `closerMux` is intended to handle the race condition between `Sender`/`Opener` and `Closer`. So instead add the new mutex(producerMtx), I still use the `closerMux` to handle it.

yanmxa:
> Looked closer at the Close/Send/OpenInbound implementations and there's still races/chances for panics. Left comments on how to simplify

Thanks a lot!
Done!

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+		if p.consumerIncoming != nil && !isClosed(p.producerDeliveryChan) {
+			close(p.consumerIncoming)
+		}
+	}()
+
+	for {
+		select {
+		case <-p.consumerCtx.Done():
+			return fmt.Errorf("the consumer context is canceled: %v", p.consumerCtx.Err())
```
embano1, line 189: return ctx.Err()

samples/kafka_confluent/receiver-assign/main.go
embano1: Just to check: you're saying the Confluent implementation handles goroutines per consumer group accordingly by default?

protocol/kafka_confluent/v2/protocol.go
yanmxa: @embano1 Done.

samples/kafka_confluent/receiver-assign/main.go
yanmxa: Yes, by default, the consumer group associated with the confluent consumer manages all consumers, whether they are implemented as multiple goroutines or multiple threads.

embano1:
@yanmxa 

I spend some time looking at the `OpenInbound/Send/Receive/Close` functions and suggesting this patch to remove some unneeded code and tidy things up:

```patch
diff --git a/protocol/kafka_confluent/v2/protocol.go b/protocol/kafka_confluent/v2/protocol.go
index 249855c..7c83087 100644
--- a/protocol/kafka_confluent/v2/protocol.go
+++ b/protocol/kafka_confluent/v2/protocol.go
@@ -99,10 +99,8 @@ func (p *Protocol) Send(ctx context.Context, in binding.Message, transformers ..
 
 	p.closerMux.Lock()
 	defer p.closerMux.Unlock()
+	
 	if p.producer.IsClosed() {
-		if p.producerDeliveryChan != nil && !isClosed(p.producerDeliveryChan) {
-			close(p.producerDeliveryChan)
-		}
 		return errors.New("producer is closed")
 	}
 
@@ -177,9 +175,7 @@ func (p *Protocol) OpenInbound(ctx context.Context) error {
 				logger.Errorf("failed to close the consumer: %v", err)
 			}
 		}
-		if p.consumerIncoming != nil && !isClosed(p.producerDeliveryChan) {
-			close(p.consumerIncoming)
-		}
+		close(p.consumerIncoming)
 	}()
 
 	for {
@@ -228,21 +224,14 @@ func (p *Protocol) Close(ctx context.Context) error {
 	p.closerMux.Lock()
 	defer p.closerMux.Unlock()
 
-	if p.consumerCancel != nil {
-		p.consumerCancel()
-	}
-
 	if p.producer != nil && !p.producer.IsClosed() {
 		p.producer.Close()
+		close(p.producerDeliveryChan)
 	}
-	return nil
-}
 
-func isClosed(ch <-chan kafka.Event) bool {
-	select {
-	case <-ch:
-		return true
-	default:
+	if p.consumerCancel != nil {
+		p.consumerCancel()
 	}
-	return false
+
+	return nil
 }
```

embano1:
Let me know if you have questions.

yanmxa:
> Let me know if you have questions.

Thanks! 
Your advice makes the code clearer and no other questions.

embano1:
> I have a slight concern, perhaps unnecessary and we can just ignore it. That is what if the `producer` isn't closed by the `Closer`, that will cause `producerDeliveryChan` not to be closed?

Please add a comment to the `Close()` method that protocol users must call `Close()` to properly clean up resources after use.

yanmxa:
> > I have a slight concern, perhaps unnecessary and we can just ignore it. That is what if the `producer` isn't closed by the `Closer`, that will cause `producerDeliveryChan` not to be closed?
> 
> Please add a comment to the `Close(ctx)` method that protocol users must call `Close(ctx)` to properly clean up resources after use. Also, add `defer p.Close(ctx)` to all your examples

Done!

embano1:
Please add a doc string to the `Close()` method as well, advising users to call `Close()` after use. Please directly squash your commits so we can merge. Also, any related documentation/README updates needed to further highlight this new protocol implementation?

embano1:
Also, why did you not use `WithPollGoroutines` in all `receiver` examples? Should be clear why this is advised e.g., also adding a comment to the example explaining that this won't have perf impact and is only required if the user cares about in-order processing when using the SDK.

yanmxa:
> Also, why did you not use `WithPollGoroutines` in all `receiver` examples? Should be clear why this is advised e.g., also adding a comment to the example explaining that this won't have perf impact and is only required if the user cares about in-order processing when using the SDK.

Thanks for your prompt feedback!

1. Add some comments for the `Close` method and `WithPollGoroutines` in all `receiver` examples. 
2. Update the related document in this repository.
3. Squash the commits

Could you please help me to check if the wording in the above document is appropriate?

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -0,0 +1,237 @@
...
+			return nil, io.EOF
+		}
+		msg := NewMessage(m)
+		return msg, nil
+	case <-ctx.Done():
+		return nil, io.EOF
+	}
+}
+
+// Please use this method to clean up resources after utilizing the protocol
```
embano1, line 222: Go comments must follow specific pattern

```suggestion
// Close cleans up resources after use. Must be called to properly close underlying Kafka resources and avoid resource leaks
```

embano1 (changes requested):
Just that one comment please

embano1:
Can you please check your integration tests? They're not looking positive, but passing: https://github.com/cloudevents/sdk-go/actions/runs/8420863202/job/23056453953?pr=988#step:5:284 (example)

embano1:
This unit test is also not looking ok: https://github.com/cloudevents/sdk-go/actions/runs/8420863203/job/23056453818#step:4:3883

protocol/kafka_confluent/v2/protocol.go
yanmxa: Thanks! https://tip.golang.org/doc/comment

yanmxa:
> This unit test is also not looking ok: https://github.com/cloudevents/sdk-go/actions/runs/8420863203/job/23056453818#step:4:3883

The error log disappears in the unit test now.

yanmxa:
> Can you please check your integration tests? They're not looking positive, but passing: https://github.com/cloudevents/sdk-go/actions/runs/8420863202/job/23056453953?pr=988#step:5:284 (example)

Done!
The integration test error log is from [here](https://github.com/cloudevents/sdk-go/pull/988/files#diff-82e585cd3911942b03502d3dc38e88ede560ecc00b6615ce9d89b25309099095R203).

embano1 (approved):
LGTM

THX @yanmxa for getting this over the line!

cc/ @duglin for final overall check

embano1:
@duglin can you please take a final look just overall whether we missed a README or general requirements for a new protocol binding?

duglin:
@lionelvillard too :-)

.github/workflows/integration.yaml
```diff
@@ -27,6 +27,15 @@ jobs:
           - 9091:9091
           - 9092:9092
 
+      kafka_confluent:
```
duglin, line 30: Should we rename the one above from "kafka" to "kafka_sarama" to be consistent with the rest of the naming?

protocol/kafka_confluent/v2/message.go
duglin: I like being forgiving and just ignoring the patch version number - if we want to do anything different at all. I'm ok with the current choice of only looking for 0.3 and 1.0.

duglin:
LGTM

I'll let @embano1 hit the button.

PR #1007, opened 2024-01-19
Support content types following structured syntax suffixes

### LINKED ISSUE #990: Support structured syntax suffixes for custom content types
We have data coming into our system which can have dynamic content types, and we attempt to handle it regardless of the underlying format. For example we have an application which is feeding us data using a vendor-specific content type, i.e. `application/vnd.my-app+json`, and we'd like to pass this into the cloudevents SDK.

The problem is that the current codec implementation expects an exact match on the content type in order to load up the `Encoder`/`Decoder`, see https://github.com/cloudevents/sdk-go/blob/main/v2/event/datacodec/codec.go

We would like to support the standard set of IANA [Structured Syntax Suffixes](https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml), so if the codec receives a custom content type, but has a `+json`, `+xml` suffix etc., then we use the correct codec.

Happy for one of our team to contribute a PR, just thought I'd open it up for discussion first.

dan-j:
See #990, this adds support for encoding/decoding cloudevent data which is defined by a content-type following the [Structured Syntax Suffixes](https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml) standard.

This enables the encoding/decoding of custom content types which use the `+json` or `+xml` suffix to indicate that their encoding is JSON or XML respectively. Users can also add their own suffixes to the registry in a similar fashion to how `datacodec.AddDecoder()` and `dataodec.AddEncoder()` work, but via the `AddStructuredSuffixDecoder` and `AddStructuredSuffixEncoder` functions.

Fixes #990

### DIFF
--- a/v2/event/datacodec/codec.go
+++ b/v2/event/datacodec/codec.go
@@ -8,6 +8,7 @@ package datacodec
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/cloudevents/sdk-go/v2/event/datacodec/json"
 	"github.com/cloudevents/sdk-go/v2/event/datacodec/text"
@@ -26,9 +27,20 @@ type Encoder func(ctx context.Context, in interface{}) ([]byte, error)
 var decoder map[string]Decoder
 var encoder map[string]Encoder
 
+// ssDecoder is a map of content-type structured suffixes as defined in
+// [Structured Syntax Suffixes](https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml),
+// which may be used to match content types such as application/vnd.custom-app+json
+var ssDecoder map[string]Decoder
+
+// ssEncoder is a map of content-type structured suffixes similar to ssDecoder.
+var ssEncoder map[string]Encoder
+
 func init() {
 	decoder = make(map[string]Decoder, 10)
+	ssDecoder = make(map[string]Decoder, 10)
+
 	encoder = make(map[string]Encoder, 10)
+	ssEncoder = make(map[string]Encoder, 10)
 
 	AddDecoder("", json.Decode)
 	AddDecoder("application/json", json.Decode)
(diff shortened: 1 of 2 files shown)


duglin (approved):
/LGTM

ping @lionelvillard @embano1

yanmxa:
Hi @dan-j! 
I'm a little bit confused by the naming `AddStructuredSuffixDecoder` or `AddStructuredSuffixEncoder` with the `structured`(or `binary`) mode based on the [content-type format](https://github.com/cloudevents/sdk-go/blob/0a3095dbbd1db5c6bbd6a039c927dcac876d4f1b/v2/binding/format/format.go#L27). 

So what's the relationship between the Add**Structured**SuffixDecoder and the **structured** mode? It seems they are describing two kinds of different things, if so, I suggest using another name that won't cause any ambiguity to register the encoder and decoder.

Thanks!

dan-j:
Hi @yanmxa, the `datacodec` package isn't concerned with how an event is encoded (whether structured/binary mode), which is what I think you're referring to?

The `datacodec` package is concerned with marshalling/unmarshalling an event's `data` attribute, and is determined by the event's `datacontenttype` attribute. So for example if you have a `cloudevent.Event` where `datacontenttype=application/vnd.github+json` like so, it would use the github.com/cloudevents/sdk-go/v2/event/datacodec/json package to unmarshal.

```go
var e cloudevent.Event

var target MyStruct
_ = e.DataAs(&target)
```

"Structure Syntax Suffix" is the formal name for this technique, as documented here: https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml. We could rename the functions `AddStructuredSyntaxSuffixEncoder/Decoder` but I feel this wouldn't really help the confusion, it's just a longer name. 

I've also tried to explain this in the GoDoc for the functions, if you have a way to better word this then I'm happy for some feedback.

yanmxa:
> Hi @yanmxa, the `datacodec` package isn't concerned with how an event is encoded (whether structured/binary mode), which is what I think you're referring to?
> 
> The `datacodec` package is concerned with marshalling/unmarshalling an event's `data` attribute, and is determined by the event's `datacontenttype` attribute. So for example if you have a `cloudevent.Event` where `datacontenttype=application/vnd.github+json` like so, it would use the github.com/cloudevents/sdk-go/v2/event/datacodec/json package to unmarshal.
> 
> ```go
> var e cloudevent.Event
> 
> var target MyStruct
> _ = e.DataAs(&target)
> ```
> 
> "Structure Syntax Suffix" is the formal name for this technique, as documented here: https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml. We could rename the functions `AddStructuredSyntaxSuffixEncoder/Decoder` but I feel this wouldn't really help the confusion, it's just a longer name.
> 
> I've also tried to explain this in the GoDoc for the functions, if you have a way to better word this then I'm happy for some feedback.

Since "Structure Syntax Suffix" is a conventional formal name for it. Using any other name seems to break the original meaning, and I agree with you that we can clarify it in the comments on the methods.

@dan-j Thank you for clearing up my doubts!

PR #1017, opened 2024-02-17
Replace time.Time.String() with time.Time.Format()

### LINKED ISSUE #987: Properly marshal event time to RFC-3339 format
In event_marshal.go, `time.String()` is currently used to marshal the time to a string: https://github.com/cloudevents/sdk-go/blob/ba92a5c9a1e2508d855930049999ccd75636f0cb/v2/event/event_marshal.go#L76-L79

However, in the go `time.Time` documentation, it says about the method `time.Time.String()`:
> The returned string is meant for debugging; for a stable serialized representation, use t.MarshalText, t.MarshalBinary, or t.Format with an explicit format string. 

In the Spec, it is required that the time is serialized to RFC-3339, so we should be using `time.Time.MarshalText` instead

MaryamTaj:
Resolves #987 
Earlier, we used time.Time.String() to marshal time into a string. According to the go time.Time documentation however, time.Time.String() is only suitable for debugging purposes. This PR replaces time.Time.String with time.Time.Format() with RFC3339 to provide a stable serialized representation of time.

### DIFF
--- a/test/integration/http/direct_v1_test.go
+++ b/test/integration/http/direct_v1_test.go
@@ -13,7 +13,6 @@ import (
 	"time"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
-	"github.com/cloudevents/sdk-go/v2/types"
 )
 
 func TestSenderReceiver_binary_v1(t *testing.T) {
@@ -101,7 +100,6 @@ func TestSenderReceiver_binary_v1(t *testing.T) {
 
 func TestSenderReceiver_structured_v1(t *testing.T) {
 	now := time.Now()
-
 	testCases := DirectTapTestCases{
 		"Structured v1.0": {
 			now: now,
@@ -131,7 +129,7 @@ func TestSenderReceiver_structured_v1(t *testing.T) {
 				Header: map[string][]string{
 					"content-type": {"application/cloudevents+json"},
 				},
-				Body:          fmt.Sprintf(`{"data":{"hello":"unittest"},"id":"ABC-123","source":"/unit/test/client","specversion":"1.0","subject":"resource","time":%q,"type":"unit.test.client.sent"}`, types.FormatTime(now.UTC())),
+				Body:          fmt.Sprintf(`{"data":{"hello":"unittest"},"id":"ABC-123","source":"/unit/test/client","specversion":"1.0","subject":"resource","time":%q,"type":"unit.test.client.sent"}`, now.Format(time.RFC3339)),
 				ContentLength: 182,
 			},
 		},
@@ -177,7 +175,7 @@ func TestSenderReceiver_data_base64_v1(t *testing.T) {
 				Header: map[string][]string{
 					"content-type": {"application/cloudevents+json"},
 				},
-				Body:          fmt.Sprintf(`{"data_base64":"aGVsbG86IHVuaXR0ZXN0","id":"ABC-123","source":"/unit/test/client","specversion":"1.0","subject":"resource","time":%q,"type":"unit.test.client.sent"}`, now.UTC().Format(time.RFC3339Nano)),
+				Body:          fmt.Sprintf(`{"data_base64":"aGVsbG86IHVuaXR0ZXN0","id":"ABC-123","source":"/unit/test/client","specversion":"1.0","subject":"resource","time":%q,"type":"unit.test.client.sent"}`, now.Format(time.RFC3339)),
 				ContentLength: 191,
 			},
 		},
--- a/v2/event/event_marshal.go
+++ b/v2/event/event_marshal.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"slices"
 	"strings"
+	"time"
 
 	jsoniter "github.com/json-iterator/go"
 )
@@ -77,7 +78,7 @@ func WriteJson(in *Event, writer io.Writer) error {
 		if eventContext.Time != nil {
 			stream.WriteMore()
 			stream.WriteObjectField("time")
-			stream.WriteString(eventContext.Time.String())
+			stream.WriteString(eventContext.Time.Format(time.RFC3339Nano))
 		}
 	case *EventContextV1:
 		// Set a bunch of variables we need later
@@ -121,7 +122,7 @@ func WriteJson(in *Event, writer io.Writer) error {
 		if eventContext.Time != nil {
 			stream.WriteMore()
 			stream.WriteObjectField("time")
-			stream.WriteString(eventContext.Time.String())
+			stream.WriteString(eventContext.Time.Format(time.RFC3339Nano))
 		}
 	default:
 		return fmt.Errorf("missing event context")


.github/workflows/draft-release.yaml
```diff
@@ -6,6 +6,7 @@ on:
     - 'protocol/**' # Push events to matching protocol/foo/bar/v*, i.e. protocol/foo/bar/v1.0, protocol/foo/bar/v20.15.10
     - 'observability/**' # Push events to matching observability/foo/bar/v*
     - 'sql/**' # Push events to matching sql/v*
+    - 'binding/**' # Push events to matching binding/foo/bar/v*
```
embano1, line 9: nit: should be separate PR

README.md
```diff
@@ -27,7 +27,7 @@ _Note:_ Supported go version: 1.14+
 Add the module as dependency using go mod:
 
 ```shell
-go get github.com/cloudevents/sdk-go/v2@v2.4.1
+go get github.com/cloudevents/sdk-go/v2@v2.5.0
```
embano1, line 30: nit: should be separate PR (incl. the other ones)

binding/format/protobuf/v2/protobuf.go
```diff
@@ -1,3 +1,8 @@
+/*
+ Copyright 2021 The CloudEvents Authors
+ SPDX-License-Identifier: Apache-2.0
+*/
+
```
embano1, line 5: nit: should be separate PR

v2/event/event_marshal.go
```diff
@@ -76,7 +76,7 @@ func WriteJson(in *Event, writer io.Writer) error {
 		if eventContext.Time != nil {
 			stream.WriteMore()
 			stream.WriteObjectField("time")
-			stream.WriteString(eventContext.Time.String())
+			stream.WriteString(eventContext.Time.Format(time.RFC3339))
```
embano1, line 79: I ran an example w/ Go 1.22 (same w/ Go 1.21) here: https://go.dev/play/p/NcMTqPu75fe

And I get:

```
2009-11-10 23:00:00 +0000 UTC m=+0.000000001
2009-11-10T23:00:00Z
```

Which could be considered a breaking change. WDYT?

cc/ @duglin

duglin: I'm ok with the breaking change. I'm more bothered that:
1 - it wasn't spec compliant before and no one noticed
2 - not a single testcase needed to be changed

I agree with splitting the upgrade stuff into a separate PR. But can we also get at least one testcase added that checks the format of the `time` attribute and that it adheres to the CE spec (RFC3339) ?

MaryamTaj: I'll get to work on it! Thank you so much for the feedback!

MaryamTaj: Cleaned up the PR. Currently working on adding test cases:)

embano1:
@MaryamTaj converted to draft, seems you're still working on this. Once ready for review, please just mark as "Ready for review" and ping us.

duglin:
@MaryamTaj any update on this?

MaryamTaj:
@duglin Hi Doug! I was running into some issues earlier. I am going to try a new approach, and I'll let you know how it goes

duglin:
ok thanks for the update.

duglin:
@MaryamTaj any update on this?

MaryamTaj:
@duglin I think I'm running into a wall. I tried to resolve as many errors as I could, but I do not think I can further proceed with this pull request.
However, I have outlined what I tried below:
https://docs.google.com/document/d/1xfac9mAtsvB8RvqKN-Fz4F96W-VfdtivWGZitTi_QZg/edit?usp=sharing

duglin:
@MaryamTaj I push a commit - see what you think.
Also, I think you need to sign your commit(s). Might want to squash 'em.
Did you add a testcase that checks the actual syntax of the "time" field?

duglin:
ping @MaryamTaj

embano1:
Glad it's sorted out! Before we merge, can you please squash/clean up commits into one?

test/integration/http/direct_v1_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"time"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
-	"github.com/cloudevents/sdk-go/v2/types"
+	// "github.com/cloudevents/sdk-go/v2/types"
```
embano1, line 16: why comment out and not remove?

embano1:
@duglin do we consider this PR a breaking change for users? I understand our code was broken, so this would be a fix - but still could be an issue impacting downstream users?

MaryamTaj:
> Glad it's sorted out! Before we merge, can you please squash/clean up commits into one?

duglin:
Yes it is a breaking change. Some options:
1 - merge it and see what happens :-)
2 - add a config option to allow people to turn it on if they wish, but no one will do so, or even know about it
3 - add a config option to get the current (wrong) behavior and only those who are broken will need to turn it on - if they don't want to be spec compliant going forward, but at least they have an immediate "fix" until they can "do the right thing"

I like 3.

Also, I think we need a test that checks the serialization of the timestamp IS actually the right format, not just that the sent and received values look the same. Unless I'm missing a test that already covers this.

MaryamTaj:
I signed off and squashed all the commits into one. I also went through the codebase. Would v2/types/timestamp_test.go check the serialization of the timestamp is actually the right format?

duglin:
> I signed off and squashed all the commits into one. I also went through the codebase. Would v2/types/timestamp_test.go check the serialization of the timestamp is actually the right format?

I'm guessing no since you didn't need to change any test in there as part of this PR. At least some test should have failed as a result of your change... hence I think we need a new test that would pass w/o your change in v2/event/event_marshal.go but would fail with it.... then you can "fix" the test to fail w/o your change, and pass with it.

embano1:
Yes, the existing tests might not detect it because if the producer is updated with the new semantics, the consumer will see automatically the new behavior. A test showing a producer with the old semantics and a consumer with the new behavior should fail with this change IIUC.

I also like option (3), but strictly speaking it requires a MAJOR version change in our semver scheme which doesn't feel justified. And my concern is that at the scale this SDK is used, consumers might not detect it right away?

v2/event/event_marshal.go
```diff
@@ -120,7 +121,7 @@ func WriteJson(in *Event, writer io.Writer) error {
 		if eventContext.Time != nil {
 			stream.WriteMore()
 			stream.WriteObjectField("time")
-			stream.WriteString(eventContext.Time.String())
+			stream.WriteString(eventContext.Time.Format(time.RFC3339Nano))
```
embano1, line 124: Why are we using Nano here, and the tests don't?

duglin: If I'm remembering correctly, it was because there were some tests that needed more precision than just "seconds".

duglin:
@MaryamTaj and @embano1 for review/merge

embano1:
@duglin I don't see agreement on how we want to handle the breaking change, only suggestions. Seems like we want to merge and point out in the release notes?

duglin:
Oops - I forgot about that.  I'm still leaning toward 3 (pull the band-aide off but give people a way to get the old behavior, at least for a while).

I'll work on adding a flag to give the old behavior, but won't add it to the PR until you agree with that direction.

embano1:
I think we should add it, yes. So we need:

- a flag/env var
- tests
- docs

duglin:
While trying to write a testcase for the new flag I'm unable to reproduce any error. I always see the right time format. I'm starting to think this isn't an issue. See: https://github.com/cloudevents/sdk-go/issues/987#issuecomment-2745937686 and the next comment.

duglin:
PR no longer needed, we already serialize things correctly

PR #1018, opened 2024-02-20
Adding a simple dataref extension

### LINKED ISSUE #198: Implement Dataref (Claim Check Pattern)
Ref: https://github.com/cloudevents/spec/blob/master/extensions/dataref.md

matzew:
Inspired by the java sdk, adding a go implementation  for `dataref` extension

Fixes #198

### DIFF
--- a/v2/extensions/dataref_extension.go
+++ b/v2/extensions/dataref_extension.go
@@ -1,0 +1,41 @@
+/*
+ Copyright 2024 The CloudEvents Authors
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+package extensions
+
+import (
+	"github.com/cloudevents/sdk-go/v2/event"
+	"net/url"
+)
+
+const DataRefExtensionKey = "dataref"
+
+// DataRefExtension represents the CloudEvents Dataref (claim check pattern)
+// extension for cloudevents contexts,
+// See https://github.com/cloudevents/spec/blob/main/cloudevents/extensions/dataref.md
+// for more info
+type DataRefExtension struct {
+	DataRef string `json:"dataref"`
+}
+
+// AddDataRefExtension adds the dataref attribute to the cloudevents context
+func AddDataRefExtension(e *event.Event, dataRef string) error {
+	if _, err := url.Parse(dataRef); err != nil {
+		return err
+	}
+	e.SetExtension(DataRefExtensionKey, dataRef)
+	return nil
+}
+
+// GetDataRefExtension returns any dataref attribute present in the
+// cloudevent event/context and a bool to indicate if it was found.
+// If not found, the DataRefExtension.DataRef value will be ""
+func GetDataRefExtension(e event.Event) (DataRefExtension, bool) {
+	if dataRefValue, ok := e.Extensions()[DataRefExtensionKey]; ok {
+		dataRefStr, _ := dataRefValue.(string)
+		return DataRefExtension{DataRef: dataRefStr}, true
+	}
+	return DataRefExtension{}, false
+}
--- a/v2/extensions/dataref_extension_test.go
+++ b/v2/extensions/dataref_extension_test.go
@@ -1,0 +1,69 @@
+/*
+ Copyright 2024 The CloudEvents Authors
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+package extensions
+
+import (
+	"testing"
+
+	"github.com/cloudevents/sdk-go/v2/event"
+)
+
+func TestAddDataRefExtension(t *testing.T) {
+	// MUST HAVE AT LEAST ONE FAILING TEST as that'll also test to make sure
+	// that the failed add() call won't actually set anything
+
+	tests := []struct {
+		dataref string
+		pass    bool
+	}{
+		{"https://example.com/data", true},
+		{"://invalid-url", false},
+	}
+
+	for _, test := range tests {
+		e := event.New()
+		err := AddDataRefExtension(&e, test.dataref)
+
+		// Make sure adding it passed/fails appropriately
+		if test.pass && err != nil {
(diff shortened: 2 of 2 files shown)


v2/extensions/dataref_extension.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"github.com/cloudevents/sdk-go/v2/event"
+	"net/url"
+)
+
+const DataRefExtensionKey = "dataref"
+
+type DataRefExtension struct {
+	DataRef string `json:"dataref"`
```
duglin, line 16: Not that it matters much but I'm curious why it's a struct instead of just a string? Are you leaving the option open that one day we may need to include extra metadata along with the URL?

duglin (approved):
LGTM

v2/extensions/dataref_extension_test.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func TestGetDataRefExtensionNotFound(t *testing.T) {
+	e := event.New()
+
+	_, ok := GetDataRefExtension(e)
+	if ok {
+		t.Fatal("Expected not to find DataRefExtension, but did")
+	}
+}
```
embano1, line 41: Looks like Java unit tests :) 

nit: can we use table-driven tests so those tests are easier to modify/add later? Example here: https://github.com/cloudevents/sdk-go/blob/4cc6c2d62d63da00f5b00efd4147e7dd5e40ee4c/v2/protocol/http/protocol_retry_test.go#L25

embano1: Just noticed you might be following the distributed tracing implementation which uses same style...whatever you prefer

matzew: good point

duglin:
rebase needed

v2/extensions/dataref_extension_test.go
duglin: @matzew I'd like to merge this PR - did you want to update something based on this comment?

duglin: ping @matzew

duglin:
Added ref to #198 so it'll close it once we merge

v2/extensions/dataref_extension_test.go
```diff
@@ -0,0 +1,44 @@
...
+			t.Fatalf("Expected not to find DataRefExtension (%s), but did",
+				test.dataref)
+		}
+	}
+}
+
+func TestGetDataRefExtensionNotFound(t *testing.T) {
+	e := event.New()
+
+	_, ok := GetDataRefExtension(e)
```
embano1, line 40: nit: missing positive case getting extension

duglin: doi!

duglin:
Added more tests.
I changed the getter to return "nil" instead of an empty struct when it's not set. I think checking for nil is more appropriate, more go-ish.

v2/extensions/dataref_extension.go
```diff
@@ -0,0 +1,36 @@
...
+
+func GetDataRefExtension(e event.Event) (*DataRefExtension, bool) {
+	if dataRefValue, ok := e.Extensions()[DataRefExtensionKey]; ok {
+		dataRefStr, ok := dataRefValue.(string)
+		if !ok {
+			return nil, false
+		}
+		return &DataRefExtension{DataRef: dataRefStr}, true
+	}
+	return nil, false
```
embano1, line 35: Sorry Doug for this nit, but let's keep the value semantics here because using pointer can lead to (ownership) surprises (without checking the rest of the code, I assume we mostly use value semantics for core CE fields).

embano1: Typically (convention?), if an error is returned, the value should not be used (we can add a comment to make it even more explicit). Alternatively, If needed, users can check for the zero value with https://pkg.go.dev/reflect#Value.IsZero.

v2/extensions/dataref_extension_test.go
```diff
@@ -0,0 +1,69 @@
...
+		if test.pass {
+			if !ok {
+				t.Fatalf("Dataref (%s) is missing after being set",
+					test.dataref)
+			}
+			if dr == nil || dr.DataRef != test.dataref {
+				t.Fatalf("Retrieved dataref(%v) doesn't match set value(%s)",
+					dr, test.dataref)
+			}
+		} else {
```
embano1, line 52: nit: I think it's safe to remove the `else` and directly jump to the next `if` without changing test semantics?

v2/extensions/dataref_extension.go
duglin: this is actually one of the reasons I prefer to return `nil` - I find the entire Zero value stuff annoying when `nil` is much more clear .  If we return a Zero struct then it's too easy for someone to think the dataref is set but with "" - which is a valid URL. But I see at least one of the other extensions does it without a pointer so I'll just hold my nose and switch it back.

v2/extensions/dataref_extension_test.go
duglin: I think we need it because we only want to fail when the dataref is set on the "failed" case, not when the pass is expected to pass.

duglin:
@embano1 updated

v2/extensions/dataref_extension.go
embano1: IIRC, Go is moving to making it easier to understand zero values (recently they did it with `time` for example) because `nil` has so many edges, most importantly surprising side effects for SDK users due to ownership/pointer semantics.

nit: we should add comments to these new public methods.

duglin: done

embano1 (approved):
LGTM

PR #1027, opened 2024-03-16
add support for publishSettings

### LINKED ISSUE #1026: Expose PublishSettings like already done for ReceiveSettings
## Motivation
The google pubsub client allows you to configure your receive and publish settings. via [pubsub.ReceiveSettings](https://pkg.go.dev/cloud.google.com/go/pubsub#ReceiveSettings) and [pubsub.PublishSettings](https://pkg.go.dev/cloud.google.com/go/pubsub#PublishSettings). This sdk exposes the [ReceiveSettings](https://github.com/cloudevents/sdk-go/blob/ef734d06ddf5da92fe111a399a8b4f852af83681/protocol/pubsub/v2/protocol.go#L37) in its interface via the protocol Struct but does not give the same treatment to PublishSettings. I propose we correct this omission with a backwards compatible change exposing this and thus allowing the user to configure things like batch publishing of messages without effecting the existing behaviour of the client.

JamesBLewis:
Resolves #1026

Update Connection struct to allow for customising PublishSettings the same way ReceiveSettings can be customised.


duglin:
rebase needed

protocol/pubsub/v2/internal/connection_test.go
```diff
@@ -161,6 +161,60 @@ func TestPublishCreateTopic(t *testing.T) {
...
+		defer client.Close()
+
+		psconn := &Connection{
+			AllowCreateSubscription: true,
+			AllowCreateTopic:        true,
+			Client:                  client,
+			ProjectID:               projectID,
+			TopicID:                 topicID,
+			SubscriptionID:          subID,
+			PublishSettings: &pubsub.PublishSettings{
```
duglin, line 186: How is topic deletion controlled by this setting?

JamesBLewis: Hi @duglin my apologies, your confusion is a result of me copying and modifying one of the existing tests. I have corrected the test now.

Issue #1030, opened 2024-04-01
Memory leak in the confluent kafka producer


yanmxa:
@embano1 I just found a memory leak in the confluent kafka protocol `Sender`. I apologize for overlooking this issue in the initial version.

The producer should fetch the event from the `producer.Events()` chan. Otherwise, the events chan will eventually fill up. 

Reference: https://github.com/confluentinc/confluent-kafka-go/issues/1043, https://github.com/confluentinc/confluent-kafka-go/issues/578

PR #1031, opened 2024-04-01
Provide report channel `Events()` for confluent kafka producer

yanmxa:
Resolved: https://github.com/cloudevents/sdk-go/issues/1030

### DIFF
--- a/protocol/kafka_confluent/v2/option.go
+++ b/protocol/kafka_confluent/v2/option.go
@@ -15,7 +15,7 @@ import (
 // Option is the function signature required to be considered an kafka_confluent.Option.
 type Option func(*Protocol) error
 
-// WithConfigMap sets the configMap to init the kafka client. This option is not required.
+// WithConfigMap sets the configMap to init the kafka client.
 func WithConfigMap(config *kafka.ConfigMap) Option {
 	return func(p *Protocol) error {
 		if config == nil {
@@ -26,7 +26,7 @@ func WithConfigMap(config *kafka.ConfigMap) Option {
 	}
 }
 
-// WithSenderTopic sets the defaultTopic for the kafka.Producer. This option is not required.
+// WithSenderTopic sets the defaultTopic for the kafka.Producer.
 func WithSenderTopic(defaultTopic string) Option {
 	return func(p *Protocol) error {
 		if defaultTopic == "" {
@@ -37,7 +37,7 @@ func WithSenderTopic(defaultTopic string) Option {
 	}
 }
 
-// WithReceiverTopics sets the topics for the kafka.Consumer. This option is not required.
+// WithReceiverTopics sets the topics for the kafka.Consumer.
 func WithReceiverTopics(topics []string) Option {
 	return func(p *Protocol) error {
 		if topics == nil {
@@ -48,7 +48,7 @@ func WithReceiverTopics(topics []string) Option {
 	}
 }
 
-// WithRebalanceCallBack sets the callback for rebalancing of the consumer group. This option is not required.
+// WithRebalanceCallBack sets the callback for rebalancing of the consumer group.
 func WithRebalanceCallBack(rebalanceCb kafka.RebalanceCb) Option {
 	return func(p *Protocol) error {
 		if rebalanceCb == nil {
@@ -59,7 +59,7 @@ func WithRebalanceCallBack(rebalanceCb kafka.RebalanceCb) Option {
 	}
 }
 
-// WithPollTimeout sets timeout of the consumer polling for message or events, return nil on timeout. This option is not required.
+// WithPollTimeout sets timeout of the consumer polling for message or events, return nil on timeout.
 func WithPollTimeout(timeoutMs int) Option {
 	return func(p *Protocol) error {
 		p.consumerPollTimeout = timeoutMs
@@ -67,7 +67,7 @@ func WithPollTimeout(timeoutMs int) Option {
 	}
 }
 
-// WithSender set a kafka.Producer instance to init the client directly. This option is not required.
+// WithSender set a kafka.Producer instance to init the client directly.
 func WithSender(producer *kafka.Producer) Option {
 	return func(p *Protocol) error {
 		if producer == nil {
@@ -78,7 +78,7 @@ func WithSender(producer *kafka.Producer) Option {
 	}
 }
 
-// WithErrorHandler provide a func on how to handle the kafka.Error which the kafka.Consumer has polled. This option is not required.
+// WithErrorHandler provide a func on how to handle the kafka.Error which the kafka.Consumer has polled.
 func WithErrorHandler(handler func(ctx context.Context, err kafka.Error)) Option {
 	return func(p *Protocol) error {
 		p.consumerErrorHandler = handler
@@ -86,7 +86,7 @@ func WithErrorHandler(handler func(ctx context.Context, err kafka.Error)) Option
 	}
 }
 
-// WithSender set a kafka.Consumer instance to init the client directly. This option is not required.
+// WithReceiver set a kafka.Consumer instance to init the client directly.
 func WithReceiver(consumer *kafka.Consumer) Option {
 	return func(p *Protocol) error {
 		if consumer == nil {
@@ -97,12 +97,12 @@ func WithReceiver(consumer *kafka.Consumer) Option {
 	}
 }
 
-// Opaque key type used to store topicPartitionOffsets: assign them from ctx. This option is not required.
+// Opaque key type used to store topicPartitionOffsets: assign them from ctx.
 type topicPartitionOffsetsType struct{}
 
 var offsetKey = topicPartitionOffsetsType{}
 
-// WithTopicPartitionOffsets will set the positions where the consumer starts consuming from. This option is not required.
+// WithTopicPartitionOffsets will set the positions where the consumer starts consuming from.
 func WithTopicPartitionOffsets(ctx context.Context, topicPartitionOffsets []kafka.TopicPartition) context.Context {
 	if len(topicPartitionOffsets) == 0 {
 		panic("the topicPartitionOffsets cannot be empty")
--- a/protocol/kafka_confluent/v2/protocol.go
+++ b/protocol/kafka_confluent/v2/protocol.go
@@ -33,15 +33,14 @@ type Protocol struct {
 	consumerTopics       []string
 	consumerRebalanceCb  kafka.RebalanceCb                          // optional
 	consumerPollTimeout  int                                        // optional
-	consumerErrorHandler func(ctx context.Context, err kafka.Error) //optional
+	consumerErrorHandler func(ctx context.Context, err kafka.Error) // optional
 	consumerMux          sync.Mutex
 	consumerIncoming     chan *kafka.Message
 	consumerCtx          context.Context
 	consumerCancel       context.CancelFunc
 
 	producer             *kafka.Producer
-	producerDeliveryChan chan kafka.Event // optional
-	producerDefaultTopic string           // optional
+	producerDefaultTopic string // optional
 
 	closerMux sync.Mutex
 }
@@ -85,10 +84,16 @@ func New(opts ...Option) (*Protocol, error) {
 	if p.kafkaConfigMap == nil && p.producer == nil && p.consumer == nil {
 		return nil, errors.New("at least one of the following to initialize the protocol must be set: config, producer, or consumer")
 	}
-	if p.producer != nil {
-		p.producerDeliveryChan = make(chan kafka.Event)
-	}
 	return p, nil
+}
+
+// Events returns the events channel used by Confluent Kafka to deliver the result from a produce, i.e., send, operation.
+// When using this SDK to produce (send) messages, this channel must be monitored to avoid resource leaks and this channel becoming full. See Confluent SDK for Go for details on the implementation.
+func (p *Protocol) Events() (chan kafka.Event, error) {
+	if p.producer == nil {
+		return nil, errors.New("producer not set")
+	}
+	return p.producer.Events(), nil
 }
 
 func (p *Protocol) applyOptions(opts ...Option) error {
@@ -100,6 +105,7 @@ func (p *Protocol) applyOptions(opts ...Option) error {
 	return nil
 }
 
+// Send message by kafka.Producer. You must monitor the Events() channel when using this function.
 func (p *Protocol) Send(ctx context.Context, in binding.Message, transformers ...binding.Transformer) (err error) {
 	if p.producer == nil {
 		return errors.New("producer client must be set")
@@ -128,19 +134,12 @@ func (p *Protocol) Send(ctx context.Context, in binding.Message, transformers ..
 		kafkaMsg.Key = []byte(messageKey)
 	}
 
-	err = WriteProducerMessage(ctx, in, kafkaMsg, transformers...)
-	if err != nil {
-		return err
+	if err = WriteProducerMessage(ctx, in, kafkaMsg, transformers...); err != nil {
+		return fmt.Errorf("create producer message: %w", err)
 	}
 
-	err = p.producer.Produce(kafkaMsg, p.producerDeliveryChan)
-	if err != nil {
-		return err
-	}
-	e := <-p.producerDeliveryChan
-	m := e.(*kafka.Message)
-	if m.TopicPartition.Error != nil {
-		return m.TopicPartition.Error
+	if err = p.producer.Produce(kafkaMsg, nil); err != nil {
+		return fmt.Errorf("produce message: %w", err)
 	}
 	return nil
 }
@@ -231,15 +230,18 @@ func (p *Protocol) Receive(ctx context.Context) (binding.Message, error) {
 func (p *Protocol) Close(ctx context.Context) error {
 	p.closerMux.Lock()
 	defer p.closerMux.Unlock()
+	logger := cecontext.LoggerFrom(ctx)
 
 	if p.consumerCancel != nil {
 		p.consumerCancel()
 	}
 
 	if p.producer != nil && !p.producer.IsClosed() {
+		// Flush and close the producer with a 10 seconds timeout (closes Events channel)
+		for p.producer.Flush(10000) > 0 {
+			logger.Info("Flushing outstanding messages")
+		}
 		p.producer.Close()
-		close(p.producerDeliveryChan)
 	}
-
 	return nil
 }
--- a/samples/kafka_confluent/sender/main.go
+++ b/samples/kafka_confluent/sender/main.go
@@ -8,6 +8,7 @@ package main
 import (
 	"context"
 	"log"
+	"sync"
 
 	confluent "github.com/cloudevents/sdk-go/protocol/kafka_confluent/v2"
 	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
@@ -26,8 +27,49 @@ func main() {
 	sender, err := confluent.New(confluent.WithConfigMap(&kafka.ConfigMap{
 		"bootstrap.servers": "127.0.0.1:9092",
 	}), confluent.WithSenderTopic(topic))
+	if err != nil {
+		log.Fatalf("failed to create protocol, %v", err)
+	}
 
-	defer sender.Close(ctx)
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	// Listen to all the events on the default events channel
+	// It's important to read these events otherwise the events channel will eventually fill up
+	go func() {
+		defer wg.Done()
+		eventChan, err := sender.Events()
+		if err != nil {
+			log.Fatalf("failed to get events channel for sender, %v", err)
+		}
+		for e := range eventChan {
+			switch ev := e.(type) {
+			case *kafka.Message:
+				// The message delivery report, indicating success or
+				// permanent failure after retries have been exhausted.
+				// Application level retries won't help since the client
+				// is already configured to do that.
+				m := ev
+				if m.TopicPartition.Error != nil {
+					log.Printf("Delivery failed: %v\n", m.TopicPartition.Error)
+				} else {
+					log.Printf("Delivered message to topic %s [%d] at offset %v\n",
+						*m.TopicPartition.Topic, m.TopicPartition.Partition, m.TopicPartition.Offset)
+				}
+			case kafka.Error:
+				// Generic client instance-level errors, such as
+				// broker connection failures, authentication issues, etc.
+				//
+				// These errors should generally be considered informational
+				// as the underlying client will automatically try to
+				// recover from any errors encountered, the application
+				// does not need to take action on them.
+				log.Printf("Error: %v\n", ev)
+			default:
+				log.Printf("Ignored event: %v\n", ev)
+			}
+		}
+	}()
 
 	c, err := cloudevents.NewClient(sender, cloudevents.WithTimeNow(), cloudevents.WithUUIDs())
 	if err != nil {
@@ -43,14 +85,13 @@ func main() {
 			"message": "Hello, World!",
 		})
 
-		if result := c.Send(
-			// Set the producer message key
-			confluent.WithMessageKey(ctx, e.ID()),
-			e,
-		); cloudevents.IsUndelivered(result) {
+		if result := c.Send(confluent.WithMessageKey(ctx, e.ID()), e); cloudevents.IsUndelivered(result) {
 			log.Printf("failed to send: %v", result)
 		} else {
 			log.Printf("sent: %d, accepted: %t", i, cloudevents.IsACK(result))
 		}
 	}
+
+	sender.Close(ctx)
+	wg.Wait()
 }
--- a/samples/kafka_confluent/receiver/main.go
+++ b/samples/kafka_confluent/receiver/main.go
@@ -33,8 +33,9 @@ func main() {
 	}
 	defer receiver.Close(ctx)
 
-	// Setting the 'client.WithPollGoroutines(1)' to make sure the events from kafka partition are processed in order
-	c, err := cloudevents.NewClient(receiver, client.WithPollGoroutines(1))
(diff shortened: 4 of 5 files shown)


protocol/kafka_confluent/v2/option.go
```diff
@@ -78,6 +78,14 @@ func WithSender(producer *kafka.Producer) Option {
 	}
 }
 
+// WithEventHandler provide a func on how to handle the kafka.Event for the producer.Events() chan. This option is not required.
```
embano1, line 81: nit: can you please remove all the "This option is not required" suffixes from the options? Missed that in the other review. Options are never required.

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -87,6 +90,9 @@ func New(opts ...Option) (*Protocol, error) {
 	}
 	if p.producer != nil {
 		p.producerDeliveryChan = make(chan kafka.Event)
+		if p.producerEventHandler == nil {
+			p.producerEventHandler = func(ctx context.Context, err kafka.Event) {}
```
embano1, line 94: Please provide a useful default implementation using logger to log errors and handling the various cases simmilar to the example shown in the Confluent issue.

embano1 (commented):
I was just starting the review until I noticed we are already passing an error channel to Produce which according to Confluent docs is an option to not have to poll Events() channel and should not lead to memory leak. Do you see memory leaks in our implementation?

> It is also possible to direct delivery reports to alternate channels by providing a non-nil `chan Event` channel to `.Produce()`

https://docs.confluent.io/platform/current/clients/confluent-kafka-go/index.html

yanmxa:
> I was just starting the review until I noticed we are already passing an error channel to Produce which according to Confluent docs is an option to not have to poll Events() channel and should not lead to memory leak. Do you see memory leaks in our implementation?
> 
> > It is also possible to direct delivery reports to alternate channels by providing a non-nil `chan Event` channel to `.Produce()`
> 
> https://docs.confluent.io/platform/current/clients/confluent-kafka-go/index.html

After running a service with our current implementation for a few days, The memory usage is from 75 MB to about 180 MB
image
After profiling the memory through pprof, find it mainly caused by the func `_Cfunc_GoBytes` and the related issues: https://github.com/confluentinc/confluent-kafka-go/issues/578#issuecomment-931642924, 

image

protocol/kafka_confluent/v2/protocol.go
yanmxa: Done.

protocol/kafka_confluent/v2/option.go
yanmxa: Done.

embano1:
Sorry for being a bit pedantic here, but since we follow Confluent's producer logic and pass an event channel, either something is fundamentally wrong with the Confluent library or we are not using it correctly. I don't see adding another goroutine to handle this solving the problem IF the Confluent library behaves as expected, allowing a return channel on produce which we use accordingly: https://github.com/cloudevents/sdk-go/blob/e6a74efbacbf4ac70fc8ed598e678f94b6be7a6d/protocol/kafka_confluent/v2/protocol.go#L136

Can you please debug further whether that channel is not drained correctly and whether you see events also in `Event()` which should not be the case then?

embano1:
> Producing is an asynchronous operation so the client notifies the application of per-message produce success or failure through something called delivery reports. Delivery reports are by default emitted on the `.Events()` channel as `*kafka.Message` and you should check `msg.TopicPartition.Error` for `nil` to find out if the message was succesfully delivered or not. It is also possible to direct delivery reports to alternate channels by providing a non-nil `chan Event` channel to `.Produce()`. If no delivery reports are wanted they can be completely disabled by setting configuration property `"go.delivery.reports": false`.

Btw (unrelated to this, but still important): we aren't calling `Flush()`, are we?

> When you are done producing messages you will need to make sure all messages are indeed delivered to the broker (or failed), remember that this is an asynchronous client so some of your messages may be lingering in internal channels or tranmission queues. To do this you can either keep track of the messages you've produced and wait for their corresponding delivery reports, or call the convenience function `.Flush()` that will block until all message deliveries are done or the provided timeout elapses.

yanmxa:
> Sorry for being a bit pedantic here, but since we follow Confluent's producer logic and pass an event channel, either something is fundamentally wrong with the Confluent library or we are not using it correctly. I don't see adding another goroutine to handle this solving the problem IF the Confluent library behaves as expected, allowing a return channel on produce which we use accordingly:
> 
> https://github.com/cloudevents/sdk-go/blob/e6a74efbacbf4ac70fc8ed598e678f94b6be7a6d/protocol/kafka_confluent/v2/protocol.go#L136
> 
> Can you please debug further whether that channel is not drained correctly and whether you see events also in `Event()` which should not be the case then?

@embano1 You are getting the point. It does seem wired to use `deliveryChan` and `producer.Events()` at the same time. Let's hold this PR. I need to get more information to understand why then doing this in the [sample](https://github.com/confluentinc/confluent-kafka-go/blob/master/examples/producer_custom_channel_example/producer_custom_channel_example.go)

yanmxa:
> > Producing is an asynchronous operation so the client notifies the application of per-message produce success or failure through something called delivery reports. Delivery reports are by default emitted on the `.Events()` channel as `*kafka.Message` and you should check `msg.TopicPartition.Error` for `nil` to find out if the message was succesfully delivered or not. **It is also possible to direct delivery reports to alternate channels by providing a non-nil `chan Event` channel to `.Produce()`**. If no delivery reports are wanted they can be completely disabled by setting configuration property `"go.delivery.reports": false`.
> 
> Btw (unrelated to this, but still important): we aren't calling `Flush()`, are we?
> 
> > When you are done producing messages you will need to make sure all messages are indeed delivered to the broker (or failed), remember that this is an asynchronous client so some of your messages may be lingering in internal channels or tranmission queues. To do this you can either keep track of the messages you've produced and wait for their corresponding delivery reports, or call the convenience function `.Flush()` that will block until all message deliveries are done or the provided timeout elapses.

Yes, We don't need to call the `Flush()`. Cause each time we send a message, then wait directly for the result from the `deliveryChan`.

embano1 (commented):
> Yes, We don't need to call the `Flush()`. Cause each time we send a message, then wait directly for the result from the `deliveryChan`. 

Makes sense.

However, I wonder what happens if a user constructs the producer with the following option:

>  If no delivery reports are wanted they can be completely disabled by setting configuration property `"go.delivery.reports": false`.

Would the producer block indefinitely or panic on `nil` event? Can you please test this and make sure your `Send` still works?

yanmxa:
> > Yes, We don't need to call the `Flush()`. Cause each time we send a message, then wait directly for the result from the `deliveryChan`.
> 
> Makes sense.
> 
> However, I wonder what happens if a user constructs the producer with the following option:
> 
> > If no delivery reports are wanted they can be completely disabled by setting configuration property `"go.delivery.reports": false`.
> 
> Would the producer block indefinitely or panic on `nil` event? Can you please test this and make sure your `Send` still works?

Sure! I verified the following cases:

**1. What happens when set both `producer.Events()` chan and the customized `derliveryChan`?**
  - Start a goroutine fetch the event from `producer.Events()` chan
  - Set a customized `derliveryChan`. 
  - Result: The producer only can get events from `derliveryChan` after producing the message.

**2. What happens with the configuration `"go.delivery.reports": false` based on the current implementation?**
  - Set the `"go.delivery.reports": false` when initialize the cloudevents client.
  - Set a customized `derliveryChan` when sending the message(current implmentation)
  - Result: The sender still works and we still can get the event from the `derliveryChan`. The configuration has no impact on the current implementation!
  
**3. What happens with the configuration `"go.delivery.reports": false` and the `producer.Events()` chan?**
  - Start a goroutine fetch the event from `producer.Events()` chan
  - Set the `"go.delivery.reports": false` when create the kafka producer
  - Result: It can't get anything from the `producer.Events()` chan, after sending a message without the `deliveryChan`, that is `producer.Produce(msg, nil)` 

I also found that this issue, https://github.com/cloudevents/sdk-go/issues/1030 is not necessarily related to our current implementation, so I am planning to close the issue and the current PR. 
But I will keep tracing why the [confluent sample](https://github.com/confluentinc/confluent-kafka-go/blob/master/examples/producer_custom_channel_example/producer_custom_channel_example.go) uses the `producer.Events()` and `deliveryChan` at the same time. And once the conclusion is reached, I will update it here.

@embano1 What do you think?

embano1:
> What happens when set both producer.Events() chan and the customized derliveryChan?

That result is what we are expecting, thanks.

> What happens with the configuration "go.delivery.reports": false based on the current implementation?

Interesting, seems custom event delivery channel overwrites this setting then. Weird configuration behavior, but in-line with the underlying Kafka callback implementation so perhaps reasonable.

> What happens with the configuration "go.delivery.reports": false and the producer.Events() chan?

OK, so that setting works then.

Thx for your investigations, proposed next steps SGTM!

yanmxa:
@embano1 
I posted an issue in the community, looks like the `producer.Events()` is also provided to handle some events not associated with the sending message. https://github.com/confluentinc/confluent-kafka-go/issues/1163

embano1:
@yanmxa thx for verifying! So I suggest implementing a `select {}` pattern on both channels then. Assuming `producer.Events()` only fires if there's an error which is not related to the provided send/produce call channel, we should be good not leaking resources, i.e., there should never be an event on the other channel we pass to produce/send and we can close it on return.

yanmxa:
> @yanmxa thx for verifying! So I suggest implementing a `select {}` pattern on both channels then. 

@embano1 I'm worried that switching `deliverChan` to `select{}` pattern would make the [Sender](https://github.com/cloudevents/sdk-go/blob/e6a74efbacbf4ac70fc8ed598e678f94b6be7a6d/v2/protocol/outbound.go#L16-L27) an asynchronous method, which means it will always return success immediately?

> Assuming `producer.Events()` only fires if there's an error which is not related to the provided send/produce call channel, we should be good not leaking resources, i.e., there should never be an event on the other channel we pass to produce/send and we can close it on return.

Yes, We [release the `deliveryChan`](https://github.com/cloudevents/sdk-go/pull/1031/files#diff-82e585cd3911942b03502d3dc38e88ede560ecc00b6615ce9d89b25309099095R132) passed to the producer/send once the producer is done. 

Is the current implementation consistent with your ideas?

embano1:
Select blocks, ie, is not making the code async. Let me see if I can propose something to simplify the current PR.

embano1:
Quick question: what was the reason we decided to use a custom event delivery channel instead of reading everything from producer.Events()?

yanmxa:
> Quick question: what was the reason we decided to use a custom event delivery channel instead of reading everything from producer.Events()?

I use "delieryChan" to get the result after each send so that we can change the asynchronous method `Produce` into a synchronous method to adapt to the `Sender`. 

I choose `deliveryChan` instead of `producer.Events()`. Because I don't find any official examples that use `producer.Events()` chan to block and wait for the result of produce/send. They all [use `producer.Events()` in a separate gourontine](https://github.com/search?q=repo%3Aconfluentinc%2Fconfluent-kafka-go%20p.Events()&type=code). But it does provide [many cases](https://github.com/search?q=repo%3Aconfluentinc%2Fconfluent-kafka-go%20%3C-deliveryChan&type=code) where produce is blocked and waiting for a result to be sent using `deliveryChan`.

yanmxa:
Hello, @embano1!

I'd like to consult your opinion on something related to [kubeCon China](https://events.linuxfoundation.org/kubecon-cloudnativecon-open-source-summit-ai-dev-china/program/call-for-proposal/#submit-your-proposal-chinese). Since I couldn't find your contact information, I'm reaching out to you here. 

Over the past year, we've brought some updates about the message queue to this repos. If you're interested, I can draft a proposal for the enhancements of cloudevents message queue to the [KubeCon China](https://events.linuxfoundation.org/kubecon-cloudnativecon-open-source-summit-ai-dev-china/program/call-for-proposal/#submit-your-proposal-chinese). The main people involved would be the two of us. The main topics covered include bringing MQTT binding to support IoT cases, and some improvements related to Kafka such as supporting message confirm, consumption from a specific point, batch subscribing to topics, and supporting ordered partition offsets, etc.

How do you feel about it?

yanmxa:
> If we agree on the above, needs code comments for public methods and usage example

It's a good choice to give the `producer.Events()` to the user！

If we deprecate the `deliveryChan`, then we need to think of another question:  when to invoke the `producer.Flush()` to deliver all the buffered messages to the Kafka broker? invoke it in `Sender` or `Closer`?  I agree with the above once this question is settled.

> * When you are done producing messages you will need to make sure all messages are indeed delivered to the broker (or failed), remember that this is an asynchronous client so some of your messages may be lingering in internal channels or tranmission queues. To do this you can either keep track of the messages you've produced and wait for their corresponding delivery reports, or call the convenience function `.Flush()` that will block until all message deliveries are done or the provided timeout elapses.

embano1:
> If we deprecate the `deliveryChan`, then we need to think of another question: when to invoke the `producer.Flush()` to deliver all the buffered messages to the Kafka broker? invoke it in `Sender` or `Closer`? I agree with the above once this question is settled.

Do we need formal deprecation of `deliveryChan` since we don't expose it? Or do you just mean removing that code? I thought about `Sender` calling `Flush` too, but that might be too much (perf). So `Close()` might be the better option since it would also release `Events()` eventually (would need to check the Confluent SDK on the behavior again).

embano1:
> I'd like to consult your opinion on something related to [kubeCon China](https://events.linuxfoundation.org/kubecon-cloudnativecon-open-source-summit-ai-dev-china/program/call-for-proposal/#submit-your-proposal-chinese). Since I couldn't find your contact information, I'm reaching out to you here.

Can you please reach out on the CloudEvents Slack channel and ping me there?

yanmxa:
> > If we deprecate the `deliveryChan`, then we need to think of another question: when to invoke the `producer.Flush()` to deliver all the buffered messages to the Kafka broker? invoke it in `Sender` or `Closer`? I agree with the above once this question is settled.
> 
> Do we need formal deprecation of `deliveryChan` since we don't expose it? Or do you just mean removing that code? I thought about `Sender` calling `Flush` too, but that might be too much (perf). So `Close()` might be the better option since it would also release `Events()` eventually (would need to check the Confluent SDK on the behavior again).

I think we can just remove the code cause use cannot sense it. 

Yes. Calling `Flush()` in `Sender` definitely affects the performance. We will do it in `Closer`

It looks like we have reached a consensus. Then I will proceed with these changes and add some examples.

embano1 (changes requested):

protocol/kafka_confluent/v2/option.go
```diff
@@ -78,15 +78,23 @@ func WithSender(producer *kafka.Producer) Option {
...
+// WithErrorHandler provide a func on how to handle the kafka.Error which the kafka.Consumer has polled.
 func WithErrorHandler(handler func(ctx context.Context, err kafka.Error)) Option {
 	return func(p *Protocol) error {
 		p.consumerErrorHandler = handler
 		return nil
 	}
 }
 
-// WithSender set a kafka.Consumer instance to init the client directly. This option is not required.
+// WithSender set a kafka.Consumer instance to init the client directly.
```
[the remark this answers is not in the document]
yanmxa: Done.

samples/kafka_confluent/sender/main.go
```diff
@@ -43,11 +83,16 @@ func main() {
...
-			// Set the producer message key
-			confluent.WithMessageKey(ctx, e.ID()),
-			e,
-		); cloudevents.IsUndelivered(result) {
+		result := c.Send(confluent.WithMessageKey(ctx, e.ID()), e)
+		if result != nil {
+			if val, ok := result.(kafka.Error); ok && val.Code() == kafka.ErrQueueFull {
+				// Producer queue is full, wait 1s for messages
+				// to be delivered then try again.
+				time.Sleep(time.Second)
```
embano1, line 91: Is this a best practice? Wondering if it complicates the example, unless you think it's always important to implement this logic.

samples/kafka_confluent/sender/main.go
```diff
@@ -43,11 +83,16 @@ func main() {
...
+				continue
+			}
+		}
+		if cloudevents.IsUndelivered(result) {
```
embano1, line 95: this is probably sufficient for error handling instead of the above?

samples/kafka_confluent/sender/main.go
```diff
@@ -26,9 +28,47 @@ func main() {
...
+				// Generic client instance-level errors, such as
+				// broker connection failures, authentication issues, etc.
+				//
+				// These errors should generally be considered informational
+				// as the underlying client will automatically try to
+				// recover from any errors encountered, the application
+				// does not need to take action on them.
+				log.Printf(" Error: %v\n", ev)
+			default:
+				log.Printf("Ignored event: %s\n", ev)
```
embano1, line 67:
```suggestion
				log.Printf("Ignored event: %v\n", ev)
```

samples/kafka_confluent/sender/main.go
```diff
@@ -26,9 +28,47 @@ func main() {
...
 	}), confluent.WithSenderTopic(topic))
-
+	if err != nil {
+		log.Fatalf("failed to create protocol, %v", err)
+	}
 	defer sender.Close(ctx)
 
+	// Listen to all the events on the default events channel
+	// It's important to read these events otherwise the events channel will eventually fill up
+	go func() {
```
embano1, line 38: nit: you're not waiting for this goroutine to finish before returning from `main`- which could prevent your goroutine from printing anything. typically use `https://pkg.go.dev/golang.org/x/sync/errgroup`

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -128,20 +133,13 @@ func (p *Protocol) Send(ctx context.Context, in binding.Message, transformers ..
...
-	if err != nil {
-		return err
+	if err = WriteProducerMessage(ctx, in, kafkaMsg, transformers...); err != nil {
+		return fmt.Errorf("create producer message: %w", err)
 	}
 
-	err = p.producer.Produce(kafkaMsg, p.producerDeliveryChan)
-	if err != nil {
+	if err = p.producer.Produce(kafkaMsg, nil); err != nil {
 		return err
```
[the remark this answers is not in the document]
yanmxa: I wanted to return a `kafka.Error` for fear that users might also want to check its type using an assert statement like [this](https://github.com/confluentinc/confluent-kafka-go/blob/master/examples/producer_example/producer_example.go#L90-L98).

samples/kafka_confluent/sender/main.go
yanmxa: I've seen them use assert like the above when sending messages asynchronously.

But I tried to set the producer buffer channel size with 1 `"go.produce.channel.size": 1`, and sending 10,000 messages within 1 second several times. It never catches the `kafka.ErrQueueFull` error. So let's remove it.

samples/kafka_confluent/sender/main.go
yanmxa: emm... remove it now.

samples/kafka_confluent/sender/main.go
yanmxa: Done

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -85,12 +84,18 @@ func New(opts ...Option) (*Protocol, error) {
...
 
+// Events returns the Events channel (read) to dlivery the event of per-message produce success or failure.
+// Must be monitered for the sender/producer, otherwise the events channel will eventually fill up
+func (p *Protocol) Events() (chan kafka.Event, error) {
+	if p.producer == nil {
+		return nil, errors.New("producer not set")
+	}
+	return p.producer.Events(), nil
+}
+
```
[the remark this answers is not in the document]
yanmxa: Done.

yanmxa:
> Thx @yanmxa ! Definitely easier code on our side - however, that Confluent SDK is really hard to use given all the asynchrony and error handling one has to perform to catch all possible scenarios - guess not much further we can do here :/

Absolutely! It seems like they traded ease of use for improved performance with their asynchronous approach. That's made us put in extra work to handle it.

I'm thinking we might consider adding an option in the future to switch the producer to a synchronous method, avoiding users having to watch `producer.Events()` channel additionally.

embano1:
> I'm thinking we might consider adding an option in the future to switch the producer to a synchronous method, avoiding users having to watch `producer.Events()` channel additionally. Like the following snippet:
> 
> ```go
> err = producer.Produce(msg, nil)
> if err != nil {
>     return fmt.Errorf("failed to produce message: %w", err)
> }
> event := <-producer.Events()
> ```

Initially I tried that, but IMHO it's not that easy. Because `Produce` to Kafka is async, you don't know which event you're getting/returning here, making it problematic for callers to understand whether the returned event/error was for the same CloudEvent just sent with `Send()`. We'd need additional correlation logic and also handle errors not caused be a specific event, but also returned on this channel (as I understood the Confluent docs).

protocol/kafka_confluent/v2/protocol.go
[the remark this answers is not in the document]
embano1: That's why I'm wrapping the error with `%w` to allow Go idiomatic use of https://pkg.go.dev/errors#As

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -231,15 +230,18 @@ func (p *Protocol) Receive(ctx context.Context) (binding.Message, error) {
...
+	logger := cecontext.LoggerFrom(ctx)
 
 	if p.consumerCancel != nil {
 		p.consumerCancel()
 	}
 
 	if p.producer != nil && !p.producer.IsClosed() {
+		// Flush and close the producer and the events channel
+		for p.producer.Flush(10000) > 0 {
+			logger.Info("Still waiting to flush outstanding messages")
```
embano1, line 242:
```suggestion
			logger.Info("Flushing outstanding messages")
```

Question: `10000` is 10 seconds timeout or delay? How often is this message printed? Risk of floding logs?

protocol/kafka_confluent/v2/protocol.go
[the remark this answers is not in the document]
embano1: @yanmxa WDYT?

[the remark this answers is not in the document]
yanmxa: While the type of [`kafka.Error`](https://github.com/confluentinc/confluent-kafka-go/blob/cb2701e8b6f43be0cefb22da5c56c6856f7f0899/kafka/error.go#L35-L41) is not the type of error we typically refer to. 
It's an object that only implements the [`error` interface](https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/error.go#L88-L92). So once the `kafka.Error` is formatted with `%w`, then it can't be unwrapped back to `kafka.Error`, like: `interface conversion: error is *fmt.wrapError, not kafka.Error`

yanmxa:
> > I'm thinking we might consider adding an option in the future to switch the producer to a synchronous method, avoiding users having to watch `producer.Events()` channel additionally. Like the following snippet:
> > ```go
> > err = producer.Produce(msg, nil)
> > if err != nil {
> >     return fmt.Errorf("failed to produce message: %w", err)
> > }
> > event := <-producer.Events()
> > ```
> 
> Initially I tried that, but IMHO it's not that easy. Because `Produce` to Kafka is async, you don't know which event you're getting/returning here, making it problematic for callers to understand whether the returned event/error was for the same CloudEvent just sent with `Send()`. We'd need additional correlation logic and also handle errors not caused be a specific event, but also returned on this channel (as I understood the Confluent docs).

Exactly! Making the `Producer` synchronous also has 3 options currently:
1. Using `producer.Events()` just like the above snippet. And like you said need additional logic to handle the errors not associated with the produced events.
2. Using the customized `deliveryChan` to handle the produced events, and watch `producer.Events()` channel in a separate goroutine to handle other errors 
3. Using `producer. Flush()` after sending the message, which also is discussed.

However, should we convert it to synchronous? Which option should we use? If it is necessary, we can discuss it later.

protocol/kafka_confluent/v2/protocol.go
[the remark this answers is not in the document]
embano1: Perhaps we're talking past each other :)

This doesn't work?

```go
                 var kafkaError kafka.Error
                 if errors.As(err, &kafkaError) {
			// use kafkaError to handle and access fields
		}
```

embano1:
> Which option should we use? If it is necessary, we can discuss it later.

Yeah, let's keep this out of this PR for now

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -100,6 +105,7 @@ func (p *Protocol) applyOptions(opts ...Option) error {
 	return nil
 }
 
+// Send message by kafka.Producer. You must monitor the Events() channel while using this function.
```
embano1, line 108:
```suggestion
// Send message by kafka.Producer. You must monitor the Events() channel when using this function.
```

protocol/kafka_confluent/v2/protocol.go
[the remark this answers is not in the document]
embano1: @yanmxa ICYMI ^

protocol/kafka_confluent/v2/protocol.go
```diff
@@ -231,15 +230,18 @@ func (p *Protocol) Receive(ctx context.Context) (binding.Message, error) {
...
 	p.closerMux.Lock()
 	defer p.closerMux.Unlock()
+	logger := cecontext.LoggerFrom(ctx)
 
 	if p.consumerCancel != nil {
 		p.consumerCancel()
 	}
 
 	if p.producer != nil && !p.producer.IsClosed() {
+		// Flush and close the producer and the events channel
```
embano1, line 240:
```suggestion
		// Flush and close the producer with a 10 seconds timeout (closes Events channel)
```

embano1 (approved):

Issue #1032, opened 2024-04-05
Race condition in func WithCustomAttributes(ctx context.Context, attrs map[string]string) 


volkerhess2:
The WithCustomAttributes function currently takes a map as input value. Hence map is a pointer value, the pointer is passed to msg.Attributes of the pubsub message and then is used throughout the sdk. The map is modified in multiple places. If the pointer to the map is used outside of the event or if the map is reused for multiple events there will be race conditions. 

https://github.com/cloudevents/sdk-go/blob/main/protocol/pubsub/v2/attributes.go
```
type withCustomAttributes struct{}

func AttributesFrom(ctx context.Context) map[string]string {
	return binding.GetOrDefaultFromCtx(ctx, withCustomAttributes{}, make(map[string]string)).(map[string]string)
}

// WithCustomAttributes sets Message Attributes without any CloudEvent logic.
// Note that this function is not intended for CloudEvent Extensions or any `ce-`-prefixed Attributes.
// For these please see `Event` and `Event.SetExtension`.
func WithCustomAttributes(ctx context.Context, attrs map[string]string) context.Context {
	return context.WithValue(ctx, withCustomAttributes{}, attrs)
}
```

https://github.com/cloudevents/sdk-go/blob/main/protocol/pubsub/v2/protocol.go
```
// Send implements Sender.Send
func (t *Protocol) Send(ctx context.Context, in binding.Message, transformers ...binding.Transformer) error {
	var err error
	defer func() { _ = in.Finish(err) }()

	topic := cecontext.TopicFrom(ctx)
	if topic == "" {
		topic = t.topicID
	}

	conn := t.getOrCreateConnection(ctx, topic, "", "")

	msg := &pubsub.Message{
		Attributes: AttributesFrom(ctx),
	}
...

```

PR #1033, opened 2024-04-05
Ensure that editing fetched Attributes is race-free in any case.

AndreasBergmeier6176:


AndreasBergmeier6176:
Fixes https://github.com/cloudevents/sdk-go/issues/1032

embano1:
Thx for the PR! Definitely a possibility for races there. However, your code is technically still not race free because during your copy operation of the map, the map could still be modified. IMHO the best option here would be to add a comment on the functions that the passed map is not supposed to be modified concurrently/not protected during concurrent operations. So callers can pass a copy of the map before using `WithCustomAttributes`. A related question: do you see the sdk-go code concurrently using and modifying this map in our code base today? Or perhaps done so in the pubsub SDK where these attributes are used? Then we definitely need to fix that on our end to protect the map.

AndreasBergmeier6176:
> Thx for the PR! Definitely a possibility for races there. However, your code is technically still not race free because during your copy operation of the map, the map could still be modified.

FWIU, that would then be a race problem introduced by the call site - AKA not our problem.

> IMHO the best option here would be to add a comment on the functions that the passed map is not supposed to be modified concurrently/not protected during concurrent operations. So callers can pass a copy of the map before using `WithCustomAttributes`.

Well I think that is common sense: "As long as a call is processing, don't modify the arguments".

And actually this PR is necessary because the implementation violated that common sense by storing the map (pointer).
Thus this PR is restoring the common sense usage of the functions.

> A related question: do you see the sdk-go code concurrently using and modifying this map in our code base today?

Nope. Then again it is perhaps not by the lack of sdk-go code trying. I mean there is a lot of modifying maps between functions and ownership/concurrency was not very clear to me on first glance.

protocol/pubsub/v2/attributes.go
```diff
@@ -14,12 +14,34 @@ import (
...
+	if attrs != nil {
+		// Since it is likely that the map gets used in another goroutine
+		// ensure that modifying passed in map is race-free.
+		cp := make(map[string]string, len(attrs))
+		for k, v := range attrs {
+			cp[k] = v
+		}
+		attrs = cp
+	}
 	return context.WithValue(ctx, withCustomAttributes{}, attrs)
```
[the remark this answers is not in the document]
AndreasBergmeier6176: Sorry, I am not following. No-op instead of what?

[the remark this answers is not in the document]
embano1: Well, right now this code means the context key `withCustomAttributes{}` is assigned `nil` if `attrs` is `nil`. So I was wondering why not just `return` without modifying the context instead of setting it to `nil` (you set the value from `attrs` which in `nil` case is in fact `nil`).

Would still work with your code above, just save a context operation.

embano1, line 46:
```suggestion
	return
```

AndreasBergmeier6176: > Well, right now this code means the context key withCustomAttributes{} is assigned nil if attrs is nil. So I was wondering why not just return without modifying the context instead of setting it to nil (you set the value from attrs which in nil case is in fact nil).

This then removes the ability to reset the Attributes saved on the Context. Might not be necessary, IDK.
That being said IF we should do that we def. need to document it because it no longer is a naive/straight-forward implementation.

embano1: Fair point. However, resetting might not be needed since this is all using context logic which could easily be done by passing a new/fresh (root) context without this key. So I wonder whether we really need the possibility to reset the key? Especially since `nil` is not really valid as attributes, which you work around in your other function.

We could document that passing `nil` as attributes is a no-op and does not modify the context?
Again, only unless `nil` is an explicitly allowed/useful value for PubSub?

duglin:
@AndreasBergmeier6176 what is the status of this one?

duglin:
@AndreasBergmeier6176 did you want address the comments above? Otherwise we may close this one

PR #1034, opened 2024-04-14
fix: support multiple amqp data fields

embano1:
Support multidimensional `data` field in AMQP implementation when converting an AMQP message to a CloudEvent. Writing to AMQP message from CloudEvent remains unchanged because the first slice of `[][]byte` is used to populate the data.

Ref: https://github.com/cloudevents/spec/issues/1275

### DIFF
--- a/protocol/amqp/v2/message.go
+++ b/protocol/amqp/v2/message.go
@@ -50,8 +50,10 @@ func NewMessage(message *amqp.Message, receiver *amqp.Receiver) *Message {
 	return &Message{AMQP: message, AMQPrcv: receiver, format: fmt, version: vn}
 }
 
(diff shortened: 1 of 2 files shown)


embano1:
cc/ @deissnerk

embano1:
cc/ @duglin for review

duglin:
The code change itself looks good w.r.t. the "just concat them" decision.
However, @deissnerk, is there really no way for someone to split the result back into separate data sections? A simple concat w/o any kind of separator seems like it would make it really hard for consumers, no?

embano1:
@deissnerk that's how I understood it as well

duglin:
ok in that case LGTM

duglin (approved):

Issue #1037, opened 2024-04-22
Create CODE OWNERS file for reviewers


embano1:
Ref: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

PR #1038, opened 2024-04-22
chore: add codeowners file

embano1:
Closes: #1037


.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+*       @cloudevents/maintainers
```
duglin, line 1: I don't think we want all CE SDK maintainers to have write access, do we?

embano1: yup, that's what I wanted to discuss. will change 👍

duglin: I mean... do we really trust @deissnerk ? :-)

duglin (approved):

PR #1040, opened 2024-04-24
#1039 Migrate azure/go-amqp to version 1.0.5

### LINKED ISSUE #1039: Support https://github.com/Azure/go-amqp stable version
Hi 😄,

Currently v2 amqp protocol is making use of Azure/go-amqp version 0.17.0. The package is already in version 1.0.5. 
This is causing constraints in upgrading the cloudevents package on our side since we can only do it when cloudevents upgrades it.

Thanks

cx-joses:
Changes:

Fixes #1039 
- migrate dependent packages to the latest version
- migrate azure/go-amqp from 0.17 to 1.0.5


protocol/amqp/v2/receiver.go
```diff
@@ -19,10 +19,13 @@ import (
 const serverDown = "session ended by server"
 
 // receiver wraps an amqp.Receiver as a binding.Receiver
-type receiver struct{ amqp *amqp.Receiver }
+type receiver struct {
+	amqp    *amqp.Receiver
+	options *amqp.ReceiveOptions
```
embano1, line 24: question: do we need pointer semantics here (and in the constructors)?

cx-joses: if you notice the changes on `sender` file require a pointer since the library used assumes that if nil is passed default is assumed. If in the future the library does the same for the receiver it will be prepared to it. WDYT?

embano1:
Thx, I still see the bumps though?

cx-joses:
> Thx, I still see the bumps though?

@embano1 I actually upgraded to 1.21 since its the version referred on build steps and others. do you see any issue in doing so?

cx-joses:
> > @embano1 I actually upgraded to 1.21 since its the version referred on build steps and others. do you see any issue in doing so?
> 
> IMHO we want to keep `1.18` so we don't introduce additional changes/features that might block users compiling with older go versions. The README states:
> 
> > Note: Supported go version: 1.18+

understood, you are absolutely right! reverted it all to 1.18 again! thanks

protocol/amqp/v2/receiver.go
embano1: That is fine, you can always use the copied value with `&var` - but make a copy of the provided struct (to avoid races/bugs) when calling `New...()`

samples/amqp/receiver/main.go
```diff
@@ -30,17 +30,22 @@ func sampleConfig() (server, node string, opts []ceamqp.Option) {
...
+		}
 	}
 	return env, strings.TrimPrefix(u.Path, "/"), opts
 }
 
 func main() {
 	host, node, opts := sampleConfig()
-	p, err := ceamqp.NewProtocol(host, node, []amqp.ConnOption{}, []amqp.SessionOption{}, opts...)
+	p, err := ceamqp.NewProtocol(context.Background(), host, node, opts, &amqp.SessionOptions{},
+		&amqp.SenderOptions{}, &amqp.ReceiverOptions{})
```
embano1, line 48: would be much nicer if those would not be pointers IMHO

protocol/amqp/v2/receiver.go
cx-joses: sure, gone back to not using pointer for this case

samples/amqp/receiver/main.go
cx-joses: made it use the struct instead of pointers. thanks

protocol/amqp/v2/protocol.go
```diff
@@ -35,54 +30,60 @@ type Protocol struct {
 }
 
 // NewProtocolFromClient creates a new amqp transport.
-func NewProtocolFromClient(client *amqp.Client, session *amqp.Session, queue string, opts ...Option) (*Protocol, error) {
+func NewProtocolFromClient(
```
embano1, line 33: question: are we ok with this breaking change on this signature? I understand why we technically might not need options anymore (since those are now structs), but you might want to keep the options pattern to not break this signature and allow for future expansions. You could introduce a new option to set sender/receiver options on the protocol.

embano1: Or are you worried that those two structs are always needed and thus an options pattern doesn't really apply?

protocol/amqp/v2/protocol.go
```diff
@@ -35,54 +30,60 @@ type Protocol struct {
...
 	}
-	t.Receiver = NewReceiver(amqpReceiver).(*receiver)
+	t.Receiver = NewReceiver(amqpReceiver, amqp.ReceiveOptions{}).(*receiver)
 	return t, nil
 }
 
 // NewProtocol creates a new amqp transport.
-func NewProtocol(server, queue string, connOption []amqp.ConnOption, sessionOption []amqp.SessionOption, opts ...Option) (*Protocol, error) {
-	client, err := amqp.Dial(server, connOption...)
+func NewProtocol(
```
embano1, line 66: same question as above

protocol/amqp/v2/sender.go
```diff
@@ -33,15 +34,13 @@ func (s *sender) Send(ctx context.Context, in binding.Message, transformers ...b
...
 	return err
 }
 
 // NewSender creates a new Sender which wraps an amqp.Sender in a binding.Sender
-func NewSender(amqpSender *amqp.Sender, options ...SenderOptionFunc) protocol.Sender {
-	s := &sender{amqp: amqpSender}
-	for _, o := range options {
-		o(s)
-	}
+func NewSender(amqpSender *amqp.Sender, options *amqp.SendOptions) protocol.Sender {
```
embano1, line 42: nit: inconsistent with other functions due to pointer 
```suggestion
func NewSender(amqpSender *amqp.Sender, options amqp.SendOptions) protocol.Sender {
```

samples/amqp/receiver/main.go
```diff
@@ -30,17 +30,22 @@ func sampleConfig() (server, node string, opts []ceamqp.Option) {
...
+		}
 	}
 	return env, strings.TrimPrefix(u.Path, "/"), opts
 }
 
 func main() {
 	host, node, opts := sampleConfig()
-	p, err := ceamqp.NewProtocol(host, node, []amqp.ConnOption{}, []amqp.SessionOption{}, opts...)
+	p, err := ceamqp.NewProtocol(context.Background(), host, node, opts, amqp.SessionOptions{},
+		amqp.SenderOptions{}, amqp.ReceiverOptions{})
```
embano1, line 48: interesting: so we don't need those options, which goes back to my comment on keeping functional options pattern

samples/amqp/sender/main.go
```diff
@@ -51,7 +53,8 @@ type Example struct {
 
 func main() {
 	host, node, opts := sampleConfig()
-	p, err := ceamqp.NewProtocol(host, node, []amqp.ConnOption{}, []amqp.SessionOption{}, opts...)
+	p, err := ceamqp.NewProtocol(context.Background(), host, node, opts, amqp.SessionOptions{}, amqp.SenderOptions{},
```
embano1, line 56: see other comment on options

test/integration/amqp/amqp_test.go
```diff
@@ -70,26 +69,31 @@ func receiverProtocolFactory(t *testing.T) *protocolamqp.Protocol {
...
+	client, err := amqp.Dial(context.Background(), s, &amqp.ConnOptions{})
 	if err != nil {
 		t.Skipf("ampq.Dial(%#v): %v", s, err)
 	}
-	session, err = client.NewSession()
+	session, err = client.NewSession(context.Background(), &amqp.SessionOptions{})
 	require.NoError(t, err)
-	return client, session, addr
+	senderOpts = amqp.SenderOptions{}
+	require.NotNil(t, senderOpts)
```
embano1, line 87: what is this testing?

test/integration/amqp_binding/amqp_test.go
```diff
@@ -90,29 +90,35 @@ func TestSendEventReceiveBinary(t *testing.T) {
...
 	}
-	session, err = client.NewSession()
+	session, err = client.NewSession(context.Background(), &amqp.SessionOptions{})
 	require.NoError(t, err)
-	return client, session, addr
+	senderOpts = &amqp.SenderOptions{}
+	require.NotNil(t, senderOpts)
+	receiverOpts = &amqp.ReceiverOptions{}
+	require.NotNil(t, receiverOpts)
+	return client, session, addr, senderOpts, receiverOpts
```
embano1, line 111: nit: unless needed, don't return pointers but structs

embano1: in the other function you used structs on return

embano1 (changes requested):

mschneider82:
any change that this will be merged soon? using the stable go-amqp version would be nice.

embano1:
@cx-joses are you still down to finish this one?

duglin:
@cx-joses not to put pressure on you :-)  but this was identified as one of the ones we'd like to get merged before we cut a new release...soon-ish.  If you don't think you'll have time to rebase and address Michael's comments soon please let us know

cx-joses:
@duglin what is the timeline you guys are expecting? sorry for the lack of time on this, this is a priority for me also but came in a difficult time! Let's see if I find the time tomorrow to work on this

duglin:
@cx-joses I don't think there's a firm date, just people getting antsy because of all of the new stuff that's gone in and in particular the release of the CESQL v1.0 spec - which the SDK now supports.

cx-joses:
> @cx-joses I don't think there's a firm date, just people getting antsy because of all of the new stuff that's gone in and in particular the release of the CESQL v1.0 spec - which the SDK now supports.

sure, I will make the comments changes today!

protocol/amqp/v2/sender.go
cx-joses: the thing here is that internally there is a default behavior assumed for a nil passing. from there I am assuming a pointer here. is it wrong?

protocol/amqp/v2/protocol.go
```diff
@@ -92,69 +101,76 @@ func NewProtocol(server, queue string, connOption []amqp.ConnOption, sessionOpti
 }
 
 // NewSenderProtocolFromClient creates a new amqp sender transport.
-func NewSenderProtocolFromClient(client *amqp.Client, session *amqp.Session, address string, opts ...Option) (*Protocol, error) {
+func NewSenderProtocolFromClient(
+	ctx context.Context,
+	client *amqp.Conn,
+	session *amqp.Session,
+	address string,
+	opts ...Option,
```
embano1, line 109: seem to be unused?

protocol/amqp/v2/protocol.go
```diff
@@ -92,69 +101,76 @@ func NewProtocol(server, queue string, connOption []amqp.ConnOption, sessionOpti
...
 	}
-	t.Receiver = NewReceiver(amqpReceiver).(*receiver)
+	t.Receiver = NewReceiver(amqpReceiver, t.receiveOpts).(*receiver)
 	return t, nil
 }
 
 // NewSenderProtocol creates a new sender amqp transport.
-func NewSenderProtocol(server, address string, connOption []amqp.ConnOption, sessionOption []amqp.SessionOption, opts ...Option) (*Protocol, error) {
-	client, err := amqp.Dial(server, connOption...)
+func NewSenderProtocol(ctx context.Context, server, address string, connOptions amqp.ConnOptions, sessionOptions amqp.SessionOptions, opts ...Option) (*Protocol, error) {
```
embano1, line 160: `opts` also seems to be unused? (seems this was the case before?)

protocol/amqp/v2/sender.go
embano1: The `Receiver` does not use this though (uses struct semantics). Is this expected? If passing `nil` is valid we

1) either need to document this or
2) use a functional options pattern where we have `nil` as default behavior

protocol/amqp/v2/protocol.go
cx-joses: what do you mean? which property? opts are applied and sent in the sender creation

protocol/amqp/v2/protocol.go
cx-joses: those are used on line 173 for the internal protocol creation

protocol/amqp/v2/options.go
```diff
@@ -12,42 +12,76 @@ import (
 // Option is the function signature required to be considered an amqp.Option.
 type Option func(*Protocol) error
 
-// WithConnOpt sets a connection option for amqp
-func WithConnOpt(opt amqp.ConnOption) Option {
+type SendOption func(sender *sender)
+type ReceiveOption func(receiver *receiver)
+
+// WithConnOpts sets a connection option for amqp
+func WithConnOpts(opt *amqp.ConnOptions) Option {
```
embano1, line 19: Since you're using pointer semantics in those options, be careful with races - users might not be aware that you're passing the pointer around (instead of copy) and this could lead to surprises for users. I suggest using value semantic to avoid races and these surprises.

protocol/amqp/v2/options.go
```diff
@@ -12,42 +12,76 @@ import (
...
+// WithReceiverOpts sets a link option for amqp
+func WithReceiverOpts(opt *amqp.ReceiverOptions) Option {
 	return func(t *Protocol) error {
-		t.receiverLinkOpts = append(t.receiverLinkOpts, opt)
+		t.receiverOpts = opt
 		return nil
 	}
 }
 
+// WithReceiveOpts sets a receive option for amqp
```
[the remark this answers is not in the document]
embano1: IIUC, this is for the sender/receiver construction, which currently does not support options. Do we need this new functionality at the cost of a more complex user API?

duglin:
@cx-joses what's the status of this one?

duglin:
@cx-joses can you rebase and fix the conflicts?

duglin:
@cx-joses are you able, or willing, to rebase this and address the comments?
If not we may close this one.

PR #1046, opened 2024-05-01
fix: LIKE expression with invalid string literals returns a parse error instead of panicking

### LINKED ISSUE #931: Panic if LIKE pattern is not a quoted string
When parsing the following cesql

```
x LIKE 123
```

the cesql parser panics

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x58 pc=0x638552]

goroutine 62 [running]:
testing.tRunner.func1.2({0x680c20, 0xa16cc0})
	D:/programs/go1.17.6/src/testing/testing.go:1209 +0x24e
testing.tRunner.func1()
	D:/programs/go1.17.6/src/testing/testing.go:1212 +0x218
panic({0x680c20, 0xa16cc0})
	D:/programs/go1.17.6/src/runtime/panic.go:1038 +0x215
github.com/cloudevents/sdk-go/sql/v2/parser.(*expressionVisitor).VisitLikeExpression(0xc00068f4a0, 0xc000416350)
	D:/projects/github/sdk-go/sql/v2/parser/expression_visitor.go:179 +0xf2
github.com/cloudevents/sdk-go/sql/v2/parser.(*expressionVisitor).Visit(0xc00068f4b8, {0x775820, 0xc000416350})
	D:/projects/github/sdk-go/sql/v2/parser/expression_visitor.go:75 +0x4a5
github.com/cloudevents/sdk-go/sql/v2/parser.(*expressionVisitor).VisitChildren(0x6d1120, {0x776cc0, 0xc00068f4b8})
	D:/projects/github/sdk-go/sql/v2/parser/expression_visitor.go:83 +0x57
github.com/cloudevents/sdk-go/sql/v2/parser.(*expressionVisitor).VisitCesql(0x608fe0, 0x76fcb8)
	D:/projects/github/sdk-go/sql/v2/parser/expression_visitor.go:296 +0x25
github.com/cloudevents/sdk-go/sql/v2/gen.(*CesqlContext).Accept(0xc000416328, {0x76fcb8, 0xc00068f4a0})
	D:/projects/github/sdk-go/sql/v2/gen/cesqlparser_parser.go:256 +0x87
github.com/cloudevents/sdk-go/sql/v2/parser.(*Parser).Parse(0xc0002ff980, {0xc0002ff980, 0xe})
	D:/projects/github/sdk-go/sql/v2/parser/parser.go:39 +0x2f4
github.com/cloudevents/sdk-go/sql/v2/parser.Parse(...)
	D:/projects/github/sdk-go/sql/v2/parser/parser.go:73
```

Cali0707:
Fixes #931 

When visiting the nodes of the AST while parsing the LIKE expression, it looks like we assumed that we would have a valid string literal. However, it is possible that there is no valid string literal, in which case we have to handle that and return an error rather than continuing to parse (and subsequently panicking)

### DIFF
--- a/sql/v2/test/tck/like_expression.yaml
+++ b/sql/v2/test/tck/like_expression.yaml
@@ -115,4 +115,11 @@ tests:
     result: false
   - name: With type coercion from bool (4)
     expression: "FALSE LIKE 'fal%'"
-    result: true 
+    result: true
+
+  - name: Invalid string literal in comparison causes parse error
+    expression: "x LIKE 123"
+    result: false
+    error: parse
+    eventOverrides:
+      x: "123"
--- a/sql/v2/parser/expression_visitor.go
+++ b/sql/v2/parser/expression_visitor.go
@@ -6,6 +6,7 @@
 package parser
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
@@ -175,9 +176,13 @@ func (v *expressionVisitor) VisitLikeExpression(ctx *gen.LikeExpressionContext) 
 	if patternContext.DQUOTED_STRING_LITERAL() != nil {
 		// Parse double quoted string
 		pattern = dQuotedStringToString(patternContext.DQUOTED_STRING_LITERAL().GetText())
-	} else {
+	} else if patternContext.SQUOTED_STRING_LITERAL() != nil {
 		// Parse single quoted string
 		pattern = sQuotedStringToString(patternContext.SQUOTED_STRING_LITERAL().GetText())
+	} else {
+		// not a string, return an error
+		v.parsingErrors = append(v.parsingErrors, fmt.Errorf("failed to parse LIKE expression: the pattern was not a string literal"))
+		return noopExpression{}
 	}
 
 	likeExpression, err := expression.NewLikeExpression(v.Visit(ctx.Expression()).(cesql.Expression), pattern)


Cali0707:
cc @duglin @pierDipi @lionelvillard

pierDipi (approved):
LGTM

For reference, here's the spec for like-operation [[ref](https://github.com/cloudevents/spec/blob/main/cesql/spec.md#23-operators)]
```
like-operation ::= expression not-operator? like-operator string-literal

string-literal ::= ( "'" ( [^'] | "\'" )* "'" ) | ( '"' ( [^"] | '\"' )* '"')
```

sql/v2/test/tck/like_expression.yaml
```diff
@@ -115,4 +115,11 @@ tests:
     result: false
   - name: With type coercion from bool (4)
     expression: "FALSE LIKE 'fal%'"
-    result: true 
+    result: true
```
duglin, line 118: Just wondering, in cases where the first operand is a string, does the spec mandate that we do a string insensitive compare? I see where it talks about doing so for ceSQL keywords, but not values. Did I miss it? Should we add testcases?

duglin:
Aside from my minor question, LGTM

sql/v2/test/tck/like_expression.yaml
Cali0707: actually, I'm not sure if this is covered in the spec. Perhaps we can open an issue in the spec repo and clarify what _should_ happen?

Cali0707: I imagine if we had a testcase `FALSE LIKE 'FAL%'`, the test would fail - but maybe the spec should clarify whether or not that should actually happen

sql/v2/test/tck/like_expression.yaml
```diff
@@ -115,4 +115,11 @@ tests:
...
+
+  - name: Invalid string literal in comparison causes parse error
+    expression: "x LIKE 123"
+    result: false
+    error: parse
+    eventOverrides:
+      x: "123"
```
pierDipi, line 125: Isn't this file coming from the spec? Should this updated there first https://github.com/cloudevents/spec/blob/main/cesql/cesql_tck/like_expression.yaml ?

Cali0707: Yeah, I just wanted to include the change here since it is catching any regressions to the bug this is fixing. I'll sync this back to the spec once this merges

sql/v2/test/tck/like_expression.yaml
Cali0707: @duglin thinking about this some more, what is happening here is that the `FALSE` value is coerced to a string first, and then the string comparison is made. 

Looking at the spec, I think there are two parts we need to clarify (I'll open a PR in a bit):
1. That the LIKE expression casts the left operand to a `string` before evaluating the comparison
2. That boolean values cast to lower case strings `"true"` and `"false"`

Cali0707:
cc @embano1

sql/v2/test/tck/like_expression.yaml
duglin: I was actually more wondering whether: `"foo" = "FOO"` is meant to return true or false

duglin:
do we have a testcase for this?

sql/v2/test/tck/like_expression.yaml
Cali0707: Oh, I think that would return `false`. My understanding is that the case insensitivity is only for the CESQL keywords, not for the actual values

Cali0707:
> do we have a testcase for this?

@duglin yes, the new tck test in this PR runs and passes with this change, but prior to the change it would panic

duglin:
> > do we have a testcase for this?
> 
> @duglin yes, the new tck test in this PR runs and passes with this change, but prior to the change it would panic

From a "spec complete" perspective, is that what we want? Or is it expected that everyone will use a `lower()` type of built-in function if they need it?  Which is fine, I just wanted to double check

sql/v2/test/tck/like_expression.yaml
duglin: From a "spec complete" perspective, is that what we want? Or is it expected that everyone will use a lower() type of built-in function if they need it? Which is fine, I just wanted to double check

Cali0707: Personally, I think having case sensitive values is easier to understand then the other way around. Normally in SQL I would expect that a string comparison on values would only be true if they match exactly, and would use something like lower() or upper() if I wanted a case insenstive comparison. @pierDipi not sure if you have any other ideas here

duglin: I've been using mySQL for my xRegistry impl and found out that it does case insensitive compares by default. Go figure! I expected the opposite, like you :-)   See: https://makingdatameaningful.com/is-sql-case-sensitive/

I'm ok either way the spec decides to go, but I think it should state its decision explicitly so it's clear. Today, it's kind of implied by the presence of the lower() and upper() funcs.

Cali0707: Ah I've only ever used postgres which is case sensitive by default! I agree, let's clarify in the spec that string comparison is case sensitive instead of just implying it with the LOWER() and UPPER() functions.

lionelvillard (approved):

lionelvillard:
@duglin for final approval/merge

duglin:
LGTM - it needed a rebase so kicked that off...

duglin:
merging...

PR #1050, opened 2024-05-02
feat: added benchmarks for CESQL

### LINKED ISSUE #927: Add benchmarks to CloudEvents SQL
It would be great to add benchmarks to the CloudEvents SQL package, so that we can find any places where performance improvements can be made as well as detect performance regressions. Ideally we can benchmark both:
1. Parse time and memory
2. Expression evaluation time and memory

Cali0707:
Fixes #927 

This PR adds benchmarks to run on every TCK test case that we have, as that seemed like the easiest way of getting a good set of test cases. For each scenario, this benchmarks:
1. The parsing 
2. The evaluation

Since all these cases are already being tested for correctness, this does not bother checking if the results are correct

### DIFF
--- a/sql/v2/test/tck_test.go
+++ b/sql/v2/test/tck_test.go
@@ -6,6 +6,7 @@
 package test
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"path"
@@ -70,7 +71,7 @@ type TckTestCase struct {
 	EventOverrides map[string]interface{} `json:"eventOverrides"`
 }
 
-func (tc TckTestCase) InputEvent(t *testing.T) cloudevents.Event {
+func (tc TckTestCase) InputEvent(tb testing.TB) cloudevents.Event {
 	var inputEvent cloudevents.Event
 	if tc.Event != nil {
 		inputEvent = *tc.Event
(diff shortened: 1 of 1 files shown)


Cali0707:
cc @duglin @pierDipi

duglin:
LGTM but I've never really used the go benchmark stuff before... so a potentially silly question, I assume this will print stats about the testcases (e.g. timing, etc...), but is there a mechanism in all of this that compares the values with the state of the code before whatever PR is being tested? How will people know if a PR impacts performance dramatically w/o the historical info?

Cali0707:
@duglin we tried to think this through in Knative and didn't come to a great conclusion. The main challenge with historical info is that it is really dependent on the machine the code is running on. We found that the github action runners were too noisy for good benchmarking.

So, our solution was to just run the benchmarks once on the `main` branch and then once on whatever branch had the changes when we were making a change that might impact performance, and then there is some tooling to compare those two results. If you can think of a better workflow I would be very interested in trying that!

duglin:
Nope - that seems reasonable. Would be really cool if the PR testing infra could do that on each PR and flag it when things look bad. But that's just a dream :-)

pierDipi (approved):
LGTM

pierDipi:
cc @cloudevents/sdk-go-maintainers

Cali0707:
cc @embano1

duglin:
rebase needed

duglin (approved):

Issue #1047, opened 2024-05-01
Expose AddFunction API for CESQL Parser


Cali0707:
As discussed in #1045 , it would be good to provide end users of the SDK the ability to add custom functions to the CESQL parser and runtime.

I think this would probably start with changes in https://github.com/cloudevents/sdk-go/blob/f97061a7e72b17f63120aa553123edb28c49e490/sql/v2/runtime/functions_resolver.go#L23, but it will likely need changes in other areas as well

Cali0707:
@duglin I believe @dgeorgievski has offered to contribute a PR for this

dgeorgievski:
Just for the record I am willing to work on this PR

duglin:
thanks @dgeorgievski

dgeorgievski:
Hi @duglin, 

Just curious how long it would take to finalize the review of the PR. I hope the code owners are not gone for the summer break :-)

lionelvillard:
I can take a look at it

Cali0707:
I believe this is complete now, closing here...

PR #1051, opened 2024-05-03
#1047 Expose AddFunction API for CESQL Parser

dgeorgievski:
As discussed in #1047 added support for addition of user defined functions.
- AddFunction API
- TCK tests for user defined functions.
- Instructions on using user defined functions in README file.

@Cali0707, waiting on your feedback. The PR has a working version of the implementation, but it might need to be refined especially when it comes to testing, 

I've noticed that the `test/tck_test.go` manages all the testing of SQL expressions and functions. I've also added `runtime/functions_resolver_test.go` to be able to test the new `runtime.AddFunction` function. Is that ok?


Cali0707 (commented):
Hey @dgeorgievski at a first look, these changes seem good! My only thought is to maybe not use the tck tests for this, as those tests are meant to test the CESQL spec and are mirrored from the cloudevents spec repo. Since the spec only says `An engine SHOULD also allow users to define their custom functions`, I think it is better in our case to just have the test cases you added be in their own test file somewhere else. 

I'll take a closer look at the implementation next Monday!

dgeorgievski:
> I think it is better in our case to just have the test cases you added be in their own test file somewhere else.
I agree. Let's decide first on the new location of the tests. My first impulse was to place  them under `runtime/`  considering I am testing `runtime.AddFunction` in the same dir as well.

Cali0707:
Yeah putting them into `runtime` sounds reasonable to me!

dgeorgievski:
- Moved user defined functions tests from `sql/v2/test`  to `sql/v2/runtime`
- Create `runtime_test` package to avoid circular import runtime package errors.
- Using the same TCK mechanism for running user defined functions test to simplify the users validation of new functions. For example, if I'd like to add a new function I'd be able to quickly define both the function and all the required uses cases in the `runtime/tck/user_defined_functions.yaml`.

Running user defined function tests

```
$ go test -v ./runtime
...
--- PASS: Test_functionTable_AddFunction (0.00s)
    --- PASS: Test_functionTable_AddFunction/Add_user_functions_to_global_table (0.00s)
    --- PASS: Test_functionTable_AddFunction/Fail_add_user_functions_to_global_table (0.00s)
...
--- PASS: Test_UserFunctions (0.01s)
    --- PASS: Test_UserFunctions/User_defined_functions (0.00s)
        --- PASS: Test_UserFunctions/User_defined_functions/KONKAT_(2) (0.00s)
        --- PASS: Test_UserFunctions/User_defined_functions/KONKAT_(3) (0.00s)
        --- PASS: Test_UserFunctions/User_defined_functions/HASPREFIX_(1) (0.00s)
        --- PASS: Test_UserFunctions/User_defined_functions/HASPREFIX_(3) (0.00s)
        --- PASS: Test_UserFunctions/User_defined_functions/HASPREFIX_(5) (0.00s)
        --- PASS: Test_UserFunctions/User_defined_functions/HASPREFIX_(4) (0.00s)
        --- PASS: Test_UserFunctions/User_defined_functions/HASPREFIX_(2) (0.00s)
        --- PASS: Test_UserFunctions/User_defined_functions/KONKAT_(1) (0.00s)
PASS
ok      github.com/cloudevents/sdk-go/sql/v2/runtime    0.022s
```

sql/v2/runtime/user_defined_functions_test.go
```diff
@@ -0,0 +1,209 @@
+/*
+ Copyright 2021 The CloudEvents Authors
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+package runtime_test
```
Cali0707, line 6: Since this is in a separate package for import purposes, do you think it would make sense to move it into a new (nested) directory like `/runtime/test` (which would also then have all the tck files in the same directory)>

Cali0707 (commented):
Just one small question about where files go, otherwise LGTM from me!

sql/v2/runtime/user_defined_functions_test.go
dgeorgievski: Yes, I believe creating a dedicated test directory would provide more flexibility for future customizations. Will be right back.

dgeorgievski: Tests assets moved to `runtime/test` dir.

Cali0707 (approved):
LGTM from me

sql/v2/runtime/test/user_defined_functions_test.go
```diff
@@ -0,0 +1,209 @@
+/*
+ Copyright 2021 The CloudEvents Authors
```
lionelvillard, line 2: should be 2024

sql/v2/runtime/test/tck/user_defined_functions.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - name: HASPREFIX (3)
+    expression: "HASPREFIX('abcdef', '')"
+    result: true  
+  - name: HASPREFIX (4)
+    expression: "HASPREFIX('abcdef', 'gh')"
+    result: false  
+  - name: HASPREFIX (5)
+    expression: "HASPREFIX('abcdef', 'abcdefg')"
+    result: false
+
```
lionelvillard, line 18: remove newline?

sql/v2/runtime/test/user_defined_functions_test.go
```diff
@@ -0,0 +1,209 @@
...
+		return int32(tc.Result.(int))
+	case float64:
+		return int32(tc.Result.(float64))
+	case bool:
+		return tc.Result.(bool)
+	}
+	return tc.Result
+}
+
+func Test_functionTable_AddFunction(t *testing.T) {
```
lionelvillard, line 115: We tend to follow the common go test naming convention: TestPascalCase. `TestFunctionTableAddFunction`)

sql/v2/runtime/test/user_defined_functions_test.go
```diff
@@ -0,0 +1,209 @@
...
+			for _, fn := range tt.args.functions {
+				if err := ceruntime.AddFunction(fn); (err != nil) != tt.wantErr {
+					t.Errorf("functionTable.AddFunction() error = %v, wantErr %v", err, tt.wantErr)
+				}
+			}
+		})
+	}
+}
+
+func Test_UserFunctions(t *testing.T) {
```
lionelvillard, line 152: same here: `TestUserFunctions`

sql/v2/runtime/test/tck/user_defined_functions.yaml
dgeorgievski: The existing sql tck yaml files use new line to separate tests cases. 
https://github.com/cloudevents/sdk-go/blob/main/sql/v2/test/tck/string_builtin_functions.yaml#L12-L18

I was following their format. I think the new line between consecutive tests cases should be kept but the new line at the end of the file should be removed.

lionelvillard (approved):

Issue #1058, opened 2024-05-21
BUG: invalid CE prefix in Confluent Kafka binding implementation


embano1:
Ref: https://github.com/cloudevents/spec/pull/1284

PR #1059, opened 2024-05-21
fix: invalid ce prefix in confluent binding

embano1:
Kafka binding spec requires prefix to be `ce_`: https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/kafka-protocol-binding.md#3231-property-names

Closes: #1058


embano1:
cc/ @yanmxa since this will be a breaking (but required) change to fix the bug.

duglin (approved):
/LGTM assuming the tests pass

Issue #1062, opened 2024-05-23
The `content-type` for the MQTT protocol might have been set multiple times


yanmxa:
When setting the event transfer with structured mode by MQTT protocol. The `content-type` might be set more than one times, see the MQTT payload below.
- Sender with structured-mode
```go
...
ctx = cloudevents.WithEncodingStructured(ctx)
if result := c.Send(
	cecontext.WithTopic(ctx, "test-topic"),
	e,
); cloudevents.IsUndelivered(result) {
	log.Printf("failed to send: %v", result)
} else {
	log.Printf("sent: %d, accepted: %t", i, cloudevents.IsACK(result))
}
```
- The MQTT Payload
```json
{
  "PacketID": 0,
  "QoS": 0,
  "Retain": false,
  "Topic": "test-topic",
  "Properties": {
    "CorrelationData": null,
    "ContentType": "",
    "ResponseTopic": "",
    "PayloadFormat": null,
    "MessageExpiry": null,
    "SubscriptionIdentifier": null,
    "TopicAlias": null,
    "User": [
      {
        "Key": "Content-Type",
        "Value": "application/cloudevents+json"
      },
      {
        "Key": "Content-Type",
        "Value": "application/cloudevents+json"
      },
      {
        "Key": "Content-Type",
        "Value": "application/cloudevents+json"
      },
      {
        "Key": "Content-Type",
        "Value": "application/cloudevents+json"
      }
    ]
  },
  "Payload": "eyJzcGVjdmVyc2lvbiI6IjEuMCIsImlkIjoiNTVkM2NhYTktOTA2ZC00MDE0LWI3YTYtZDUyOTA3NjAxZTE2Iiwic291cmNlIjoiaHR0cHM6Ly9naXRodWIuY29tL2Nsb3VkZXZlbnRzL3Nkay1nby9zYW1wbGVzL21xdHQvc2VuZGVyIiwidHlwZSI6ImNvbS5jbG91ZGV2ZW50cy5zYW1wbGUuc2VudCIsImRhdGFjb250ZW50dHlwZSI6ImFwcGxpY2F0aW9uL2pzb24iLCJ0aW1lIjoiMjAyNC0wNS0yM1QwMjowNTozMS4zMzU5NzNaIiwiZGF0YSI6eyJpZCI6MywibWVzc2FnZSI6IkhlbGxvLCBXb3JsZCEifSwiZXZlbnRrZXkiOiJldmVudHZhbHVlIn0="
}
```

[...]

PR #1063, opened 2024-05-23
Fix the `content-type` issue for the MQTT protocol

yanmxa:
Fixed: https://github.com/cloudevents/sdk-go/issues/1062

### DIFF
--- a/protocol/mqtt_paho/v2/message.go
+++ b/protocol/mqtt_paho/v2/message.go
@@ -17,8 +17,7 @@ import (
 )
 
 const (
-	prefix      = "ce-"
-	contentType = "Content-Type"
+	prefix = "ce-"
 )
 
 var specs = spec.WithPrefix(prefix)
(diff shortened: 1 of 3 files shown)


yanmxa:
This is a bug of the MQTT protocol in the structured mode. Please take a look @embano1 ~

protocol/mqtt_paho/v2/message.go
```diff
@@ -42,7 +41,7 @@ func NewMessage(msg *paho.Publish) *Message {
 	var v spec.Version
 	if msg.Properties != nil {
 		// Use properties.User["Content-type"] to determine if message is structured
```
embano1, line 43: should this comment be removed then?

protocol/mqtt_paho/v2/message.go
```diff
@@ -88,14 +87,20 @@ func (m *Message) ReadBinary(ctx context.Context, encoder binding.BinaryWriter)
...
 			}
-		} else if userProperty.Key == contentType {
-			err = encoder.SetAttribute(m.version.AttributeFromKind(spec.DataContentType), string(userProperty.Value))
 		}
 		if err != nil {
 			return
 		}
 	}
 
+	contentType := m.internal.Properties.ContentType
```
embano1, line 96: question: no changes needed for structured encoding?

embano1 (commented):
Just a nit and a question

protocol/mqtt_paho/v2/message.go
yanmxa: Yes! The binary message doesn't need the content type attribute.  Remove it now.

protocol/mqtt_paho/v2/message.go
yanmxa: Done.

protocol/mqtt_paho/v2/message.go
yanmxa: We still need it here, cause the attribute `datacontenttype` of the event is decoded from this field with a binary message

embano1 (approved):
LGTM (merging once tests pass)

Issue #834, opened 2023-02-03
returning protocol.ResultACK in receiver doesn't acknowledge the event


presmihaylov:
Given this example within a consumer using google pubsub:
```
	err = cloudEventsClient.StartReceiver(ctx, func(ctx context.Context, event cloudevents.Event) protocol.Result {
		log.Infow("received event from topic", "event.id", event.Context.GetID())

		var data map[string]interface{}
		if err := event.DataAs(&data); err != nil {
			log.Errorw("error decoding data", "err", err)
		}

		log.Infow("decoded event", "event.id", event.Context.GetID(), "event.data", data)

                 // this will lead to the event being consumed by this receiver infinitely
		return protocol.ResultACK

                 // this, otoh, will consume the event once
		// return nil
	})
```

returning `protocol.ResultACK` will actually not acknowledge the message and it will be retried infinitely. 
Returning`nil` on the other hand, acknowledges the message, but this behavior feels wonky.

Is this expected?

embano1:
Before I dig deeper, can you please provide the exact module (e.g. `v2`) and version/commit you're using for the above code? Is there any error printed which would help me debugging the line?

preslavmihaylov:
Example receiver:
```
package main

import (
	"context"
	"fmt"
	"log"

	cepubsub "github.com/cloudevents/sdk-go/protocol/pubsub/v2"
	cloudevents "github.com/cloudevents/sdk-go/v2"
	"github.com/cloudevents/sdk-go/v2/event"
	"github.com/cloudevents/sdk-go/v2/protocol"
)

func receive(ctx context.Context, event event.Event) error {
	fmt.Printf("------------------ Received MSG ------------------\n")
	return protocol.ResultACK
}

func main() {
	ctx := context.Background()
	t, err := cepubsub.New(context.Background(),
		cepubsub.WithProjectID("helloworld"),
		cepubsub.AllowCreateTopic(true),
		cepubsub.AllowCreateSubscription(true),
		cepubsub.WithSubscriptionAndTopicID("helloworld", "helloworld"))
	if err != nil {
		log.Fatalf("failed to create pubsub protocol, %s", err.Error())
	}
	c, err := cloudevents.NewClient(t)

	if err != nil {
		log.Fatalf("failed to create client, %s", err.Error())
	}

	log.Println("Created client, listening...")

	if err := c.StartReceiver(ctx, receive); err != nil {
[...]

PR #1064, opened 2024-06-10
fix: pubsub ack must rely on protocol.IsAck

chapurlatn:
PUB/SUB Message acking is missing protocol.Result support.
For example, stan seems to properly handle it: https://github.com/cloudevents/sdk-go/blob/main/protocol/stan/v2/message.go#L66.

This PR fixes the issue with related tests.

Old cases supported:

| Function result                     | Acked/Nacked | Finish result       |
|-------------------------------------|--------------|---------------------|
| nil                                 | ACKED        | nil                 |
| any error | NACKED       | the original error  |

New behaviour:

| Function result                     | Acked/Nacked | Finish result       |
|-------------------------------------|--------------|---------------------|
| nil                                 | ACKED        | nil                 |
| any error except cloudevents.Result | NACKED       | the original error  |
| cloudevents.Result (param acked)    | ACKED        | nil                 |
| cloudevents.Result (param nacked)   | NACKED       | the protocol.Result |

### DIFF
--- a/protocol/pubsub/v2/message.go
+++ b/protocol/pubsub/v2/message.go
@@ -126,9 +126,8 @@ func (m *Message) GetExtension(name string) interface{} {
 func (m *Message) Finish(err error) error {
 	if protocol.IsACK(err) {
 		m.internal.Ack()
-		return nil
+	} else {
+		m.internal.Nack()
 	}
-
-	m.internal.Nack()
 	return err
 }
--- a/protocol/pubsub/v2/message.go
+++ b/protocol/pubsub/v2/message.go
@@ -14,6 +14,7 @@ import (
 	"github.com/cloudevents/sdk-go/v2/binding"
 	"github.com/cloudevents/sdk-go/v2/binding/format"
 	"github.com/cloudevents/sdk-go/v2/binding/spec"
+	"github.com/cloudevents/sdk-go/v2/protocol"
 )
 
 const (
@@ -120,14 +121,14 @@ func (m *Message) GetExtension(name string) interface{} {
 }
 
 // Finish marks the message to be forgotten.
-// If err is nil, the underlying Pubsub message will be acked;
-// otherwise nacked and return the error.
+// Regarding protocol.IsAck, the Pubsub message will be acked or not.
+// If not acked, the original error will be returned.
 func (m *Message) Finish(err error) error {
-	if err != nil {
-		m.internal.Nack()
-		return err
+	if protocol.IsACK(err) {
+		m.internal.Ack()
+		return nil
 	}
 
-	m.internal.Ack()
-	return nil
+	m.internal.Nack()
+	return err
 }
--- a/protocol/pubsub/v2/message.go
+++ b/protocol/pubsub/v2/message.go
@@ -120,9 +120,8 @@ func (m *Message) GetExtension(name string) interface{} {
 	return m.internal.Attributes[prefix+name]
 }
 
-// Finish marks the message to be forgotten.
-// Regarding protocol.IsAck, the Pubsub message will be acked or not.
-// If not acked, the original error will be returned.
+// Finish marks the message to be forgotten and returns the provided error without modification.
+// If err is nil or of type protocol.ResultACK the PubSub message will be acknowledged, otherwise nack-ed.
 func (m *Message) Finish(err error) error {
 	if protocol.IsACK(err) {
 		m.internal.Ack()
--- a/protocol/pubsub/v2/message_test.go
+++ b/protocol/pubsub/v2/message_test.go
@@ -13,6 +13,7 @@ import (
 	"cloud.google.com/go/pubsub"
 	"github.com/cloudevents/sdk-go/v2/binding"
 	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/cloudevents/sdk-go/v2/protocol"
 )
 
 func TestReadStructured(t *testing.T) {
@@ -61,6 +62,22 @@ func TestFinish(t *testing.T) {
 			},
 			err:     fmt.Errorf("error"),
 			wantErr: true,
+		},
+		{
+			name: "result not acked",
+			pm: &pubsub.Message{
+				ID: "testid",
+			},
+			err:     protocol.NewReceipt(false, "error"),
+			wantErr: true,
+		},
+		{
+			name: "result acked",
+			pm: &pubsub.Message{
+				ID: "testid",
+			},
+			err:     protocol.NewReceipt(true, "error"),
+			wantErr: false,
 		},
 		{
 			name: "no errors",
--- a/protocol/pubsub/v2/message_test.go
+++ b/protocol/pubsub/v2/message_test.go
@@ -77,7 +77,7 @@ func TestFinish(t *testing.T) {
 				ID: "testid",
 			},
 			err:     protocol.NewReceipt(true, "error"),
-			wantErr: false,
+			wantErr: true,
 		},
 		{
 			name: "no errors",
--- a/v2/client/client.go
+++ b/v2/client/client.go
@@ -47,6 +47,13 @@ type Client interface {
 	// * func(event.Event) (*event.Event, protocol.Result)
 	// * func(context.Context, event.Event) *event.Event
 	// * func(context.Context, event.Event) (*event.Event, protocol.Result)
+	// Details about the "fn" function returned values:
+	// * When using protocol.NewReceipt, protocol.ResultACK or protocol.ResultNACK as a result
+	//   (*protocol.Receipt implements protocol.Result) rather than an error,
+	//   protocols that support message ack will react regarding Receipt.ACK field.
+	// * A nil result is taken into account as a protocol.Receipt with acked enabled.
+	// * Also, Receipts whose ACK field is true will not trigger warning logs.
+	//   Note that you may use ObservabilityService to setup a custom logging policy.
 	StartReceiver(ctx context.Context, fn interface{}) error
 }
 
--- a/v2/client/client.go
+++ b/v2/client/client.go
@@ -52,8 +52,6 @@ type Client interface {
 	//   (*protocol.Receipt implements protocol.Result) rather than an error,
 	//   protocols that support message ack will react regarding Receipt.ACK field.
 	// * A nil result is taken into account as a protocol.Receipt with acked enabled.
-	// * Also, Receipts whose ACK field is true will not trigger warning logs.
-	//   Note that you may use ObservabilityService to setup a custom logging policy.
 	StartReceiver(ctx context.Context, fn interface{}) error
 }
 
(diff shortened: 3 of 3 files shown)


chapurlatn:
Note: I've found an unclosed issue about that so I mentionned the PR is there: https://github.com/cloudevents/sdk-go/issues/834

protocol/pubsub/v2/message.go
```diff
@@ -120,14 +121,14 @@ func (m *Message) GetExtension(name string) interface{} {
...
-// otherwise nacked and return the error.
+// Regarding protocol.IsAck, the Pubsub message will be acked or not.
+// If not acked, the original error will be returned.
 func (m *Message) Finish(err error) error {
-	if err != nil {
-		m.internal.Nack()
-		return err
+	if protocol.IsACK(err) {
+		m.internal.Ack()
+		return nil
```
embano1, line 129: question: this would mean we're dropping any potential error if `err != nil` here. Would there be a risk of losing error information then? Why not `return err` here like you do in the other case below?

protocol/pubsub/v2/message_test.go
```diff
@@ -62,6 +63,22 @@ func TestFinish(t *testing.T) {
...
+			err:     protocol.NewReceipt(false, "error"),
+			wantErr: true,
+		},
+		{
+			name: "result acked",
+			pm: &pubsub.Message{
+				ID: "testid",
+			},
+			err:     protocol.NewReceipt(true, "error"),
+			wantErr: false,
```
embano1, line 80: see other discussion: IMHO we want to surface the error here or is it already handled/logged somewhere before?

embano1 (commented):
Can you please also take a look at action item (2) here and update the doc string?

https://github.com/cloudevents/sdk-go/issues/834#issuecomment-1671193683

protocol/pubsub/v2/message.go
chapurlatn: Hi @embano1,
When an error is returned here, it's forwarded as the result of the "invoke" func of the client. In that case, it just triggers some logs: https://github.com/cloudevents/sdk-go/blob/main/v2/client/client.go#L261
The error returned remains available when wrapping receiver using RecordCallingInvoker https://github.com/cloudevents/sdk-go/blob/main/v2/client/invoker.go#L91 (the defer Finish is not called at this point).
So I'm assuming the developer made an assumed choice by using a protocol.Result rather than an error and he probably don't want to be anoyed by some warning logs he may probably handle himself in the receive func or in custom callinginvoker hook (the choice made here is not 100% sure to me but I think it may the better one in that case).

protocol/pubsub/v2/message_test.go
chapurlatn: The point is, I think that, when returning protocol.Result instead of error, that's much more a "state" than a real error. But, we could keep the error and just mitigate pubsub ack behaviour.
(In my case, I would for sure disable cloudevents logging in that case but it's fine to me)

chapurlatn:
> Can you please also take a look at action item (2) here and update the doc string?
> 
> [#834 (comment)](https://github.com/cloudevents/sdk-go/issues/834#issuecomment-1671193683)

Hi @embano1,
Not sure to properly understand the expected action about the doc here.
The StartReceiver doc string seems right to me: https://github.com/cloudevents/sdk-go/blob/310da9018ea36b7c6faa3c06aa95c8d8f7a834db/v2/client/client.go#L50
Eventually, we may bring additinal information, such:
```
When protocol.NewReceipt as a result (*protocol.Receipt implements protocol.Result) rather than an error, the protocols that supports ack/nack like PUB/SUB will react regarding Receipt.ACK field. Also, Receipts whose ACK field is true will not trigger warning logs. Note that you may use ObservabilityService to setup a custom logging policy.
```

protocol/pubsub/v2/message.go
embano1: IMHO the current change is a breaking change and we shouldn't do it (I don't see a strong reason to change error handling here). So let's please `return err` here as well (`err` could be `nil` of course, which also handles `nil` scenarios). Please add a test to assert this.

protocol/pubsub/v2/message_test.go
embano1: `protocol.Result` is of type `error` and implements `error` specific [logic](https://github.com/cloudevents/sdk-go/blob/972dfc7354b54f0a7dc6929a1a47930062a8596e/v2/protocol/result.go) e.g., assertions so I would classify it as `error` because that's how users of the SDK can use it (uses error wrapping internally). Please see other comment regarding returning `error` and adding a test.

v2/client/client.go
```diff
@@ -47,6 +47,13 @@ type Client interface {
 	// * func(event.Event) (*event.Event, protocol.Result)
 	// * func(context.Context, event.Event) *event.Event
 	// * func(context.Context, event.Event) (*event.Event, protocol.Result)
+	// Details about the "fn" function returned values:
+	// * When using protocol.NewReceipt, protocol.ResultACK or protocol.ResultNACK as a result
+	//   (*protocol.Receipt implements protocol.Result) rather than an error,
+	//   protocols that support message ack will react regarding Receipt.ACK field.
+	// * A nil result is taken into account as a protocol.Receipt with acked enabled.
+	// * Also, Receipts whose ACK field is true will not trigger warning logs.
```
embano1, line 55: Thx! With the new changes, please remove everything from "Also ..." here

protocol/pubsub/v2/message.go
chapurlatn: Done ✅

v2/client/client.go
chapurlatn: @embano1 I carrefully reread the doc and find a mistake (please check)
I previously wrote:
```
// * A nil result is taken into account as a protocol.Receipt with acked enabled.
```
I think it should be:
```
// * A nil result is taken into account as a protocol.Receipt with acked disabled.
```

protocol/pubsub/v2/message.go
```diff
@@ -120,14 +121,13 @@ func (m *Message) GetExtension(name string) interface{} {
 }
 
 // Finish marks the message to be forgotten.
-// If err is nil, the underlying Pubsub message will be acked;
-// otherwise nacked and return the error.
+// Regarding protocol.IsAck, the Pubsub message will be acked or not.
```
embano1, line 124: I suggest changing the description to:

```
Finish marks the message to be forgotten and returns the provided error without modification. If err is nil or of type protocol.ResultACK the PubSub message will be acknowledged, otherwise nack-ed.
```

embano1: Thinking more about this here, I think we should always return nil and document this here because otherwise the caller could think that finish returned an error which is not the case. WDYT?

v2/client/client.go
```diff
@@ -47,6 +47,11 @@ type Client interface {
 	// * func(event.Event) (*event.Event, protocol.Result)
 	// * func(context.Context, event.Event) *event.Event
 	// * func(context.Context, event.Event) (*event.Event, protocol.Result)
+	// Details about the "fn" function returned values:
```
embano1, line 50: IMHO we don't need to go into too much detail here. What do you think about changing all the `protocol.Result` returns in this doc string to `error` to keep it simple?

protocol/pubsub/v2/message.go
```diff
@@ -119,15 +120,13 @@ func (m *Message) GetExtension(name string) interface{} {
 	return m.internal.Attributes[prefix+name]
 }
 
-// Finish marks the message to be forgotten.
-// If err is nil, the underlying Pubsub message will be acked;
-// otherwise nacked and return the error.
+// Finish marks the message to be forgotten and returns the provided error without modification.
+// If err is nil or of type protocol.ResultACK the PubSub message will be acknowledged, otherwise nack-ed.
```
chapurlatn, line 124: I've updated the pubsub doc according to your comment.
About the error returned by Finish, I agree this is "weird" currently.
But note that the error returned is actually used to perform some logging...
I suggest we open another Issue to discuss these topics:
- do not forward the error in Finish
- use error only for real acknowledgement error
- add an eventual "status" result to finish result to share with "ceClient" (logs, observability, ...)

v2/client/client.go
```diff
@@ -36,17 +36,18 @@ type Client interface {
...
-	// * func(event.Event) (*event.Event, protocol.Result)
 	// * func(context.Context, event.Event) *event.Event
-	// * func(context.Context, event.Event) (*event.Event, protocol.Result)
+	// These signatures are available with an additional error returned:
+	// * func() error
+	// * ...
+	// * func(context.Context, event.Event) (*event.Event, error)
+	// The error returned may impact the messages processing made by the protocol
+	// used (example: message acknowledgement). Please refer to each protocol's
+	// package documentation of the function "Finish(err error) error".
```
chapurlatn, line 50: I removed all functions with additional error returned and added a dedicated notice for result with errors.
I made a reference to each protocol's Finish doc to detail acknowledgement behaviours.
It seems much simpler and avoids doc duplication.
wdyt?

v2/client/client.go
chapurlatn: Done

protocol/pubsub/v2/message.go
chapurlatn: replied here: https://github.com/cloudevents/sdk-go/pull/1064#discussion_r1665418583

chapurlatn: @embano1
I've updated the pubsub doc according to your comment.

About your suggestion for the Finish function error processing:

Thinking more about this here, I think we should always return nil and document this here because otherwise the caller could think that finish returned an error which is not the case. WDYT?

I agree this is "weird" currently but note that the error returned is actually used to perform some logging...
https://github.com/cloudevents/sdk-go/blob/main/v2/client/client.go#L170

I suggest we open another Issue to discuss these topics:

do not forward the error in Finish
return error in Finish only for real acknowledgement error
add an additional "status" result to share all processing info with ceClient in shared layers (logs, observability, ...)

v2/client/client.go
chapurlatn, line 50: - replaced all protocol.Result by simple error
- removed all functions with additional error returned and added a dedicated notice about ability to returned an additional error for al cases
- made a reference to each protocol's Finish doc to detail acknowledgement behaviours.

wdyt?

chapurlatn:
Hi @embano1,
The fork has been rebased from cloudevents/sdk-go main branch.
I'm on vacation during the next 2 weeks. I'll take a look anyway if you ask for changes.
Given previous comments, all has been taken into account apart the error propagation refactoring (but as I've suggested in comments, It may/should be handled in a separated PR).

v2/client/client.go
embano1: LGTM

embano1: For clarity I'd prefer the old style of having the individual function signatures (with and without error). The "..." could be misinterpreted and become outdated if we add another signate without an error type. So my recommendation would be to switch back to the previous "explicit" style, use `error` as you have and then finish the comment with your note with "The error returned may...". OK?

embano1 (commented):
Just a minor nit on the comment and +1 on keeping the existing `Finish` error handling behavior.

duglin:
@chapurlatn wanna rebase and address Michael's comment?

chapurlatn:
@embano1 @duglin Rebased, comments fixed as suggested, CI is green. Seems good to me now.

duglin (approved):
LGTM
@embano1 for the final review/merge

embano1:
Thank you!

PR #1066, opened 2024-06-13
CESQL v1 fixes

Cali0707:
This PR updates the CESQL implementation here to match the changes made in the spec v1 that was approved today. Since there are a lot of changes, I've tried to organize them by commit


Cali0707:
There is one more change I would like to include (either as a follow up or in this PR), but I wasn't sure what the API should be for it:

Currently this SDK only supports the "fail fast" error handling mode, where the CESQL engine returns as soon as it encounters an error. However, I would like to allow users to choose which error handling strategy they would like to use (and then maybe default to fail fast to avoid breaking changes as much as possible). Any ideas on what a good API could be for that?

Cali0707:
cc @duglin @pierDipi @embano1

duglin:
rebase is needed

Cali0707:
@duglin can you re-check?

sql/v2/errors/errors.go
```diff
@@ -0,0 +1,151 @@
+/*
+ Copyright 2021 The CloudEvents Authors
```
lionelvillard, line 2: change to 2024

lionelvillard (approved):
LGTM

lionelvillard:
@duglin for second approval/merge

sql/v2/errors/errors.go
Cali0707: Fixed!

matzew:
LGTM, thanks for the work, @Cali0707

duglin:
rebase needed

sql/v2/errors/errors.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	errorMessages := make([]string, 0, len(errs))
+	for _, err := range errs {
+		errorMessages = append(errorMessages, err.Error())
+	}
+
+	return cesqlError{
+		kind:    parseError,
+		message: strings.Join(errorMessages, ","),
```
duglin, line 65: Bold assumption that "," will never be used in an error message.... ever. I would have thought that "\n" would be better (or a less likely character, like "|", if you still wanted a single line) so that people could more easily split this into individual errors later on if needed. Not a show-stopper though.

sql/v2/errors/errors.go
```diff
@@ -0,0 +1,151 @@
...
+func NewFunctionEvaluationError(err error) error {
+	return cesqlError{
+		kind:    functionEvaluationError,
+		message: err.Error(),
+	}
+}
+
+func IsGenericError(err error) bool {
+	if cesqlErr, ok := err.(cesqlError); ok {
+		return cesqlErr.kind < parseError || cesqlErr.kind > functionEvaluationError
```
duglin, line 148: This feels risky to me. If someone adds a new error type will they remember to update this "return" line? I wonder if it would be safer to add a dummy error type to the end of the "const" list and then say `>= dummyLastError` here instead?

duglin: I'd probably also use "0" instead of "parseError" in case someone adds a new item to the front of the const list.

sql/v2/test/tck/binary_comparison_operators.yaml
```diff
@@ -56,6 +64,10 @@ tests:
   - name: abc is not equal to abc (diamond operator)
     expression: "'abc' <> 'abc'"
     result: false
+  - name: Diamond operator returns false when encountering a missing attribute
```
duglin, line 67: Funny, never heard it called "Diamond operator" before. I like it! Now I feel like I've been missing out all of these years.

sql/v2/types.go
```diff
@@ -34,6 +34,21 @@ func (t Type) String() string {
...
+	switch t {
+	case StringType:
+		return ""
+	case IntegerType:
+		return 0
+	case BooleanType:
+		return false
+	case AnyType:
+		// by default, return false
+		return false
```
duglin, line 47: I don't have a strong opinion on this one, but I'm curious why it defaults to "false" (here and before) instead of "nil" ?

duglin: Or is it just because it needs to be of a well-known type, and "nil" doesn't fit that requirement?

duglin:
Just a few minor comments, none should stop this from being merged though.
so LGTM once the rebase is one and @Cali0707 says it's ready to go after looking at the comments.

sql/v2/types.go
Cali0707: Throughout the spec whenever the return type is ambiguous we always have it defaulting to `Boolean`, which has a zero value of `false`. So, this way that works without much effort on our part in the SDK

duglin: makes sense - thanks

duglin (approved):
LGTM

PR #1070, opened 2024-07-02
Add WithHost Option

jaxtonw:
Implements suggested solution #2 from [Issue #1069](https://github.com/cloudevents/sdk-go/issues/1069). 

Adds a new `WithHost` option for HTTP Protocols to enable the ability to set the HTTP `Host` header manually.

PR #396, opened 2020-03-18
Bringing back the pub/sub binding for V2

nlopezgi:
Creating a V2 version of a pub/sub binding by refactoring the existing bits of V1 pub/sub and using as sample the existing binding for http/nats/sarama/kafka.

PR #1072, opened 2024-07-05
exposed custom publish settings for pubsub client

JamesBLewis:


protocol/pubsub/v2/internal/connection_test.go
```diff
@@ -161,6 +161,58 @@ func TestPublishCreateTopic(t *testing.T) {
...
+		defer client.Close()
+
+		psconn := &Connection{
+			AllowCreateSubscription: true,
+			AllowCreateTopic:        true,
+			Client:                  client,
+			ProjectID:               projectID,
+			TopicID:                 topicID,
+			SubscriptionID:          subID,
+			PublishSettings: &pubsub.PublishSettings{
```
duglin, line 186: The change itself looks good to me, but I'm curious how this tests it. What in the test makes it fail and what would change if the PublishSettings were not specified?

JamesBLewis: If not specified then default publish settings would be used as documented here https://github.com/googleapis/google-cloud-go/blob/265963bd5b91c257b3c3d3c1f52cdf2b5f4c9d1a/pubsub/topic.go#L76

With regards to what would make this test fail, it exists to more than anything to statically verify the setting is exposed.

duglin: What verifies that the settings used came from what's on line 186 instead of the defaults?

JamesBLewis: Given that the change is to the interface there is a limit to how far we can verify this. that limit (without testing google's own pubsub code) is that we can verify the Public variable is exposed in by the struct.

JamesBLewis: I see two other similar PRs to draw potential inspiration from:
1. The original PR adding receiveSettings https://github.com/cloudevents/sdk-go/pull/396/files in which the default value is declared in the sdk go and then a simple check is performed in the test to see if the value is equal to the default.
1. This PR exposing a custom host option https://github.com/cloudevents/sdk-go/pull/1070/files in which a setter function is added and unit tested. 

There is an argument to be made that we should define our own default publish settings like receive although I don't really see how this would make the sdk more intuitive though.

JamesBLewis: my underlying point here remains that the unit test for receive goes arguably less far then this publishSettings test

duglin: Is there a way to make it so that w/o the parameter the operation works and then with your settings it fails? I could live with that much.

Or if that can't be done (e.g. because we need an instance of google pubsub and we don't have that), then perhaps try to get an error from the call to CreateTopic() that complains about a parameter in your settings. That'll prove your settings made it to the pubsub code.

WDYT?

duglin: @JamesBLewis  ping

philicious: @duglin I could offer to take over and finish this up, after I know the PubSub protocol codebase and tests after doing #1180 
I would likely go with a simple assertion testing if a provided override is used instead of the default.

but I would leave it up to @JamesBLewis if he might still wanna do it?

PR #1073, opened 2024-07-09
add explicit ack/nak for nats jetstream

stephen-totty-hpe:
In some cases, it is necessary to control manual ACK/NCK based on the existance of errors.  It is ok to ACK/NCK even if the message has already been committed.  In that case, the ACK/NCK is short-circuited.

### DIFF
--- a/protocol/nats_jetstream/v2/message.go
+++ b/protocol/nats_jetstream/v2/message.go
@@ -99,7 +99,24 @@ func (m *Message) ReadBinary(ctx context.Context, encoder binding.BinaryWriter) 
 
 // Finish *must* be called when message from a Receiver can be forgotten by the receiver.
 func (m *Message) Finish(err error) error {
-	return nil
+	// Ignore error if message has already been Ack/Nak.
+	// Ack and Nak first checks to see if the message has been acknowleged
+	// and immediately returns an error without applying any logic to the message on the server.
+	// A message will auto acknowledge unless manual ack is set or policy is set to AckNonePolicy
+	// see: ManualAck() subscription option
+	// see: ConsumerConfig.AckPolicy
+	if err != nil {
+		errNak := m.Msg.Nak()
+		if errNak == nats.ErrMsgAlreadyAckd || errNak == nats.ErrCantAckIfConsumerAckNone {
+			return nil
+		}
+		return errNak
+	}
+	errAck := m.Msg.Ack()
+	if errAck == nats.ErrMsgAlreadyAckd || errAck == nats.ErrCantAckIfConsumerAckNone {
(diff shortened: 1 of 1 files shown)


protocol/nats_jetstream/v2/message.go
```diff
@@ -99,6 +99,17 @@ func (m *Message) ReadBinary(ctx context.Context, encoder binding.BinaryWriter)
...
 // Finish *must* be called when message from a Receiver can be forgotten by the receiver.
 func (m *Message) Finish(err error) error {
+	// Ignore error if message has already been Ack/Nak.
+	// Ack and Nak first checks to see if the message has been acknowleged
+	// and immediately returns an error without applying any logic to the message on the server.
+	// A message will auto acknowledge unless manual ack is set or policy is set to AckNonePolicy
+	// see: ManualAck() subscription option
+	// see: ConsumerConfig.AckPolicy
+	if err != nil {
+		_ = m.Msg.Nak()
```
francois-du-toit-hpe, line 109: Does ignoring the return here potentially cause issues? Is the 'already ak'd' error the only kind of error that we can get from this operation?

stephen-totty-hpe: Added code to handle expected error cases.  In all other cases, will return error.

protocol/nats_jetstream/v2/message.go
```diff
@@ -99,7 +99,24 @@ func (m *Message) ReadBinary(ctx context.Context, encoder binding.BinaryWriter)
...
 func (m *Message) Finish(err error) error {
-	return nil
+	// Ignore error if message has already been Ack/Nak.
+	// Ack and Nak first checks to see if the message has been acknowleged
+	// and immediately returns an error without applying any logic to the message on the server.
+	// A message will auto acknowledge unless manual ack is set or policy is set to AckNonePolicy
+	// see: ManualAck() subscription option
+	// see: ConsumerConfig.AckPolicy
+	if err != nil {
+		errNak := m.Msg.Nak()
```
embano1, line 109: What are other errors that `Ack` and `Nak` can return? Can you please run a local dummy test with a real NATS backend and always return an error here so we understand how the system behaves end-to-end? IMHO it should be fine (warning logged), but want to make sure this doesn't lead to an endless invocation loop since with this change we're changing the behavior for users.

duglin: @stephen-totty-hpe ping

stephen-totty-hpe: I will close this PR if https://github.com/cloudevents/sdk-go/pull/1095 gets merged.

PR #1074, opened 2024-07-09
avoid creating nats stream if stream is passed as empty

stephen-totty-hpe:
A stream name is not needed when sending a NATS message.  The subject should be enough to send a NATS message.  Currently, a stream is required only because a stream is being used to "auto-create" a stream.  

In many cases:
- You might want the send to fail if the stream does not exist.  
- The code that is sending might not even know the stream name.
- The auto-create uses default options.  There are 15-20 options and many of the defaults might not be correct.


protocol/nats_jetstream/v2/sender.go
```diff
@@ -50,15 +50,21 @@ func NewSenderFromConn(conn *nats.Conn, stream, subject string, jsmOpts []nats.J
...
-			Name:     stream,
-			Subjects: []string{stream + ".*"},
-		})
-		if err != nil {
-			return nil, err
+	// A Stream parameter is not needed for a send operation.
+	// A subject is all that is needed.
+	// Below, a stream is created with default stream config which may not be desired.
+	// It may be that the intention is for the call to fail if a stream does not exist.
+	if stream != "" {
```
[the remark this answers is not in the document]
stephen-totty-hpe: Modifed parameter documentation to include the fact the stream is not required.

protocol/nats_jetstream/v2/sender.go
```diff
@@ -44,21 +44,28 @@ func NewSender(url, stream, subject string, natsOpts []nats.Option, jsmOpts []na
...
+		streamInfo, err := jsm.StreamInfo(stream, jsmOpts...)
+
+		if streamInfo == nil || err != nil && err.Error() == "stream not found" {
```
embano1, line 61: I know you're reusing the existing implementation here, but couple questions:

1) Could there be a scenario where both, streamInfo and err are nil? If not, then just checking the error might be enough here
2) Instead of asserting on err.Error() please use typed error checking `ErrStreamNotFound`

protocol/nats_jetstream/v2/sender.go
```diff
@@ -44,21 +44,28 @@ func NewSender(url, stream, subject string, natsOpts []nats.Option, jsmOpts []na
...
-	if streamInfo == nil || err != nil && err.Error() == "stream not found" {
-		_, err = jsm.AddStream(&nats.StreamConfig{
-			Name:     stream,
-			Subjects: []string{stream + ".*"},
-		})
-		if err != nil {
-			return nil, err
+	// A Stream parameter is not needed for a send operation.
+	// A subject is all that is needed.
+	// Below, a stream is created with default stream config which may not be desired.
```
embano1, line 56: The wording is a bit misleading here IMHO. IIUC, NATS will only fail if a stream exists but different options are supplied, correct? I.e., the operation is not always idempotent.


```suggestion
	// If a stream name is supplied, an attempt is made to create that stream with default parameters. If that stream does not exist or exists with the same parameters specified in the below call, the operation will succeed. Otherwise an error is returned.
```

protocol/nats_jetstream/v2/sender.go
```diff
@@ -44,21 +44,28 @@ func NewSender(url, stream, subject string, natsOpts []nats.Option, jsmOpts []na
 
 // NewSenderFromConn creates a new protocol.Sender which leaves responsibility for opening and closing the NATS
 // connection to the caller
+// The stream parameter is only needed if auto-creation of the stream is needed when the stream does not exist.
```
embano1, line 47:
```suggestion
// The stream parameter is optional. If specified, a stream with default stream parameters will be created. If that stream already exists with non-default parameters, creation will fail with an error message.
```

embano1 (changes requested):

protocol/nats_jetstream/v2/sender.go
stephen-totty-hpe: Looking at the source, it looks like there is only one line that returns an info object.  All other lines return a non-nil error.  Once again, I only added an if statement around the existing code.  I can make that change if you like.  Also, if I were to embark on a new implementation based on the new jetstream api, I might simply make changes there.

embano1: Yeah, probably better to fix in the other PR when using the new JetStream API. I was just bringing those concerns up here because I wasn't part of the initial reviews and by asking these questions we have opportunities to simplify the code base/not complicate it further for our future selves :)

Appreciate your work!

duglin:
@stephen-totty-hpe ping

PR #1075, opened 2024-07-09
add ability to use PullConsumers with NATS jetstream with a specified callback

stephen-totty-hpe:
In many cases, the reciever of a message might want some ability to exercise control over the fetching.  Also, with NATS, there are options inside the NATS server that might be beneficial to use a pull-based consumer.

### DIFF
--- a/protocol/nats_jetstream/v2/receiver.go
+++ b/protocol/nats_jetstream/v2/receiver.go
@@ -140,9 +140,22 @@ func (c *Consumer) OpenInbound(ctx context.Context) error {
 	}
 
 	// Wait until external or internal context done
-	select {
-	case <-ctx.Done():
-	case <-c.internalClose:
+	pullSubscriber, isPullSubscriber := c.Subscriber.(*PullSubscriber)
+	if isPullSubscriber {
+		err = c.pullSubscriptionProcessor(ctx, sub, pullSubscriber)
+		if err != nil {
+			errDrain := sub.Drain()
+			if errDrain != nil {
+				return errDrain
+			}
+			return err
+		}
+	} else {
+		// Wait until external or internal context done
+		select {
+		case <-ctx.Done():
+		case <-c.internalClose:
+		}
 	}
 
 	// Finish to consume messages in the queue and close the subscription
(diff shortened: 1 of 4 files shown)


protocol/nats_jetstream/v2/receiver.go
```diff
@@ -163,6 +169,28 @@ func (c *Consumer) applyOptions(opts ...ConsumerOption) error {
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-c.internalClose:
+			return
+		default:
+			msgs, err := ps.FetchCallback(natsSub)
+			if err != nil || msgs == nil {
+				continue
+			}
```
francois-du-toit-hpe, line 185: Not sure I'm a fan of just dropping the error message here, unless it is logged internall before it is called?

Also there's no need for the nil check on `msgs` - `range` over a nil slice is a valid operation and just skips.

protocol/nats_jetstream/v2/receiver.go
```diff
@@ -163,6 +169,28 @@ func (c *Consumer) applyOptions(opts ...ConsumerOption) error {
...
+			for i := range msgs {
+				msg := msgs[i]
```
francois-du-toit-hpe, line 187: I don't think an assignment is needed here given `MsgHandler` is synchonous?

stephen-totty-hpe: removed assignment

protocol/nats_jetstream/v2/receiver.go
stephen-totty-hpe: returning error to calling function.

embano1 (changes requested):
I'm not a NATS expert, but IMHO this PR is using a legacy API:

> // NOTE: JetStream is part of legacy API.
// Users are encouraged to switch to the new JetStream API for enhanced capabilities and
// simplified API. Please refer to the `jetstream` package.
// See: https://github.com/nats-io/nats.go/blob/main/jetstream/README.md


I like the idea of pull-based subscriptions to give more control, but IMHO we should switch to the newer API and clean up the whole code base of this implementation to avoid setting up both, push and pull-based consumers. This can also simplify the current proposal, removing having two similar `for select` implementations as proposed in this PR.

stephen-totty-hpe:
> I'm not a NATS expert, but IMHO this PR is using a legacy API:
> 
> > // NOTE: JetStream is part of legacy API.
> > // Users are encouraged to switch to the new JetStream API for enhanced capabilities and
> > // simplified API. Please refer to the `jetstream` package.
> > // See: https://github.com/nats-io/nats.go/blob/main/jetstream/README.md
> 
> I like the idea of pull-based subscriptions to give more control, but IMHO we should switch to the newer API and clean up the whole code base of this implementation to avoid setting up both, push and pull-based consumers. This can also simplify the current proposal, removing having two similar `for select` implementations as proposed in this PR.

@embano1  Using the new api is a good idea.  The old api was already in use and it seemed I would have to version the library because many of the exposed options are different.  Or I would have to do some hacky translation of the old option to the new option.  There are several slight issues I have with the current implementation that I believe using the new api might benefit. 
If I was to write a new version, I assume it would be at:
protocol/nats_jetstream/v3 ?

embano1:
> If I was to write a new version, I assume it would be at:
protocol/nats_jetstream/v3 ?

Do you think it would be possible to use the new NATS API/library without breaking CloudEvents users? We're currently doing a similar migration in AMQP and while we're still discussing how to avoid small breaking changes (there might be some, but minor), I think we could do the same with NATS incrementally and not break users?

duglin:
@stephen-totty-hpe ping

stephen-totty-hpe:
> @stephen-totty-hpe ping

I will close this PR if https://github.com/cloudevents/sdk-go/pull/1095 gets merged.

PR #1083, opened 2024-07-29
Implement nats_jetstream using new jetstream package

stephen-totty-hpe:
Implement nats_jetstream using new jetstream package.  

Because many of the options were incompatible, I was forced to create new constructors for Consumer, Sender, and Protocol.

Also, *nats.Msg does not implement jetstream.Msg.  Because of this, there is a MsgHandlerV2.

Integration tests are modified to test the new code.

@embano1, let me know if you would rather create a V2 implementation elsewhere


protocol/nats_jetstream/v2/options.go
```diff
@@ -9,9 +9,14 @@ import (
 	"errors"
 
 	"github.com/nats-io/nats.go"
+	"github.com/nats-io/nats.go/jetstream"
 )
 
 var ErrInvalidQueueName = errors.New("invalid queue name for QueueSubscriber")
+var ErrNoConsumerConfig = errors.New("no consumer config was given")
```
embano1, line 16: Question: do we need all those typed errors? They're now part of the API. Also, in those cases a more simple pattern is to use an error struct, for example https://github.com/aws-controllers-k8s/eventbridge-controller/blob/67a9375a6ac7f95f1a3eb8b1b49f329755305c24/pkg/resource/rule/hooks.go#L29-L43

protocol/nats_jetstream/v2/protocol.go
```diff
@@ -67,6 +86,37 @@ func NewProtocolFromConn(conn *nats.Conn, stream, sendSubject, receiveSubject st
 	return p, nil
 }
 
+func NewProtocolFromConnV2(conn *nats.Conn, stream, sendSubject string, jsOpts []jetstream.JetStreamOpt, opts ...ProtocolOption) (*Protocol, error) {
```
embano1, line 89: @evankanderson what do you think about the V2 change? We have to trade backwards-compatibility with maintenance and simple APIs.

@stephen-totty-hpe 
If we go down this route, we need to mark the other method(s) deprecated so linters can catch it. Also, should we use this as opportunity to also add `context.Context` as first parameter to the API?

protocol/nats_jetstream/v2/protocol.go
```diff
@@ -67,6 +86,37 @@ func NewProtocolFromConn(conn *nats.Conn, stream, sendSubject, receiveSubject st
...
+	var err error
+	var js jetstream.JetStream
```
embano1, line 91: Do we need to define those? Is it because of the `err` in L96 and L100?

protocol/nats_jetstream/v2/protocol.go
```diff
@@ -67,6 +86,37 @@ func NewProtocolFromConn(conn *nats.Conn, stream, sendSubject, receiveSubject st
...
+	}
+
+	if err := p.applyOptions(opts...); err != nil {
+		return nil, err
+	}
+
+	if js, err = jetstream.New(conn, jsOpts...); err != nil {
+		return nil, err
+	}
+	ctx := context.Background()
```
embano1, line 103: Ah, see earlier comment on `ctx` :)

protocol/nats_jetstream/v2/receiver.go
```diff
@@ -81,6 +104,26 @@ func NewConsumer(url, stream, subject string, natsOpts []nats.Option, jsmOpts []
 	return c, err
 }
 
+// NewConsumerV2 consumes from filterSubjects given in ConsumerConfig or OrderedConsumerConfig
+// embedded in ConsumerOption.
+// See: WithConsumerConfig(...) and WithOrderedConsumerConfig(...)
+func NewConsumerV2(url string, natsOpts []nats.Option, jsOpts []jetstream.JetStreamOpt, opts ...ConsumerOption) (*Consumer, error) {
```
embano1, line 110: Question: do those need to be public?

protocol/nats_jetstream/v2/receiver.go
```diff
@@ -117,14 +160,53 @@ func NewConsumerFromConn(conn *nats.Conn, stream, subject string, jsmOpts []nats
...
+		err = c.createConsumer(ctx)
+		if err != nil {
+			return err
+		}
+		consumeContext, err = c.JetstreamConsumer.Consume(c.MsgHandlerV2, c.PullConsumeOpt...)
+		if err != nil {
+			return err
+		}
+	} else {
+		return ErrNoJetstream
```
embano1, line 209: can be simplified to:

- just `if c.JetStream != nil {...}`
- remove final `else` and return error directly

embano1: (avoids else logic)

protocol/nats_jetstream/v2/receiver.go
```diff
@@ -134,7 +216,12 @@ func (c *Consumer) OpenInbound(ctx context.Context) error {
 	}
 
 	// Finish to consume messages in the queue and close the subscription
-	return sub.Drain()
+	if sub != nil {
+		return sub.Drain()
+	} else if consumeContext != nil {
```
embano1, line 221: same as above

protocol/nats_jetstream/v2/receiver.go
```diff
@@ -163,6 +250,59 @@ func (c *Consumer) applyOptions(opts ...ConsumerOption) error {
...
+	if err != nil {
+		return err
+	}
+	if c.OrderedConsumerConfig != nil {
+		consumer, err = c.JetStream.OrderedConsumer(ctx, stream, *c.OrderedConsumerConfig)
+	} else if c.ConsumerConfig != nil {
+		consumer, err = c.JetStream.CreateOrUpdateConsumer(ctx, stream, *c.ConsumerConfig)
+	} else {
+		return ErrNoConsumerConfig
+	}
```
embano1, line 267: can we make this switch/case style?

protocol/nats_jetstream/v2/receiver.go
```diff
@@ -163,6 +250,59 @@ func (c *Consumer) applyOptions(opts ...ConsumerOption) error {
...
+		return err
+	}
+	c.JetstreamConsumer = consumer
+	return nil
+}
+
+// getStreamFromSubjects finds the unique stream for the set of filter subjects
+// If more than one stream is found, returns ErrMoreThanOneStream
+func (c *Consumer) getStreamFromSubjects() (string, error) {
+	ctx := context.Background()
```
embano1, line 278: allow passing ctx as parameter?

embano1 (commented):
I'm still torn between adding a new V2 API vs introducing a breaking change and remove the old code.

stephen-totty-hpe:
> I'm still torn between adding a new V2 API vs introducing a breaking change and remove the old code.

@embano1
Yeah, I made this PR as a quick proof of concept to get some feedback on which way to go.  I have the same questions about how to proceed.

protocol/nats_jetstream/v2/protocol.go
stephen-totty-hpe: BTW, I have other PRs to modify behavior in the existing implementation that are not needed using the implementation I have here with the jetstream package.  So however it is done, the new jetstream library would be a welcome change.  For instance, the new jetstream consumers are pull consumers.  Also, an array of filter subjects are exposed.  An there are error handlers that can be passed through the "config" objects.

protocol/nats_jetstream/v2/protocol.go
stephen-totty-hpe: It is because of the combined "if"+"assignment"  on line 100. The `js` would be undefined later on line 105.

protocol/nats_jetstream/v2/receiver.go
stephen-totty-hpe: I assume you mean the options?  Maybe possible to drop jetstream.JetStreamOpt

nats.Option:
About 40 to 50 options

jetstream.JetStreamOpt:
1. WithClientTrace(ct *ClientTrace) JetStreamOpt
2. WithPublishAsyncErrHandler(cb MsgErrHandler) JetStreamOpt 
3. WithPublishAsyncMaxPending(max int) JetStreamOpt 

The ConsumerOption options are needed because there are variations of how the NATS consumer is created. 
There are two variations, "normal" and "ordered".  Each has mutiple fields on how to configure.

embano1:
@duglin @lionelvillard what are your thoughts on changing the NATS JetStream implementation to the latest NATS SDK which provides several advantages for users, and is the recommended way to use JetStream. However, this means:

- We either need to define `V2` methods on this package to expose the new capabilities
- Introduce a `protocol/nats_jetstream/v3` Go module
- Break existing APIs without introducing new Go module/V2 above

What are your preferences?

protocol/nats_jetstream/v2/options.go
evankanderson: Actually, with [`errors.As`](https://pkg.go.dev/errors#As), using simple exported errors works rather than needing to do something more complicated unless you want additional methods.

protocol/nats_jetstream/v2/receiver.go
```diff
@@ -38,6 +39,20 @@ func (r *Receiver) MsgHandler(msg *nats.Msg) {
...
+func (r *Receiver) MsgHandlerV2(msg jetstream.Msg) {
+	// *nats.Msg does not implement jetstream.Msg.
+	// jetstream.MessageHandler.Ack() returns error where nats.Msg.Ack() returns nothing
+	natsMessage := &nats.Msg{
+		Subject: msg.Subject(),
+		Reply:   msg.Reply(),
+		Header:  msg.Headers(),
+		Data:    msg.Data(),
+	}
+	r.incoming <- msgErr{msg: NewMessage(natsMessage)}
```
evankanderson, line 53: Another option would be to define an interface that both `nats.Msg` and `jetstream.Msg` could be wrapped to implement, e.g.

```go
type natsData interface {
  Subject() string
  Reply() string
  Headers() nats.Header
  Data() []byte
}
```

But this seems likely to be just as effective.

protocol/nats_jetstream/v2/receiver.go
```diff
@@ -54,12 +69,20 @@ func (r *Receiver) Receive(ctx context.Context) (binding.Message, error) {
 type Consumer struct {
```
evankanderson, line 69: Given that we've exposed this as a struct with named v1 fields, it feels like maybe introducing a new `v3` version of `nats_jetstream` might be the way to go.  I don't have a really strong feeling about it, though.

stephen-totty-hpe:
> embano1

@embano1 @evankanderson 
Any thoughts on path forward?  I am fine either way.

embano1:
@duglin @lionelvillard any thoughts on https://github.com/cloudevents/sdk-go/pull/1083#issuecomment-2282654670 ?

stephen-totty-hpe:
> > embano1
> 
> @embano1 @evankanderson Any thoughts on path forward? I am fine either way.

I will close this PR if https://github.com/cloudevents/sdk-go/pull/1095 gets merged.

Issue #1093, opened 2024-09-04
Race condition when sending messages with the MQTT protocol


yanmxa:
I launched 10 goroutines to send messages using the MQTT protocol. After a while, it might throw an error like this:

Code to reproduce the issue:
```golang
	for i := 0; i < count; i++ {
		go func() {
			for {
				e := cloudevents.NewEvent()
				e.SetID(uuid.New().String())
				e.SetType("com.cloudevents.sample.sent")
				e.SetSource("https://github.com/cloudevents/sdk-go/samples/mqtt/sender")
				err = e.SetData(cloudevents.ApplicationJSON, map[string]interface{}{
					"id":      rand.Intn(100), // generates a random int between 0 and 99
					"message": "Hello, World!",
				})
				if err != nil {
					log.Printf("failed to set data: %v", err)
				}
				if result := c.Send(
					cecontext.WithTopic(ctx, "test-topic"),
					e,
				); cloudevents.IsUndelivered(result) {
					log.Printf("failed to send: %v", result)
				} else {
					log.Printf("sent: %d, accepted: %t", i, cloudevents.IsACK(result))
				}
				time.Sleep(time.Second)
			}
		}()
		time.Sleep(1 * time.Second)
	}
	<-ctx.Done()
```
Error message:
```bash
panic: runtime error: growslice: len out of range [recovered]
        panic: bytes.Buffer: too large

goroutine 9 [running]:
bytes.growSlice.func1()
        /usr/local/go/src/bytes/buffer.go:232 +0x45
panic({0x725280?, 0x813980?})
        /usr/local/go/src/runtime/panic.go:914 +0x21f
bytes.growSlice({0xc000500bc0, 0x16, 0x44df49?}, 0x40?)
        /usr/local/go/src/bytes/buffer.go:249 +0x8e
bytes.(*Buffer).grow(0xc0007119c8, 0x101000000000000)
        /usr/local/go/src/bytes/buffer.go:151 +0x13d
bytes.(*Buffer).WriteString(0xc0007119c8, {0x0, 0x101000000000000})
        /usr/local/go/src/bytes/buffer.go:191 +0x59
github.com/eclipse/paho.golang/packets.writeString({0x0, 0x101000000000000}, 0x4f66ab?)
        /home/myan/go/pkg/mod/github.com/eclipse/paho.golang@v0.12.0/packets/packets.go:437 +0x3c
[...]

PR #1094, opened 2024-09-04
Fix race condition in MQTT protocol when sending messages

yanmxa:

Resolved: https://github.com/cloudevents/sdk-go/issues/1093

**Summary:**

**Reason:** This issue occurs when a single client sends events across multiple goroutines. Specifically, in the code here:
https://github.com/cloudevents/sdk-go/blob/1aecb204b50da7dae3d5586030c7047479334bdc/protocol/mqtt_paho/v2/protocol.go#L92-L101
at line 98, the MQTT `msg` is designed to hold a single `binding.Message` (or Event) payload. However, in a multi-goroutine environment, multiple `m` values may be written to the shared `msg`, causing a panic like `bytes.Buffer: too large`.

**Solution:** Instead of sharing the same `msg` across goroutines, we will create a copy of it each time a message is sent.

### DIFF
--- a/protocol/mqtt_paho/v2/protocol.go
+++ b/protocol/mqtt_paho/v2/protocol.go
@@ -20,7 +20,6 @@ import (
 
 type Protocol struct {
 	client          *paho.Client
-	config          *paho.ClientConfig
 	connOption      *paho.Connect
 	publishOption   *paho.Publish
 	subscribeOption *paho.Subscribe
@@ -89,7 +88,7 @@ func (p *Protocol) Send(ctx context.Context, m binding.Message, transformers ...
 	var err error
 	defer m.Finish(err)
 
-	msg := p.publishOption
+	msg := p.publishMsg()
 	if cecontext.TopicFrom(ctx) != "" {
 		msg.Topic = cecontext.TopicFrom(ctx)
 		cecontext.WithTopic(ctx, "")
@@ -105,6 +104,16 @@ func (p *Protocol) Send(ctx context.Context, m binding.Message, transformers ...
 		return err
 	}
 	return err
+}
+
+// publishMsg generate a new paho.Publish message from the p.publishOption
+func (p *Protocol) publishMsg() *paho.Publish {
+	return &paho.Publish{
+		QoS:        p.publishOption.QoS,
+		Retain:     p.publishOption.Retain,
+		Topic:      p.publishOption.Topic,
+		Properties: p.publishOption.Properties,
+	}
 }
 
 func (p *Protocol) OpenInbound(ctx context.Context) error {
--- a/test/integration/mqtt_paho/concurrent_test.go
+++ b/test/integration/mqtt_paho/concurrent_test.go
@@ -1,0 +1,98 @@
+/*
+Copyright 2024 The CloudEvents Authors
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package mqtt_paho
+
+import (
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/errgroup"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	cecontext "github.com/cloudevents/sdk-go/v2/context"
+)
+
+func TestConcurrentSendingEvent(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	topicName := "test-ce-client-" + uuid.New().String()
+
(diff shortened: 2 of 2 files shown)


protocol/mqtt_paho/v2/protocol.go
```diff
@@ -107,6 +106,16 @@ func (p *Protocol) Send(ctx context.Context, m binding.Message, transformers ...
...
 
+// publishMsg generate a new paho.Publish message from the p.publishOption
+func (p *Protocol) publishMsg() *paho.Publish {
+	return &paho.Publish{
+		QoS:        p.publishOption.QoS,
+		Retain:     p.publishOption.Retain,
+		Topic:      p.publishOption.Topic,
+		Properties: p.publishOption.Properties,
+	}
+}
```
Cali0707, line 117: If I'm understanding correctly, the race condition is that the `publishOption` can be changed before the message is published?

Since there is only one place to set this field in the struct (https://github.com/cloudevents/sdk-go/blob/e6a74efbacbf4ac70fc8ed598e678f94b6be7a6d/protocol/mqtt_paho/v2/option.go#L29-L37), maybe it would make sense to make the copy there instead of every time a message is published?

protocol/mqtt_paho/v2/protocol.go
```diff
@@ -20,7 +20,6 @@ import (
 
 type Protocol struct {
 	client          *paho.Client
-	config          *paho.ClientConfig
```
embano1, line 23: why are removing it? unused?

protocol/mqtt_paho/v2/protocol.go
embano1: +1 

Left a comment in the issue to better understand the issue/root cause. What is causing the race? Why is `publishOption` changing during concurrent calls? Asking all these questions because I didn't write the implementation and not familiar with MQ. Furthermore, if we change as per the above, I wonder what the implication is for existing users because we're changing the semantics here (copy instead of pointer).

protocol/mqtt_paho/v2/protocol.go
yanmxa: Yes, It's unused.

protocol/mqtt_paho/v2/protocol.go
yanmxa: Thanks @Cali0707!  The code block you referred to is only invoked once when initializing the `client`. The issues arose when sending messages in many goroutines. Here is the [detail](https://github.com/cloudevents/sdk-go/pull/1094#issue-2505143016)

yanmxa: FYI @embano1.

I append the root cause [here](https://github.com/cloudevents/sdk-go/pull/1094#issue-2505143016). 

The `publishOption` is a pointer reference by the MQTT message, which is used to hold the CloudEvent payload. The message(`publishOption`) is shared across goroutines and will be changed by them.

If we change as per the above, the user should not perceive anything.

embano1: > and will be changed by them

Can you please point me to the code/line where this happens?
Also, ideally you also write a test please to surface the current issue (panics) and how this PR fixes it, perhaps there's more issues due to the current implementation which we can detect this way 🙏

yanmxa: @embano1 No worries!

Here’s the [code line](https://github.com/cloudevents/sdk-go/pull/1094/files#diff-c0aa4c2cb67e2b13ef708c2704bf681052f3c816e730574f94ff807cf8dd4b6bR97) where the panic occurs.

I’ve also written an [integration test](https://github.com/cloudevents/sdk-go/pull/1094/commits/74611564cee99d2bb35e7b532107c44df62140be) that covers the concurrent panic issue, which you can use to reproduce it.

test/integration/mqtt_paho/concurrent_test.go
```diff
@@ -0,0 +1,88 @@
...
+	defer close(readyCh)
+
+	senderNum := 10  // 10 gorutine to sending the events
+	eventNum := 1000 // each gorutine sender publishs 1,000 events
+
+	// start a receiver
+	c, err := cloudevents.NewClient(protocolFactory(ctx, t, topicName), cloudevents.WithUUIDs())
+	require.NoError(t, err)
+	go func() {
+		//	sending 10,000 events concurrenly and verify all of them can be recieved
```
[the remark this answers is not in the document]
yanmxa: Done

embano1:
Great, nice work! Please squash your commits and force-push.

yanmxa:
> Great, nice work! Please squash your commits and force-push.

Done.

embano1 (approved):
Thank you!

PR #1095, opened 2024-09-10
Add v3 version of NATS JetStream protocol

stephen-totty-hpe:
As been discussed previously, there has been a desire to implement the CloudEventSDK nats jetstream protocol using the newer jetstream package.  The newer jetstream package exposes more jetstream specific functionality that should make it more flexible to use.  

Much of this code is stolen from the v2 implementation and changed where necessary.  It builds upon comments made in:
https://github.com/cloudevents/sdk-go/pull/1083

Under the covers, this uses a jetstream.Consumer, which can be "normal" or "ordered".  This is done using consumer options WithConsumerConfig and WithOrderedConsumerConfig.

Also, the need to know the stream names upfront is not needed.  Many of the internal options are available if needed, although the only required option is the ConsumerOption with the correct config.


protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,94 @@
...
+type ConsumerOption func(*Consumer) error
+
+// ConsumerType - consumer types that have configurations defined in jetstream package
+type ConsumerType int
+
+const (
+	ConsumerType_Unknown ConsumerType = iota
+	ConsumerType_Ordinary
+	ConsumerType_Ordered
+)
```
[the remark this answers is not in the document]
embano1: Also, since they're not used in this file/context, please move to receiver.go if they're needed

protocol/nats_jetstream/v3/receiver.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	consumerType := c.getConsumerType()
+	switch consumerType {
+	case ConsumerType_Unknown:
+		return ErrNoConsumerConfig
+	case ConsumerType_Ordinary:
+		c.JetstreamConsumer, err = c.JetStream.CreateOrUpdateConsumer(ctx, stream, *c.ConsumerConfig)
+	case ConsumerType_Ordered:
+		c.JetstreamConsumer, err = c.JetStream.OrderedConsumer(ctx, stream, *c.OrderedConsumerConfig)
+	}
```
embano1, line 181: nit: feels overengineered to introduce types here. The simple "getConsumerType" logic can be copied and used here without introducing new types/const

protocol/nats_jetstream/v3/sender.go
```diff
@@ -0,0 +1,119 @@
...
+type Sender struct {
+	JetStream   jetstream.JetStream
+	PublishOpts []jetstream.PublishOpt
+	Conn        *nats.Conn
+	Subject     string
+	connOwned   bool
+}
+
+// NewSender creates a new protocol.Sender responsible for opening and closing the NATS connection
+func NewSender(ctx context.Context, url, subject string, natsOpts []nats.Option, jsOpts []jetstream.JetStreamOpt, opts ...SenderOption) (*Sender, error) {
```
embano1, line 29: Same feedback as in receiver.go

samples/nats_jetstream/v3/message-interoperability/main.go
```diff
@@ -0,0 +1,78 @@
...
+
+package main
+
+import (
+	"context"
+	"io"
+	"log"
+	"os"
+
+	"github.com/kelseyhightower/envconfig"
```
embano1, line 14: question: do we use this package/style in other examples?

samples/nats_jetstream/v3/message-interoperability/main.go
```diff
@@ -0,0 +1,78 @@
...
+
+func main() {
+	var env envConfig
+	if err := envconfig.Process("", &env); err != nil {
+		log.Printf("[ERROR] Failed to process env var: %s", err)
+		os.Exit(1)
+	}
+	ctx := context.Background()
+
+	natsProtocol, err := cloudeventsjsm.NewSender(ctx, env.NATSServer, env.Subject, cloudeventsjsm.NatsOptions(), nil)
```
embano1, line 39: why isn't this calling NewProtocol? The variable is even named accordingly?

samples/nats_jetstream/v3/message-interoperability/main.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		log.Fatalf("failed to create nats protcol, %s", err.Error())
+	}
+
+	defer natsProtocol.Close(ctx)
+
+	httpProtocol, err := cloudeventshttp.New(cloudeventshttp.WithPort(env.Port))
```
embano1, line 46: What is this interoperability test asserting? Not sure this is needed here because it's an example. Consider removing or moving to integration test

samples/nats_jetstream/v3/receiver/main.go
```diff
@@ -0,0 +1,71 @@
...
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/kelseyhightower/envconfig"
```
embano1, line 13: see comments from message-interoperability

embano1 (commented):
Regarding the samples, please take a look at how we use a protocol in https://github.com/cloudevents/sdk-go/blob/main/samples/kafka_confluent/sender/main.go

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,94 @@
...
+// WithPublishOptions configures the Sender
+func WithPublishOptions(publishOpts []jetstream.PublishOpt) SenderOption {
+	return func(s *Sender) error {
+		s.PublishOpts = publishOpts
+		return nil
+	}
+}
+
+// WithConsumerConfig configures the Consumer with the given config
+func WithConsumerConfig(consumerConfig *jetstream.ConsumerConfig) ConsumerOption {
```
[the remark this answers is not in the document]
stephen-totty-hpe: I was able to combine the SenderOptions, CustomerOptions, and JetstreamOptions into the one and only  ProtocolOptions.

I was not able to combine the NatsOptions because the NatsOptions are used early in the process to create the connection.

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/nats-io/nats.go"
+	"github.com/nats-io/nats.go/jetstream"
+)
+
+// Protocol is a reference implementation for using the CloudEvents binding
+// integration. Protocol acts as both a NATS client and a NATS handler.
+type Protocol struct {
+	Conn *nats.Conn
+
+	Consumer        *Consumer
```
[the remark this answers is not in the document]
stephen-totty-hpe: Because of the integration tests needing access to the Sender and Consumer, I left them as exported.  Most everything else of hopefully private.

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,116 @@
...
+	consumerOptions []ConsumerOption
+
+	Sender        *Sender
+	senderOptions []SenderOption
+
+	connOwned bool // whether this protocol created the nats connection
+}
+
+// NewProtocol creates a new NATS protocol.
+func NewProtocol(ctx context.Context, url, stream, sendSubject string, natsOpts []nats.Option, jsOpts []jetstream.JetStreamOpt, opts ...ProtocolOption) (*Protocol, error) {
```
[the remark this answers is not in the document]
stephen-totty-hpe: I think I have removed as much as I can now.

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,116 @@
...
+	return p.Consumer.OpenInbound(ctx)
+}
+
+// Receive implements Receiver.Receive
+func (p *Protocol) Receive(ctx context.Context) (binding.Message, error) {
+	return p.Consumer.Receive(ctx)
+}
+
+// Close implements Closer.Close
+func (p *Protocol) Close(ctx context.Context) error {
```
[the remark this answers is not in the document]
stephen-totty-hpe: I have not seen any race conditions so far.

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,92 @@
...
+var (
+	ErrNoFilterSubjects          = errors.New("no filter subjects were given")
+	ErrMoreThanOneStream         = errors.New("more than one stream for given filter subjects")
+	ErrNoConsumerConfig          = errors.New("no consumer config was given")
+	ErrMoreThanOneConsumerConfig = errors.New("more than one consumer config given")
+)
+
+// NatsOptions is a helper function to group a variadic nats.ProtocolOption into
+// []nats.Option that can be used by either Sender, Consumer or Protocol
+func NatsOptions(opts ...nats.Option) []nats.Option {
```
embano1, line 24: nit: IMHO this helper can either by internal if we need it or removed because it increases the API for no real benefit

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/nats-io/nats.go"
+)
+
+// Protocol is a reference implementation for using the CloudEvents binding
+// integration. Protocol acts as both a NATS client and a NATS handler.
+type Protocol struct {
+	conn *nats.Conn
+
+	Consumer *consumer
```
embano1, line 22: question: why do we need to keep those still public?

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,100 @@
...
+
+	Sender *sender
+
+	connOwned bool // whether this protocol created the nats connection
+}
+
+// NewProtocol creates a new NATS protocol.
+func NewProtocol(ctx context.Context, url, sendSubject string, natsOpts []nats.Option, opts ...ProtocolOption) (*Protocol, error) {
```
embano1, line 30: question: is natsOpts optional (can be nil/0 len) or required? If not, please make it a ProtocolOption otherwise check for its validity and return error early

embano1: Similarly, is url and sendSubject required or can it be "" ?

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, err
+	}
+
+	p.connOwned = true
+
+	return p, nil
+}
+
+// NewProtocolFromConn creates a new NATS protocol with the given nats connection.
+func NewProtocolFromConn(ctx context.Context, conn *nats.Conn, sendSubject string, opts ...ProtocolOption) (*Protocol, error) {
```
embano1, line 48: question: should we accept an interface for nats.Conn here to simplify mocking or hardcode it to this type?

stephen-totty-hpe: unfourtunately, the jetstream library explictly uses a "pointer to struct"

protocol/nats_jetstream/v3/protocol.go
stephen-totty-hpe: url is required by both Consumer and Sender.  sendSubject is required by Sender.

stephen-totty-hpe: natsOpts is optional, but is highly used for things like credentials for security.

protocol/nats_jetstream/v3/protocol.go
stephen-totty-hpe: in integration test:
```cleanup, s, r := testProtocol(ctx, t, conn, tt.args.consumerConfig, tt.args.opts...)```
the sender and receiver needs to be accesible from the protocol in order to call
```test.SendReceive(t, binding.WithPreferredEventEncoding(context.TODO(), tt.args.bindingEncoding), in, s, r,  ...```

I can add a GetSender() or GetConsumer() in order to get the values from the Protocol object.

File: protocol/nats_jetstream/v3/protocol.go (hunk 14 above)
Comment (stephen-totty-hpe) [reply protocol/nats_jetstream/v3/protocol.go:30]: The problem with natsOpts, is that I need it before applying the options in order to get a `*nats.Conn`.  So a sort of chicken-egg problem.  I need the  `*nats.Conn` to create a consumer, but I need a consumer to apply an option.

Comment (stephen-totty-hpe) [reply protocol/nats_jetstream/v3/protocol.go:30]: I could set the `*nats.Conn` on the consumer after applying the options, but I would have to see if any other option or logic needs the `*nats.Conn` sooner.

Comment (stephen-totty-hpe) [reply protocol/nats_jetstream/v3/protocol.go:30]: If you like, I can add three more options:
`func WithURL(url string, natsOpts []nats.Option) ProtocolOption`
`func WithConnection(conn *nats.Conn) ProtocolOption`
`func WithSenderSubject(sendSubject string) ProtocolOption`

And then I could reduce to only one constructor that takes `(ctx context.Context, opts ...ProtocolOption)`

Of course, there would be more errors needed in case a Sender was created without sendSubject. 
Also "one and only one" of WithURL/WithConnection must be provided.

File: protocol/nats_jetstream/v3/message.go [hunk 16]
```diff
@@ -0,0 +1,145 @@
+/*
+ Copyright 2021 The CloudEvents Authors
```
yanmxa, line 2: 's/2021/2024/'

protocol/nats_jetstream/v3/message.go
```diff
@@ -0,0 +1,145 @@
...
+
+// ReadBinary transfers a binary-mode event to an BinaryWriter.
+func (m *Message) ReadBinary(ctx context.Context, encoder binding.BinaryWriter) error {
+	if m.encoding != binding.EncodingBinary {
+		return binding.ErrNotBinary
+	}
+
+	version := m.GetVersion()
+	if version == nil {
+		return binding.ErrNotBinary
```
yanmxa, line 72: In line 66, binary encoding is already asserted by the time execution reaches line 79. It might be beneficial to use a more specific error message to clearly indicate that the version must be provided when using binary mode, such as 'Must provide the version for binary mode.' 

This will improve clarity and distinguish the error from other potential issues.

protocol/nats_jetstream/v3/message.go
```diff
@@ -0,0 +1,145 @@
...
+func (m *Message) GetAttribute(attributeKind spec.Kind) (spec.Attribute, interface{}) {
+	key := withPrefix(attributeKind.String())
+	if m.Msg.Headers() != nil {
+		headerValue := m.Msg.Headers().Get(key)
+		if headerValue != "" {
+			version := m.GetVersion()
+			return version.Attribute(key), headerValue
+		}
+	}
+	return nil, nil
```
yanmxa, line 115: Should `return version.Attribute(key), nil` if the spec version supports the attribute kind, but doesn't have any value

https://github.com/cloudevents/sdk-go/blob/2db66e0958b5649f9851719b7c6d11988f4d1c0d/v2/binding/message.go#L60-L61

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// Protocol is a reference implementation for using the CloudEvents binding
+// integration. Protocol acts as both a NATS client and a NATS handler.
+type Protocol struct {
+	conn *nats.Conn
+
+	Consumer *consumer
+
+	Sender *sender
```
yanmxa, line 24: Regarding the naming, should it be `consumer-producer` or `sender-receiver` to better align with conventions? @stephen-totty-hpe @embano1

protocol/nats_jetstream/v3/protocol.go
yanmxa: Maybe it's a good choice to keep it as concise as the last one, but it requires comments to clarify the function parameter and validation logic

protocol/nats_jetstream/v3/receiver.go
```diff
@@ -0,0 +1,219 @@
...
+			return nil, io.EOF
+		}
+		return msgErr.msg, nil
+	case <-ctx.Done():
+		return nil, io.EOF
+	}
+}
+
+type consumer struct {
+	receiver
```
yanmxa, line 55: It seems that there are too many layers of nesting in the implementation of the `protocol.Receiver`. (Protocol -> consumer -> receiver). This deep nesting appears to increase the complexity of the code maintenance

protocol/nats_jetstream/v3/receiver.go
```diff
@@ -0,0 +1,219 @@
...
+func (c *consumer) Close(ctx context.Context) error {
+	// Before closing, let's be sure OpenInbound completes
+	// We send a signal to close and then we lock on subMtx in order
+	// to wait OpenInbound to finish draining the queue
+	c.internalClose <- struct{}{}
+	c.subMtx.Lock()
+	defer c.subMtx.Unlock()
+
+	if c.connOwned {
+		c.conn.Close()
```
yanmxa, line 148: nit: is there a possibility of encountering a panic when calling `c.conn.Close()` more than once?

protocol/nats_jetstream/v3/protocol.go
embano1: IMHO, since we're starting somewhat from scratch here, the API should be oriented around the use of it: from a sender/receiver perspective. Required arguments should always be part of the signature, anything optional should be a variadic option. Then internally, after all options are applied without throwing individual errors from the option handler, we use a validation function (to keep the constructor clean) to check for more complex validation cases e.g., the "one and only one" case you mention above.

protocol/nats_jetstream/v3/protocol.go
embano1: +1 e.g., sender/receiver

protocol/nats_jetstream/v3/protocol.go
embano1: Hm, I wish there was a way we can keep this `internal` without external users of the SDK being able to access those fields/having to understand why they're public.

stephen-totty-hpe: I realized that Protocol IS a sender and IS a receiver.  So I was able to make them private

protocol/nats_jetstream/v3/protocol.go
stephen-totty-hpe: I added a validateOptions() method.  There is now only one constructor in Protocol, Receiver, Sender
I also added a WithURL() and WithConnection() option.

protocol/nats_jetstream/v3/message.go
stephen-totty-hpe: Added a new error.

protocol/nats_jetstream/v3/message.go
stephen-totty-hpe: Added another case.  At the end, there is no Header, so there is no version.

protocol/nats_jetstream/v3/protocol.go
stephen-totty-hpe: Removed Consumer.  Everything is folded into Receiver

protocol/nats_jetstream/v3/receiver.go
stephen-totty-hpe: added a nil check

yanmxa:
@stephen-totty-hpe Thanks! It looks good to me. Leave it to @embano1 ~

protocol/nats_jetstream/v3/message_test.go
```diff
@@ -0,0 +1,116 @@
...
+var (
+	outBinaryMessage = bindingtest.MockBinaryMessage{
+		Metadata:   map[spec.Attribute]interface{}{},
+		Extensions: map[string]interface{}{},
+	}
+	outStructMessage = bindingtest.MockStructuredMessage{}
+
+	testEvent     = test.FullEvent()
+	binaryData, _ = json.Marshal(map[string]string{
+		"ce_type":            testEvent.Type(),
```
embano1, line 39: Question: Why the "_" field names and not "-"?

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,123 @@
...
+	ErrMoreThanOneStream         = errors.New("more than one stream for given filter subjects")
+	ErrNoConsumerConfig          = errors.New("no consumer config was given")
+	ErrMoreThanOneConsumerConfig = errors.New("more than one consumer config given")
+)
+
+// ProtocolOption is the function signature required to be considered an nats.ProtocolOption.
+type ProtocolOption func(*Protocol) error
+
+// WithURL configures the Sender and/or Receiver
+func WithURL(url string, natsOpts ...nats.Option) ProtocolOption {
```
embano1, line 27: There's quite a lot of logic in this option. I thought URL was always required? If so, better to make it a constructor parameter and configure these fields after all options are applied.

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,123 @@
...
+
+var (
+	ErrNoConnection              = errors.New("URL or nats connection must be given")
+	ErrNoFilterSubjects          = errors.New("no filter subjects were given")
+	ErrMoreThanOneStream         = errors.New("more than one stream for given filter subjects")
+	ErrNoConsumerConfig          = errors.New("no consumer config was given")
+	ErrMoreThanOneConsumerConfig = errors.New("more than one consumer config given")
+)
+
+// ProtocolOption is the function signature required to be considered an nats.ProtocolOption.
```
embano1, line 23:
```suggestion
// ProtocolOption provides a way to configure the protocol
```

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,123 @@
...
+		if p.receiver != nil {
+			p.receiver.url = url
+			p.receiver.natsOpts = opts
+			p.receiver.connOwned = true
+		}
+		return nil
+	}
+}
+
+// WithConnection configures the Sender and/or Receiver
```
embano1, line 48: Please update all option comments with a clear description of what they do e.g., 


```suggestion
// WithConnection uses the provided connection in the protocol sender and receiver
```

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,123 @@
...
+			p.sender.conn = conn
+		}
+		if p.receiver != nil {
+			p.receiver.conn = conn
+		}
+		return nil
+	}
+}
+
+// WithJetStreamOptions configures the Sender and/or Receiver
```
embano1, line 62: See comment above (applies to all)

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,131 @@
...
+	natsOpts  []nats.Option
+	connOwned bool
+
+	receiver *receiver
+
+	sender *sender
+}
+
+// NewProtocol creates a new NATS protocol.
+// Use WithURL() OR WithConnection()
```
embano1, line 33: Ok, so only one of the two can be specified? Just a connection without URL is also valid?

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,131 @@
...
+	url       string
+	natsOpts  []nats.Option
+	connOwned bool
+
+	receiver *receiver
+
+	sender *sender
+}
+
+// NewProtocol creates a new NATS protocol.
```
embano1, line 32: Elaborate more on the protocol constructor please. Also, don't use newlines, just keep describing.


```suggestion
// NewProtocol creates a new protocol using NATS JetStream. [... <describe the default behavior and which options to use/which conflict>]
```

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	if err := p.validateOptions(); err != nil {
+		return nil, err
+	}
+
+	var errConnection error
+	defer func() {
+		// close connection if an error occurs and we created the nats connection
+		if p.connOwned && p.conn != nil && errConnection != nil {
```
embano1, line 48: Do we need `errConnection != nil` check? Can we not just close the connection if p.conn != nil?

embano1: Also, do we need to check for p.connOwned if p.conn != nil? Can we not just close the connection then (not sure what p.connOwned actually does)

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,123 @@
...
+type ProtocolOption func(*Protocol) error
+
+// WithURL configures the Sender and/or Receiver
+func WithURL(url string, natsOpts ...nats.Option) ProtocolOption {
+	return func(p *Protocol) error {
+		opts := []nats.Option{}
+		opts = append(opts, natsOpts...)
+		p.url = url
+		p.natsOpts = opts
+		p.connOwned = true
```
embano1, line 33: Furthermore, I think we can get away without the connOwned field. A protocol user should always call Close on the protocol. Therefore he/she has control over when the connection is closed. I suggest removing this logic here and implementing closing logic in the Close method on the protocol (and document what it does).

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Receive implements Receiver.Receive
+func (p *Protocol) Receive(ctx context.Context) (binding.Message, error) {
+	return p.receiver.Receive(ctx)
+}
+
+// Close implements Closer.Close
+func (p *Protocol) Close(ctx context.Context) error {
+	if p.connOwned {
```
embano1, line 94: See my comment on this field. IMHO we can remove this field, document that a user has to call Close on the protocol, i.e., it's in their control when the connection will be closed, and therefore avoid duplicate calls to Close (is it already idempotent?) e.g., from your constructor where you also close the connection in a defer call when an error occurs.

protocol/nats_jetstream/v3/protocol.go
embano1: Now I see why I'm confused. See my comments on the Close handling below where this whole logic can be removed from the constructor.

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"errors"
+
+	"github.com/nats-io/nats.go"
+	"github.com/nats-io/nats.go/jetstream"
+)
+
+var (
+	ErrNoConnection              = errors.New("URL or nats connection must be given")
+	ErrNoFilterSubjects          = errors.New("no filter subjects were given")
```
embano1, line 17: nit: given -> provided (in all errors)

protocol/nats_jetstream/v3/receiver.go
```diff
@@ -0,0 +1,229 @@
...
+	for _, fn := range opts {
+		if err := fn(p); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// validateOptions runs after all options have been applied and makes sure needed options were set correctly.
+func (r *receiver) validateOptions() error {
```
embano1, line 161: Please remove this redundant validation logic from the sender/receiver and put it into the protocol. Unless you see a need for additional validation logic in the receiver/sender and not in the higher-level protocol which can do the same, avoid duplication, and return earlier?

protocol/nats_jetstream/v3/sender.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	_, err = s.jetStream.PublishMsg(ctx, natsMsg, s.publishOpts...)
+
+	return err
+}
+
+// Close implements Closer.Close
+// This method only closes the connection if the Sender opened it
+func (s *sender) Close(_ context.Context) error {
```
embano1, line 110: Question: do we need those Close() methods on the sender instead of at the protocol layer?

protocol/nats_jetstream/v3/sender.go
```diff
@@ -0,0 +1,139 @@
...
+	if err := s.applyOptions(opts...); err != nil {
+		return nil, err
+	}
+
+	if err := s.validateOptions(); err != nil {
+		return nil, err
+	}
+
+	var errConnection error
+	defer func() {
```
embano1, line 51: See my protocol comments how this can also be simplified

samples/nats_jetstream/v3/receiver/main.go
```diff
@@ -0,0 +1,62 @@
...
+
+func main() {
+	ctx := context.Background()
+
+	natsURL := "nats://localhost:4222"
+	natsSubject := "sample"
+
+	consumerOpt := cejsm.WithConsumerConfig(&jetstream.ConsumerConfig{FilterSubjects: []string{natsSubject}})
+	urlOpt := cejsm.WithURL(natsURL)
+	receiver, err := cejsm.NewReceiver(ctx, consumerOpt, urlOpt)
```
embano1, line 27: Please rewrite your examples using the protocol methods and not sender/receiver implementations

samples/nats_jetstream/v3/sender/main.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := envconfig.Process("", &env); err != nil {
+		log.Fatalf("Failed to process env var: %s", err)
+	}
+
+	natsURL := "nats://localhost:4222"
+	natsSubject := "sample"
+
+	ctx := context.Background()
+	urlOpt := cejsm.WithURL(natsURL)
+	sender, err := cejsm.NewSender(ctx, natsSubject, urlOpt)
```
embano1, line 49: See comment in the receiver example. Please follow Kafka/HTTP samples where we construct the sender/receiver from the CloudEvents client using protocol instead of lower-level methods.

protocol/nats_jetstream/v3/sender.go
```diff
@@ -0,0 +1,139 @@
...
+	jetSteamOpts []jetstream.JetStreamOpt
+	jetStream    jetstream.JetStream
+	publishOpts  []jetstream.PublishOpt
+
+	subject string
+}
+
+// NewSender creates a new protocol.Sender responsible for opening and closing the NATS connection
+// Use WithURL() OR WithConnection()
+func NewSender(ctx context.Context, subject string, opts ...ProtocolOption) (*sender, error) {
```
embano1, line 37: Can we make these sender/receiver constructors private?

protocol/nats_jetstream/v3/receiver.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+		return msgErr.msg, nil
+	case <-ctx.Done():
+		return nil, io.EOF
+	}
+}
+
+// OpenInbound implements Opener.OpenInbound
+func (r *receiver) OpenInbound(ctx context.Context) error {
+	r.subMtx.Lock()
```
embano1, line 106: Since only one protocol will ever use sender/receiver, you can implement locking at the protocol level instead of lower fields/methods here. WDYT? Makes locking simpler, e.g., during Close as well

embano1: Please see other protocol implementations e.g., HTTP/Kafka (Confluent) which use a more simpler flow to reduce protocol complexity

protocol/nats_jetstream/v3/receiver.go
```diff
@@ -0,0 +1,229 @@
...
+		return err
+	}
+	if consumeContext, err = r.jetstreamConsumer.Consume(r.MsgHandler, r.pullConsumeOpts...); err != nil {
+		return err
+	}
+
+	// Wait until external or internal context done
+	select {
+	case <-ctx.Done():
+	case <-r.internalClose:
```
embano1, line 121: not needed if we simplify the whole close logic as per other comments

embano1 (changes requested):
Thx for all the hard work you put into this! Some minor comments on the Options implementation (really like it) and I think we can further reduce the complexity of this code base, especially in the sender/receiver implementations. I suggested to take a look at the HTTP/Kafka (Confluent) protocol implementation for how to simplify the internal logic of this protocol.

protocol/nats_jetstream/v3/options.go
stephen-totty-hpe: You can use one of two options:
1) URL +nats.Options  // if you do not have a connection yet
2) *nats.Conn // if you already have a connection.

This requires either:
1) two constructors
   a) context.Context, URL , nats.Options,   ...ProtocolOption
   b) context.Context, *nats.Conn,   ...ProtocolOption
2) Options that makes the two cases "mutually exclusive"

protocol/nats_jetstream/v3/sender.go
stephen-totty-hpe: The protocol constructor makes both a sender and receiver.  in some cases you might only want a sender OR receiver.

protocol/nats_jetstream/v3/options.go
stephen-totty-hpe: Simplified more.  One constructor.

protocol/nats_jetstream/v3/message_test.go
```diff
@@ -0,0 +1,116 @@
...
+			expectedBinaryError:     nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := NewMessage(tt.receiverMessage)
+			if got == nil {
+				t.Errorf("Error in NewMessage!")
+			}
+			err := got.ReadBinary(context.TODO(), &outBinaryMessage)
```
embano1, line 103: nit: shouldn't you also assert that the contents of the messages are as expected?

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,102 @@
...
+// WithSendSubject sets the subject used in the protocol sender
+func WithSendSubject(sendSubject string) ProtocolOption {
+	return func(p *Protocol) error {
+		p.sendSubject = sendSubject
+		return nil
+	}
+}
+
+// WithConsumerConfig creates a unordered consumer used in the protocol receiver.
+// If WithOrderedConsumerConfig is also given, the receiver will error.
```
embano1, line 73:
```suggestion
// This option is mutually exclusive with WithOrderedConsumerConfig.
```

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,102 @@
...
+		if p.orderedConsumerConfig != nil {
+			return ErrMoreThanOneConsumerConfig
+		}
+		p.consumerConfig = consumerConfig
+		return nil
+	}
+}
+
+// WithOrderedConsumerConfig creates a ordered consumer used in the protocol receiver.
+// If WithConsumerConfig is also given, the receiver will error.
```
embano1, line 85:
```suggestion
// This option is mutually exclusive with WithConsumerConfig.
```

protocol/nats_jetstream/v3/options_test.go
```diff
@@ -0,0 +1,341 @@
+/*
+ Copyright 2024 The CloudEvents Authors
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+package nats_jetstream
```
embano1, line 6: nit: many of these tests don't assert error cases. If you run your tests with coverage, you'll see that several code paths are not checked

samples/nats_jetstream/v3/receiver/main.go
```diff
@@ -0,0 +1,82 @@
...
+	ctx := context.Background()
+
+	natsURL := "nats://localhost:4222"
+	natsSubject := "sample"
+	natsStream := "stream"
+
+	createStream(natsURL, natsStream, natsSubject)
+	consumerOpt := cejsm.WithConsumerConfig(&jetstream.ConsumerConfig{FilterSubjects: []string{natsSubject}})
+	urlOpt := cejsm.WithURL(natsURL)
+	receiver, err := cejsm.New(ctx, consumerOpt, urlOpt)
```
embano1, line 30:
```suggestion
	protocol, err := cejsm.New(ctx, consumerOpt, urlOpt)
```

samples/nats_jetstream/v3/sender/main.go
```diff
@@ -0,0 +1,103 @@
...
+
+	natsURL := "nats://localhost:4222"
+	natsSubject := "sample"
+	natsStream := "stream"
+
+	createStream(natsURL, natsStream, natsSubject)
+	ctx := context.Background()
+	urlOpt := cejsm.WithURL(natsURL)
+	sendopt := cejsm.WithSendSubject(natsSubject)
+	sender, err := cejsm.New(ctx, urlOpt, sendopt)
```
embano1, line 54:
```suggestion
	protocol, err := cejsm.New(ctx, urlOpt, sendopt)
```

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,269 @@
...
+func (p *Protocol) MsgHandler(msg jetstream.Msg) {
+	p.incoming <- msgErr{msg: NewMessage(msg)}
+}
+
+// Close implements Closer.Close
+func (p *Protocol) Close(ctx context.Context) error {
+	// Before closing, let's be sure OpenInbound completes
+	// We send a signal to close and then we lock on subMtx in order
+	// to wait OpenInbound to finish draining the queue
+	p.internalClose <- struct{}{}
```
embano1, line 175: If OpenInbound is already done (ctx cancelled) this will deadlock

protocol/nats_jetstream/v3/message_test.go
stephen-totty-hpe: done

protocol/nats_jetstream/v3/options.go
stephen-totty-hpe: done

protocol/nats_jetstream/v3/options.go
stephen-totty-hpe: done

protocol/nats_jetstream/v3/options_test.go
stephen-totty-hpe: added more unit tests and added extra asserts

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,269 @@
...
+		if err := fn(p); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// validateOptions runs after all options have been applied and makes sure needed options were set correctly.
+func (p *Protocol) validateOptions() error {
+	if p.url == "" && p.conn == nil {
```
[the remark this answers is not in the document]
stephen-totty-hpe: done

test/integration/nats_jetstream/v3/nats_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	conn, err := nats.Connect(s)
+	if err != nil {
+		t.Skipf("Cannot create STAN client to NATS server [%s]: %v", s, err)
+	}
+
+	return conn
+}
+
+func testProtocol(ctx context.Context, t testing.TB, natsConn *nats.Conn, consumerConfig any, opts ...ce_nats.ProtocolOption) (func(), bindings.Sender,
```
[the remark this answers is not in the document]
stephen-totty-hpe: this is an internal method.  It is called from:
`cleanup, s, r := testProtocol(ctx, t, conn, tt.args.consumerConfig, tt.args.opts...)`

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+}
+
+// MsgHandler implements nats.MsgHandler and publishes messages onto our internal incoming channel to be delivered
+// via r.Receive(ctx)
+func (p *Protocol) MsgHandler(msg jetstream.Msg) {
+	p.incoming <- msgErr{msg: NewMessage(msg)}
+}
+
+// Close implements Closer.Close
```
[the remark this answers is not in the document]
stephen-totty-hpe: done

protocol/nats_jetstream/v3/protocol.go
stephen-totty-hpe: The code does this:
`internalClose: make(chan struct{}, 1),`

I believe closing a channel will not block:
```
// The close built-in function closes a channel, which must be either
// bidirectional or send-only. It should be executed only by the sender,
// never the receiver, and has the effect of shutting down the channel after
// the last sent value is received. After the last value has been received
// from a closed channel c, any receive from c will succeed without
// blocking, returning the zero value for the channel element. The form
//
//	x, ok := <-c
//
// will also set ok to false for a closed and empty channel.
```

protocol/nats_jetstream/v3/options.go
stephen-totty-hpe: The `natsOpts ...nats.Option` is critical.  In fact, we could not use this option without the nats.Option argument.  If youe NATS server needs any sort of TLS or authentication with JWTs, these options must be exposed.

stephen-totty-hpe: see:
https://docs.nats.io/using-nats/developer/connecting/creds

or any of the other authentication methods of NATS.

embano1: I see what you mean. Well, it does break the functional options pattern a bit IMHO. I'd expect another option to configure nats options in fact e.g., `WithNatsOptions`. I can see this being useful for other scenarios so can we add this option?

Also this is a bit confusing

> If WithConnection is also used, WithURL will be ignored.

Because ultimately if you specify both options our validation would fail. Can you please change the comment to say that WithURL cannot be used with WithConnection?

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,290 @@
...
+// Close must be called after use to releases internal resources and close the underlying connection
+func (p *Protocol) Close(ctx context.Context) error {
+	// Before closing, let's be sure OpenInbound completes
+	// We send a signal to close and then we lock on subMtx in order
+	// to wait OpenInbound to finish draining the queue
+	p.internalClose <- struct{}{}
+	p.subMtx.Lock()
+	defer p.subMtx.Unlock()
+
+	if p.connOwned && p.conn != nil {
```
embano1, line 179: Do we still want to track owned connections? Currently the comment makes it sound that even if I specify a custom connection, conn would be closed (which would not be the case). I feel we can remove connOwned checks and handle connection closing for both situations

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,290 @@
...
+type Protocol struct {
+
+	// nats connection
+	conn      *nats.Conn
+	url       string
+	natsOpts  []nats.Option
+	connOwned bool
+
+	// jetstream options
+	jetSteamOpts []jetstream.JetStreamOpt
```
embano1, line 33: is this used somewhere?

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,105 @@
...
+package nats_jetstream
+
+import (
+	"errors"
+
+	"github.com/nats-io/nats.go"
+	"github.com/nats-io/nats.go/jetstream"
+)
+
+var (
```
embano1, line 15: Tbh, I feel we're artificially increasing our API surface (coupling) with all these typed errors. If we want to expose a validation error type, why do we feel we need typed errors instead of a simplified approach using a generic validation error type with specific error messages?

Something that gives users a simple assertion approach?

```go
   var vError ValidationError
   if errors.As(err, &vError) {...}
```

embano1: Imagine how a user would have to assert against all these errors below? Do you think there's need/someone would actually do this?

embano1: nit: validation error type(s) should live in the protocol.go

protocol/nats_jetstream/v3/options_test.go
```diff
@@ -0,0 +1,453 @@
...
+			wants: wants{
+				err: nil,
+				protocol: &Protocol{
+					url:       expectedURL,
+					connOwned: true,
+				},
+			},
+		},
+		{
+			name: "with options",
```
embano1, line 51: can you please be a bit more descriptive with your test cases? they're pretty generic and I don't know what they assert/test without reading the code, making reviews/debugging harder

protocol/nats_jetstream/v3/protocol.go
embano1: Sorry, I somehow missed that you already had used a buffered channel here

test/integration/nats_jetstream/v3/nats_test.go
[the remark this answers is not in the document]
embano1: Got it. Does it make sense to add `t.Helper()` in the function body then?

embano1 (commented):
Thx again for making the changes! I think this is close to merging, major feedback is now about:

- error types, I think we should use a different approach here to keep the code simple, maintainable, and easy to use for developers, such as using [this](https://github.com/aws-controllers-k8s/eventbridge-controller/blob/d234915bd60fed78224894fc580625381538c9b9/pkg/resource/rule/hooks.go#L29-L43) approach (with exported type though)
- options, room for some developer experience improvements to avoid confusion how to use the protocol

protocol/nats_jetstream/v3/options.go
stephen-totty-hpe: done

test/integration/nats_jetstream/v3/nats_test.go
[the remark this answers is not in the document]
stephen-totty-hpe: changed to executeProtocol() and added t.Helper()

protocol/nats_jetstream/v3/protocol.go
stephen-totty-hpe: `func New(ctx context.Context, opts ...ProtocolOption) (*Protocol, error) `
calls
`if p.jetStream, errConnection = jetstream.New(p.conn, p.jetSteamOpts...); errConnection != nil {`

protocol/nats_jetstream/v3/options.go
stephen-totty-hpe: added validationError.  I added it as not exported type.

protocol/nats_jetstream/v3/protocol.go
stephen-totty-hpe: The v2 version had this concept.  The idea is that if a connection was passed in, we would use it and then exit without closing the connection (since it was opened before use). 

With a URL, we create a  connection and then close it when Close() is called:
`		p.conn.Close()`

I am not sure if users would complain if we closed their "passed in connection".

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,110 @@
...
+	// error messages
+	messageNoConnection                 = "URL or nats connection must be given."
+	messageConflictingConnection        = "URL and nats connection were both given."
+	messageNoConsumerConfig             = "No consumer config was given."
+	messageNoFilterSubjects             = "No filter subjects were given."
+	messageMoreThanOneStream            = "More than one stream for given filter subjects."
+	messageNoSendSubject                = "Cannot send without a NATS subject defined."
+	messageMoreThanOneConsumerConfig    = "More than one consumer config given."
+	messageReceiverOptionsWithoutConfig = "Receiver options given without consumer config."
+	messageSenderOptionsWithoutSubject  = "Sender options given without send subject."
```
embano1, line 31: nit: they're not used in this file. Suggesting to move all validation logic and errors to `validation.go` to keep the code easy to follow

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,310 @@
...
+
+// validationError is returned when an invalid option is given
+type validationError struct {
+	field   string
+	message string
+}
+
+// Error returns a message indicating an error condition, with the nil value representing no error.
+func (v validationError) Error() string {
+	return fmt.Sprintf("invalid Spec: %q: %s", v.field, v.message)
```
embano1, line 296: nit: copy paste typo ;)

```suggestion
	return fmt.Sprintf("invalid parameters provided: %q: %s", v.field, v.message)
```

protocol/nats_jetstream/v3/protocol.go
embano1: > I am not sure if users would complain if we closed their "passed in connection".

Well, this is what I was referring to earlier: users are in full control (and it's actually what your doc string says for `Close()`). So users can decide when they underlying connection (either ours or their provided) will be closed.

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,310 @@
...
+	for _, fn := range opts {
+		if err := fn(p); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// validateOptions runs after all options have been applied and makes sure needed options were set correctly.
+func (p *Protocol) validateOptions() error {
```
embano1, line 200: nit: see comment in `options.go` where I suggested to move whole validation logic into a separate file (same package).

protocol/nats_jetstream/v3/protocol.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+	return finalStream, nil
+}
+
+type msgErr struct {
+	msg binding.Message
+}
+
+// validationError is returned when an invalid option is given
+type validationError struct {
```
embano1, line 289: nit: do you intentionally want to keep this internal? Fine with me to not increase the API surface for now :)

samples/nats_jetstream/v3/receiver/main.go
```diff
@@ -0,0 +1,82 @@
...
+	natsURL := "nats://localhost:4222"
+	natsSubject := "sample"
+	natsStream := "stream"
+
+	createStream(natsURL, natsStream, natsSubject)
+	consumerOpt := cejsm.WithConsumerConfig(&jetstream.ConsumerConfig{FilterSubjects: []string{natsSubject}})
+	urlOpt := cejsm.WithURL(natsURL)
+	protocol, err := cejsm.New(ctx, consumerOpt, urlOpt)
+	if err != nil {
+		log.Fatalf("failed to create nats protocol, %s", err.Error())
```
embano1, line 32: nit: typically use `:` here (applies to others as well)

```suggestion
		log.Fatalf("failed to create nats protocol: %s", err.Error())
```

samples/nats_jetstream/v3/receiver/main.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	defer protocol.Close(ctx)
+
+	c, err := cloudevents.NewClient(protocol)
+	if err != nil {
+		log.Fatalf("failed to create client, %s", err.Error())
+	}
+
+	for {
```
embano1, line 42: why the for loop?

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,110 @@
...
+// WithURL creates a connection to be used in the protocol sender and receiver
+func WithURL(url string) ProtocolOption {
+	return func(p *Protocol) error {
+		p.url = url
+		p.connOwned = true
+		return nil
+	}
+}
+
+// WithNatsOptions is used in with WithURL() when creating a NATS connection
```
embano1, line 46:
```suggestion
// WithNatsOptions can be used together with WithURL() to specify NATS connection options
```

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,110 @@
...
+func WithNatsOptions(natsOpts []nats.Option) ProtocolOption {
+	return func(p *Protocol) error {
+		p.natsOpts = natsOpts
+		return nil
+	}
+}
+
+// WithConnection uses the provided connection in the protocol sender and receiver
```
embano1, line 54:
```suggestion
// WithConnection uses the provided connection in the protocol sender and receiver. Must not be used together with WithURL.
```

protocol/nats_jetstream/v3/options.go
```diff
@@ -0,0 +1,110 @@
...
+	messageNoSendSubject                = "Cannot send without a NATS subject defined."
+	messageMoreThanOneConsumerConfig    = "More than one consumer config given."
+	messageReceiverOptionsWithoutConfig = "Receiver options given without consumer config."
+	messageSenderOptionsWithoutSubject  = "Sender options given without send subject."
+)
+
+// ProtocolOption provides a way to configure the protocol
+type ProtocolOption func(*Protocol) error
+
+// WithURL creates a connection to be used in the protocol sender and receiver
```
embano1, line 37:
```suggestion
// WithURL creates a connection to be used in the protocol sender and receiver. Must not be used together with WithConnection.
```

embano1: You can also use the "mutually exclusive" wording int hose connection-related options that you did below for others

samples/nats_jetstream/v3/receiver/main.go
stephen-totty-hpe: removed

protocol/nats_jetstream/v3/protocol.go
stephen-totty-hpe: moved

protocol/nats_jetstream/v3/protocol.go
stephen-totty-hpe: I changed the comments for Close() and added a few new comments.  I removed the `connOwned` member and only use the `url` member.

embano1 (approved):
Thank you so much for your contribution and patience along the review! I really like the new NATS JS API you created here!

I think your other JS-related PRs can now be closed?

PR #1114, opened 2025-01-09
chore(deps): Bump github.com/stretchr/testify to v1.10.0

aalekseevx:
Updating testify introduces stricter rules for using `require.NotSame`. The implementation now enforces passing pointers, making certain checks that were previously always valid unnecessary. These redundant checks have been removed from the codebase.`


duglin:
Actually, it might be more dirs. Try:  hack/go-mod-tidy.sh
```
=== RUN   TestEvent_Clone
    event_test.go:533: 
                Error Trace:    /root/go/src/github.com/cloudevents/sdk-go/v2/event/event_test.go:533
                Error:          Both arguments must be pointers
                Test:           TestEvent_Clone
    event_test.go:539: 
                Error Trace:    /root/go/src/github.com/cloudevents/sdk-go/v2/event/event_test.go:539
                Error:          Both arguments must be pointers
                Test:           TestEvent_Clone
```

aalekseevx:
@duglin Thank you for your help! Could you please review the new fixes and let me know if they work for you?

duglin (approved):
LGTM

ping @embano1 for the merge

v2/types/value_test.go
```diff
@@ -277,15 +275,13 @@ func TestCloneTimestampPointer(t *testing.T) {
...
 	require.NotSame(t, original, cloned)
-	require.NotSame(t, original.Time, cloned.Time)
 }
 
 func TestCloneTime(t *testing.T) {
 	original := time.Now()
 	cloned := types.Clone(original).(types.Timestamp)
 
 	require.Equal(t, original, cloned.Time)
-	require.NotSame(t, original, cloned.Time)
```
duglin, line 288: I don't understand the original... how can it be equal and not equal?

aalekseevx: I believe it was intended to check whether the values are equal and their addresses differ.

duglin: that really hurts my head :-)
Sure reads like a bug in "require" to me.
So, in this case did you remove it because "NotSame" fails or because it's redundant?

duglin: Ah "same" vs "equal". I missed that!

aalekseevx: Yes! Removed because it seems redundant for me to check that `&original` и `&cloned.Time` are 2 different addresses

aalekseevx:
@duglin, @embano1 are there any blockers for merge?

embano1:
Fantastic, thx"

embano1 (approved):
LGTM

PR #1137, opened 2025-03-24
fix automerge by giving better names

duglin:
better names will allow us to add them to the "auto-merge" checker. Today more than one check shows up as the same name which makes it harder to list them all.

Also, revert dependabot's latest change since it breaks things.


duglin:
@embano1

embano1 (approved):

PR #1145, opened 2025-04-19
Dependency Management Improvements

embano1:
This PR replaces Dependabot with a custom dependency update workflow to better handle our multi-module repository structure.

## Changes:

- Added a new script (`hack/update-deps.sh`) to automatically update Go dependencies across all modules
- Created a GitHub workflow (`update-dependencies.yml`) that runs weekly to update dependencies and create PRs
- Modified Dependabot configuration to only handle GitHub Actions updates, as Go dependencies are now managed by our custom workflow
- Removed stale and broken OpenTelemetry samples that were no longer maintained
- Updated Go dependencies to latest compatible versions, as well as Go version and toolchain across all modules (`go.mod`) to `1.23.*`
- Updated README.md to reflect current Go version support (1.23+)

This change provides a more reliable and consistent approach to dependency management across our multi-module repository structure, ensuring all modules stay up-to-date with the latest security patches and improvements.


hack/update-deps.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# Process each directory
+COUNTER=1
+for DIR in $DIRS; do
+  echo "[$COUNTER/$DIR_COUNT] Processing $DIR"
+
+  # Change to the directory
+  pushd "$DIR" >/dev/null
+
+  # disable go toolchain switching
```
duglin, line 40: I'm curious, why do we need this?
If we're using the same go version for everything maybe we should add a script to check to make sure all go.mod files use the same version. Or am I missing something that would prevent that?

embano1: Fixed, sorry I was still working on this :) PR is now ready for review. Couldn't fully test the new script yet because of the "limited" testing capabilities with Github Actions ;) We'll see after merge...

duglin:
Will this create a PR even if nothing has changed?

embano1:
> Will this create a PR even if nothing has changed?

Nope, the action is smart in many ways - no-op is the default.

embano1:
I did a dry run in my repo, worked like a charm. Just didn't test if there's a dep change and how the PR will look like. But I used the action in other repos before and worked well.

embano1:
Also: should make your life and relationship to Go dep management better 😜

duglin:
> Also: should make your life and relationship to Go dep management better 😜

Sure! Next you'll want me to believe in unicorns

duglin (approved):

Issue #1161, opened 2025-05-28
Event `data` field is not marshaled as a JSON value but string when DataContentType=application/cloudevents+json


alank-ps:
According to https://github.com/cloudevents/spec/blob/main/cloudevents/formats/json-format.md#311-payload-serialization,

> If the datacontenttype declares the data to contain JSON-formatted content, a JSON serializer MUST translate the data value to a [JSON value](https://tools.ietf.org/html/rfc7159#section-3), and use the member name data to store it inside the JSON representation. The data value MUST be stored directly as a JSON value, rather than as an encoded JSON document represented as a string. An implementation MAY fail to serialize the event if it is unable to translate the runtime value to a JSON value.

However,  the `WriteJson()` function in in `event_marshal.go` only consider `ApplicationJSON` and `TextJSON` (and also default empty) data content type as `isJson=true`. 

https://github.com/cloudevents/sdk-go/blob/main/v2/event/event_marshal.go#L153

It results in writing "structured" type `ApplicationCloudEventsJSON` as a string, not a JSON value.

https://github.com/cloudevents/sdk-go/blob/main/v2/event/event_marshal.go#L172

alank-ps:
Hi @embano1 Thanks for reviewing my PR #1162. Do you have an estimated timeline about upcoming releases by any chance?

alank-ps:
> Ahhhh, sorry. Please hold on. I find the same problem on the other side - `event_unmarshal.go` as well...I am raising another PR for that. Apologies

problem solved. 
@embano1 @duglin Looking forward to new release. Thanks again!!!

embano1:
https://github.com/cloudevents/sdk-go/releases/tag/v2.16.1

PR #1162, opened 2025-05-28
fix: WriteJson should write data as a JSON value when dataContentType…

alank-ps:
…=application/cloudevents+json (and batch)

https://github.com/cloudevents/sdk-go/issues/1161

### DIFF
--- a/v2/event/event_marshal.go
+++ b/v2/event/event_marshal.go
@@ -150,7 +150,9 @@ func WriteJson(in *Event, writer io.Writer) error {
 			mediaType = strings.TrimSpace(strings.ToLower(contentType[0:i]))
 		}
 
-		isJson := mediaType == "" || mediaType == ApplicationJSON || mediaType == TextJSON
+		isJson := mediaType == "" ||
+			mediaType == ApplicationJSON || mediaType == TextJSON ||
+			mediaType == ApplicationCloudEventsJSON || mediaType == ApplicationCloudEventsBatchJSON
 
 		// If isJson and no encoding to base64, we don't need to perform additional steps
 		if isJson && !isBase64 {
--- a/v2/event/event_marshal_test.go
+++ b/v2/event/event_marshal_test.go
@@ -165,7 +165,97 @@ func TestMarshal(t *testing.T) {
 				"source":          "http://example.com/source",
 			},
 		},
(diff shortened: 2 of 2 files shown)


v2/event/event_marshal.go
```diff
@@ -150,7 +150,9 @@ func WriteJson(in *Event, writer io.Writer) error {
 			mediaType = strings.TrimSpace(strings.ToLower(contentType[0:i]))
 		}
 
-		isJson := mediaType == "" || mediaType == ApplicationJSON || mediaType == TextJSON
+		isJson := mediaType == "" ||
+			mediaType == ApplicationJSON || mediaType == TextJSON ||
+			mediaType == ApplicationCloudEventsJSON || mediaType == ApplicationCloudEventsBatchJSON
 
 		// If isJson and no encoding to base64, we don't need to perform additional steps
 		if isJson && !isBase64 {
```
embano1, line 158: nit: suggesting for readability/future extension

```suggestion
jsonMediaTypes := map[string]bool{
    "":                                 true,
    ApplicationJSON:                   true,
    TextJSON:                          true,
    ApplicationCloudEventsJSON:       true,
    ApplicationCloudEventsBatchJSON:  true,
}

if jsonMediaTypes[mediaType] && !isBase64 {

```

alank-ps: thanks for the neater go :bow:

v2/event/event_marshal_test.go
```diff
@@ -165,7 +165,97 @@ func TestMarshal(t *testing.T) {
 				"source":          "http://example.com/source",
 			},
 		},
-		"struct data v1.0": {
+		"structured mode data v1.0": {
```
duglin, line 168: Can you add a test that sends a string that just happens to look like json? Just to make sure it stays a string on both ends.

alank-ps: Are you referring to this test case?
https://github.com/cloudevents/sdk-go/blob/d35a778aafda9f0c483e62baf69125d90d181048/v2/event/event_marshal_test.go#L338
If so, it is already there and no regression (test still pass ✅ )

embano1: I think this is good. Merging.

PR #1164, opened 2025-06-06
fix: consumeData(and consumeDataAsBytes) should consider structued mode JSON …

alank-ps:
…content types also JSON

#1161

### DIFF
--- a/v2/event/content_type.go
+++ b/v2/event/content_type.go
@@ -14,11 +14,9 @@ const (
 	ApplicationCloudEventsBatchJSON = "application/cloudevents-batch+json"
 )
 
-type ContentType string
-
-// IsJSON returns true if the content type is a JSON type.
-func (c ContentType) IsJSON() bool {
-	switch c {
+// isJSON returns true if the content type is a JSON type.
+func isJSON(contentType string) bool {
+	switch contentType {
 	case ApplicationJSON, TextJSON, ApplicationCloudEventsJSON, ApplicationCloudEventsBatchJSON:
 		return true
 	case "":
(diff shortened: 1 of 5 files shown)


v2/event/event_unmarshal.go
```diff
@@ -416,6 +415,16 @@ func consumeData(e *Event, isBase64 bool, iter *jsoniter.Iterator) error {
 	return nil
 }
 
+func isJson(contentType string) bool {
```
embano1, line 418: Can we combine and reuse this one in the other change you did earlier to keep it consistent and reduce code duplication?

alank-ps:
Hi @embano1 , do you mind having another look?

v2/event/content_type.go
```diff
@@ -14,6 +14,20 @@ const (
 	ApplicationCloudEventsBatchJSON = "application/cloudevents-batch+json"
 )
 
+type ContentType string
```
embano1, line 17: Hm, can we 1/ not export this and 2/ simplify this into just a simple function? See below.

v2/event/content_type.go
```diff
@@ -14,6 +14,20 @@ const (
...
+
+// IsJSON returns true if the content type is a JSON type.
+func (c ContentType) IsJSON() bool {
```
embano1, line 20: Instead of introducing an exported type, how about:

```suggestion
func isJSON(contentType string) bool { ... }
```

embano1 (changes requested):

v2/event/content_type.go
alank-ps: The unit test cannot access it then :(

alank-ps: I can ditch the unit test if preferred

alank-ps: @embano1 updated as suggested, thanks 🙇

PR #1177, opened 2025-08-17
Add SQS protocol

### LINKED ISSUE #377: AWS SNS/SQS Binding
Since you are supporting a lot of bindings out there I would like to ask if you have any plan on supporting AWS SNS/SQS.
AWS SNS/SQS is really used a lot when working in their cloud and i believe the project woud benefit from this addition.

yvigara:
Resolves #377
Implement SQS protocol and provides examples.


yvigara:
@duglin & @embano1 is there anything else you'd need to progress this PR?
I could try joining a community call to answer your questions. It's just that the time of the call is not very friendly for Australians 😅

embano1:
Hi and sorry for my delay, distracted with work. The reason a SPEC might be useful is that adapters might not help here as they're relevant in consumers/transports which can actively reshape data structures to CE. SQS doesn't qualify here, but I think a Lambda adapter might be a good candidate since it's often used with SQS. EventBridge also doesn't "understand" CloudEvents today, so users currently have to use custom input transformers to reshape JSON to CloudEvents, but those are limited when it comes to logic required to handle conditional formatting and error cases. Or @duglin are you saying that adapters can also be used to tell producers how to convert CloudEvents into the transport format (i.e., the reverse of the examples used in the adapters today which go from custom-to-CE)?

I think the right approach here for SQS and EventBridge is to create SPECs and host them at AWS (similar to Google Pub/Sub). Especially the SPEC for EventBridge would be pretty lightweight currently (only structured mode support, mappings between EventBridge and CloudEvents fields, etc.) but allows us to evolve over time in case those services add native support for CloudEvents. As you call out, SQS has all the properties to model binary/structured encodings and so a SPEC makes sense to also document the limits.

I can try to talk to the SQS and Lambda team and see if they'd be interested in supporting here (especially with repositories, reviews and maintenance). 

@duglin do we have a document which describes how SPECs can evolve e.g., a SPEC for a transport with limited/no support CE today?

PR #1180, opened 2025-08-18
Refactor pubsub protocol to use new upstream v2 library

philicious:
The GCP GO SDK introduced a new v2 version of their PubSub library with the v1.50 release of the pubsub pkg. Deprecating the old interface and encouraging to upgrade within a 1yr grace period. See https://github.com/googleapis/google-cloud-go/pull/12218 for reference

For downstream users (like me) being able to update to the new v2 library while also using the cloudevents SDK, its necessary to upgrade the pubsub protocol in the cloudevents SDK or we cannot pass in types to funcs like `WithReceiveSettings` anymore.

Sooo I took the chance and migrated the code to use pubsub/v2. 
Noteworthy changes:
- upstream dropped the `Exists` functions on Topic/Subscription and suggest optimistically using them and handling NotFound errors.
- the CreateX funcs got moved to an admin plane
- Topic and Subscription names now need to be fully-qualified names, including projectID
- Topic and Subscription types got renamed to Publisher/Subscriber

(https://github.com/googleapis/google-cloud-go/blob/main/pubsub/MIGRATING.md)

PS: as a separate commit, I added a fix for the existing ReceiveSettings assertion that didnt actually test for the passed in ReceiveSettings anymore as the NumGoRoutines override was removed after these assertions were initially added

### DIFF
--- a/protocol/pubsub/v2/internal/connection_test.go
+++ b/protocol/pubsub/v2/internal/connection_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"google.golang.org/api/option"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/protobuf/types/known/durationpb"
 )
 
@@ -47,9 +48,9 @@ func (pc *testPubsubClient) New(ctx context.Context, projectID string, failureMa
 	var err error
 	var conn *grpc.ClientConn
 	if len(failureMap) == 0 {
-		conn, err = grpc.Dial(pc.srv.Addr, grpc.WithInsecure())
+		conn, err = grpc.NewClient(pc.srv.Addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	} else {
-		conn, err = grpc.Dial(pc.srv.Addr, grpc.WithInsecure(), grpc.WithUnaryInterceptor(makeFailureIntercept(failureMap)))
+		conn, err = grpc.NewClient(pc.srv.Addr, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithUnaryInterceptor(makeFailureIntercept(failureMap)))
 
 	}
 	if err != nil {
--- a/protocol/pubsub/v2/protocol_test.go
+++ b/protocol/pubsub/v2/protocol_test.go
@@ -24,7 +24,7 @@ type testPubsubClient struct {
 
 func (pc *testPubsubClient) NewWithAttributesInterceptor(ctx context.Context, projectID, orderingKey string) (*pubsub.Client, error) {
 	pc.srv = pstest.NewServer()
-	conn, err := grpc.Dial(pc.srv.Addr, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithUnaryInterceptor(customAttributesInterceptor(map[string]string{
+	conn, err := grpc.NewClient(pc.srv.Addr, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithUnaryInterceptor(customAttributesInterceptor(map[string]string{
 		"Content-Type":        "text/json",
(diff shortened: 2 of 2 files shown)


tyrone-anz:
Just here to say we have same issue. We are unable to upgrade to libraries using pubsub/v2 because the cloudevents SDK is not yet supporting it.

protocol/pubsub/v2/go.mod
```diff
@@ -8,13 +8,14 @@ replace github.com/cloudevents/sdk-go/v2 => ../../../v2
 
 require (
 	cloud.google.com/go/pubsub v1.50.0
```
embano1, line 10: I was expecting this to be removed too? Where are we still using it?

philicious: ah missed a file

philicious:
> > Topic and Subscription names now need to be fully-qualified names, including projectID
> 
> Can you please update the public config parameters/related options to indicate this and reduce developer experience issues?

imho there is no change in cloudevents public config by that change. It still requires a ProjectID and topic/subscription ID as formerly, which are then passed down to PubSub SDK. But now they get concat before being passed down and formerly PubSub SDK did that internally I assume.

> > Topic and Subscription types got renamed to Publisher/Subscriber
> 
> Seems those are mainly used internally and not really breaking the public API (despite the upgrade to v2 obviously being considered a breaking change)?

Correct: The new `Publisher` and `Subscriber` types loosely reference a Topic/Subscription and some funcs got moved from the latter to the new types. However those new handlers are not exposed to the cloudevents public interface

embano1:
> imho there is no change in cloudevents public config by that change. It still requires a ProjectID and topic/subscription ID as formerly

But IMHO that contradicts your comment `now need to be fully-qualified names, including projectID` which seemed not to be the case before? If these strings now require a different (semantical) format, it should be documented in the option/API and would be considered breaking.

protocol/pubsub/v2/internal/connection.go
```diff
@@ -89,8 +93,35 @@ const (
...
+func topicExists(ctx context.Context, client *pubsub.Client, topicID string) (bool, error) {
+	// Check if the topic exists.
+	_, err := client.TopicAdminClient.GetTopic(ctx, &pubsubpb.GetTopicRequest{
+		Topic: fmt.Sprintf("projects/%s/topics/%s", client.Project(), topicID),
+	})
+	if err != nil {
+		if st, ok := status.FromError(err); ok && st.Code() == codes.NotFound {
+			return false, nil // Topic does not exist.
+		}
+		return false, fmt.Errorf("unable to get topic %q, %v", topicID, err)
```
embano1, line 108: nit: use `%w` for error wrapping?

embano1 (approved):
Just a nit, otherwise LGTM if CI passes and assuming you've verified this change in a real PubSub environment?

protocol/pubsub/v2/internal/connection.go
philicious: good idea but since the calling function bubbles-up any caught errors and wraps them again with %v, it would need changes there too. So I'll leave it as is as its likely not used with `errors.Is` anywhere or would have caused issues before

philicious:
> Thx, I'll merge once #1183 is merged as we're upgrading tooling and requiring updated workflows to pass.

I rebased my branch onto latest master after you merged. I also added a little chore commit for replacing deprecated grpc calls in two of the tests. Hope you dont mind

embano1 (approved):

embano1:
Thx, running CI with auto-merge.

philicious:
@embano1 I would like to kindly ask for a new release as there doesnt seem to be any objection :) Thanks in advance for your efforts!

embano1:
On it. Sorry for my silence, been traveling.

embano1:
Done, just don't have the time to create release notes ATM: https://github.com/cloudevents/sdk-go/releases/tag/v2.16.2

Issue #1185, opened 2025-09-05
Logging is missing on error convert message to event


nsaltun:
It could be really difficult to understand why an event is not consuming when there is an issue with converting message to event due to not logging the error in "cloudevents/sdk-go/v2"

It would be great to add a log there to be aware why event is not going to handler function. Currently there is no log movement in console when such an error is occuring so it's not possible to understand the cause without debugging

embano1:
Which protocol/binding are you using. I can see that we propagate any error up to the protocol which might not be correctly handled/overwritten though.

nsaltun:
@canni @embano1 Actually I think I managed to log it by implementing my own messagebinding wrapper in protocol. I'm able to log error on my custom Finish(error) method implementation which is calling after message receiving completed by cloudevent sdk.

nsaltun:
Closed the PR now. Thanks a lot for the discussion and guidance!

PR #1186, opened 2025-09-05
fix: add log for convert message to event error

nsaltun:
Resolves https://github.com/cloudevents/sdk-go/issues/1185


embano1:
Let's further discuss the issue first (I left a comment). I don't think we should add logging to the critical path here but rather address any issue if the returned error is not correctly propagated to/handled in the protocol.

nsaltun:
Closed as not needed. Check issue resolution #1185

PR #1189, opened 2025-09-17
add log options for mqtt client

skeeey:


yanmxa:
Looks good to me. Could you take a look as well, @embano1

PR #1191, opened 2025-09-19
change jetstream integration CI service from bitnami to nats:latest

stephen-totty-hpe:
bitnami public images are being removed September 29th.

see: https://github.com/bitnami/containers/issues/83267

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -46,7 +46,7 @@ jobs:
 
       jetstream:
         image: nats:latest
-        options: "-js"
+        options: -js
         ports:
           - 4223:4223
 


stephen-totty-hpe:
> @stephen-totty-hpe sorry, why does that not work? Also, if you still plan to work on it, please use at least a tag (ideally SHA) - I know in the past maintainers didn't follow that best practice, we should though ;)

Seems to be a bug or missing feature in github actions.  The `nats` image when using `jetstream`  requires the `-js` argument to the command.  However, github actions when using `options` in the yaml modifies the `docker` arguments, not the `command` arguments.

Ideally, if github would merge PRs mentioned in my comments, this would not be an problem.

PR #1202, opened 2025-11-09
feat: support to inject tracerProvider

minuk-dev:
- There is no option to inject tracerProvider instead of global provider
- So, add `WithTracerProvider` as an option.

### DIFF
--- a/observability/opentelemetry/v2/client/otel_observability_service.go
+++ b/observability/opentelemetry/v2/client/otel_observability_service.go
@@ -24,8 +24,7 @@ import (
 
 // OTelObservabilityService implements the ObservabilityService interface from cloudevents
 type OTelObservabilityService struct {
-	traceProvider trace.TracerProvider
-
+	traceProvider        trace.TracerProvider
 	tracer               trace.Tracer
 	spanAttributesGetter func(cloudevents.Event) []attribute.KeyValue
 	spanNameFormatter    func(cloudevents.Event) string


observability/opentelemetry/v2/client/otel_options.go
```diff
@@ -37,6 +38,15 @@ func WithSpanNameFormatter(nameFormatter func(cloudevents.Event) string) OTelObs
 	}
 }
 
+// WithTracerProvider sets the tracer provider to use for creating spans.
+func WithTracerProvider(tracerProvider trace.TracerProvider) OTelObservabilityServiceOption {
+	return func(os *OTelObservabilityService) {
+		if tracerProvider != nil {
```
duglin: Why not let them set it to 'nil' if they want?

observability/opentelemetry/v2/client/otel_observability_service.go
```diff
@@ -24,6 +24,8 @@ import (
 
 // OTelObservabilityService implements the ObservabilityService interface from cloudevents
 type OTelObservabilityService struct {
+	traceProvider trace.TracerProvider
+
```
duglin: Can you remove this blank line? If anything I would expect 'traceProvider' and 'tracer' to be grouped and the blank line after 'tracer'. But I'm not sure we need a line break with only 4 fields

observability/opentelemetry/v2/client/otel_observability_service.go
```diff
@@ -47,6 +46,12 @@ func NewOTelObservabilityService(opts ...OTelObservabilityServiceOption) *OTelOb
 		opt(o)
 	}
 
+	o.tracer = o.traceProvider.Tracer(
```
duglin: Would it be better/safer to check o.traceProvider for 'nil' and either set tracer to nil or call o.traceProvider.Tracer ?

observability/opentelemetry/v2/client/otel_options.go
minuk-dev: To prevent, panic.
If someone want to disable trace, we can use noop.TracerProvider.

https://github.com/open-telemetry/opentelemetry-go/blob/trace/v1.38.0/trace/noop/noop.go#L36

observability/opentelemetry/v2/client/otel_observability_service.go
minuk-dev: I followed the below rule. I think it's very useful way to set a guard.

https://github.com/open-telemetry/opentelemetry-go-contrib/issues/1104

ref. https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1107

PR #1252, opened 2026-03-09
fix(docs): correct list formatting in Go further section

archy-rock3t-cloud:
The "Go further" section uses `_. Check out...` which does not render as a
markdown list. Replaced with proper numbered list formatting.


embano1:
Thx, please rebase onto `main` so we can merge.

archy-rock3t-cloud:
@embano1 done

PR #1272, opened 2026-04-09
fix: preserve HTTP response body after retries

### LINKED ISSUE #1238: HTTP protocol doesn't return body response when retrying
**Description**
If I send event with HTTP protocol (without retrying), in the Result I can see also body response. This is useful to understand what went wrong.

However, if I send event with HTTP Protocol and passing retry options via context like 
```go
Send(cloudevents.ContextWithRetriesExponentialBackoff(ctx, time.Second, 3), evt)
```
I'm getting back only status code. Not the body itself.

**Source of bug**
I believe the issue is on line 181 in ⬇️ because you are expecting `err` is type of `http.Result`. But after retrying, it is `protocol.Result`. 

https://github.com/cloudevents/sdk-go/blob/44dff966be6a0f804261ceb68930d333b13a507d/v2/protocol/http/protocol.go#L179-L197

officialasishkumar:
Fixes: https://github.com/cloudevents/sdk-go/issues/1238

## Summary

- add `RetriesResult.Unwrap()` so wrapped HTTP results remain discoverable through standard error traversal
- add regression coverage for unwrapping retries to `*http.Result`
- add a retry-path `Send()` test that checks the final response body is preserved

## Testing

- `go test ./protocol/http`

### DIFF
--- a/v2/protocol/http/protocol_retry_test.go
+++ b/v2/protocol/http/protocol_retry_test.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"net/http"
 	"net/http/httptest"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


embano1:
@officialasishkumar needs rebase and signed-off commits.

kunalworldwide (commented):
Adding `Unwrap()` to `RetriesResult` fixes the error-traversal issue without breaking the existing API. The test shows that `protocol.ResultAs(err, &result)` now works after retries, which is the real-world use case from #1238.

Preserving the final response body in the retry path is also handled — the test confirms the body can still be read after `Send` returns. This avoids the common pitfall where wrapping a result and then retrying loses the original HTTP response for callers.

LGTM.

Issue #1273, opened 2026-04-10
Support Other JSON Content-Types in content_type.go


Evynglais:
Hi,

At the moment, `isJSON(contentType string) bool` in [content_type.go](https://github.com/cloudevents/sdk-go/blob/main/v2/event/content_type.go) only allows a specific subset of JSON Content-Types*. This means (for example), we are unable to use `application/protobuf+json` unless we escape the data and pass it as a single string element.

*ApplicationJSON, TextJSON, ApplicationCloudEventsJSON, ApplicationCloudEventsBatchJSON

This list of types should be expanded (or preferably, the method refactored to avoid hard-coding a list of strings) so a more generic set of JSON Content-Types can be supported. Alternatively, a way to register our own Content-Type so that it can be treated as a JSON object rather than a string.

Thanks,
Evynglai

Evynglais:
## Examples 

### Structured Single

```json
"Content-Type": "application/cloudevents+json"
{
	"specversion": "1.0",
	"id": "event",
	"source": "test/source",
	"type": "test.type",
	"datacontenttype": "application/protobuf+json",
	"data": { "identifier": "event-1" }
}
```

Result:

```shell
failed to parse CloudEvents HTTP response into single event: unexpected data payload for media type "application/protobuf+json", expected a string: ReadString: expects " or n, but found {, error found in #10 byte of ...|	"data": { "identifi|..., bigger context ...|tenttype": "application/protobuf+json",
        	            		"data": { "identifier": "batch-event-1" }
        	            	}|...
```

### Structured Batch

```json
"Content-Type": "application/cloudevents-batch+json"
[{
	"specversion": "1.0",
	"id": "batch-event-1",
	"source": "test/source",
	"type": "test.type",
	"datacontenttype": "application/protobuf+json",
	"data": { "identifier": "batch-event-1" }
},
{
	"specversion": "1.0",
	"id": "batch-event-2",
	"source": "test/source",
	"type": "test.type",
	"datacontenttype": "application/protobuf+json",
	"data": { "identifier": "batch-event-2" }
}]
```

Result:

```shell
failed to parse CloudEvents HTTP response into batched events: unexpected data payload for media type "application/protobuf+json", expected a string: ReadString: expects " or n, but found {, error found in #10 byte of ...|	"data": { "identifi|..., bigger context ...|tenttype": "application/protobuf+json",
        	            		"data": { "identifier": "batch-event-1" }
        	            	}|...
```

PR #1274, opened 2026-04-10
Support more JSON ContentTypes in isJSON

Evynglais:
Adds support for other JSON Content-Types outside of those defined in `content_type.go`.

See: [Issue](https://github.com/cloudevents/sdk-go/issues/1273)


embano1:
Thx, overall LGTM. Pinging @duglin if there's some SPEC-specifics which we need to consider here.

v2/event/content_type.go
```diff
@@ -5,6 +5,11 @@
 
 package event
 
+import (
+	"mime"
+	"strings"
+)
+
 const (
```
embano1, line 14: Are those `consts` still needed?

Evynglais:
Apologies, had to fix one of the test cases for invalid mediatype parameters. Have set it to ignore parameter issues, since this function doesn't use parameters (it won't block us from checking the subtype). Not sure how I missed that the first time!

v2/event/content_type.go
Evynglais: They're used in the `StringOfApplicationJSON()`, `StringOfApplicationXML()`, etc. functions below (which are pretty widely used), and they're also used directly in numerous places too.

v2/event/content_type.go
```diff
@@ -14,16 +20,26 @@ const (
...
+	// if err is for an invalid media parameter, we can still check the mediatype, since we don't use parameters:
+	if err == nil || errors.Is(err, mime.ErrInvalidMediaParameter) {
+		if _, subtype, ok := strings.Cut(mediaType, "/"); ok {
+			parts := strings.Split(subtype, "+")
+			return parts[len(parts)-1] == "json"
+		}
+	}
+
+	// fallback to checking the suffix of the entire Content-Type:
+	return strings.HasSuffix(contentType, "json")
```
duglin: I might be wrong, but I think needs to check for "/json" and not just "json" because otherwise a contentType of "foojson" would return true and I'm not sure that's what we want. Thoughts?

Let's add a test for "foojson" either way to make sure it always returns the correct result

duglin:
looks good - just one minor suggestion

v2/event/content_type.go
Evynglais: Agreed, added the test and made it check for `/json`. Thanks for the suggestion :)

duglin (approved):

duglin:
I think you need to rebase

Evynglais:
> I think you need to rebase

Done ✅

PR #1286, opened 2026-05-24
fix: omit protobuf data for events without payload

### LINKED ISSUE #1193: JSON marshalling produces invalid JSON when data is a zero length slice

```go
import (
	"log"

)

func main() {
	ev.SetType("type")
	ev.SetSource("source")
	ev.SetID("id")
	ev.DataEncoded = []byte{}

	}

	dta, err := ev.MarshalJSON()
	if err != nil {
	}

	var res ce.Event
	err = res.UnmarshalJSON(dta)
	if err != nil {
	}
}
```

immanuwell:
Fixes #1193

protobuf marshal always sets `data`, even when the event has no payload.
after a protobuf round trip that becomes `DataEncoded=[]byte{}`, then JSON marshal emits busted JSON:

```json
{"specversion":"1.0","id":"id","source":"source","type":"type","datacontenttype":"application/json","data":}
```

Repro on `main`:

```go
ev.SetID("id")
ev.SetSource("source")
ev.SetType("type")

b, _ := format.Protobuf.Marshal(&ev)
var out ce.Event
_ = format.Protobuf.Unmarshal(b, &out)
_, _ = out.MarshalJSON()
```

This patch skips protobuf `data` when `e.Data() == nil`, so no payload stays no payload. small fix, but pretty real.

Tests:
- add a regression test for protobuf -> event -> json round trip with no data
- `go test ./...` in `binding/format/protobuf/v2`
- `go test ./...` in `v2`

### DIFF
--- a/binding/format/protobuf/v2/protobuf.go
+++ b/binding/format/protobuf/v2/protobuf.go
@@ -108,10 +108,12 @@ func ToProto(e *event.Event) (*pb.CloudEvent, error) {
 		}
 		container.Attributes[name] = attr
 	}
-	container.Data = &pb.CloudEvent_BinaryData{
-		BinaryData: e.Data(),
+	if e.Data() != nil {
+		container.Data = &pb.CloudEvent_BinaryData{
+			BinaryData: e.Data(),
+		}
 	}
-	if e.DataContentType() == ContentTypeProtobuf {
+	if e.Data() != nil && e.DataContentType() == ContentTypeProtobuf {
 		anymsg := &anypb.Any{
 			TypeUrl: e.DataSchema(),
 			Value:   e.Data(),
--- a/binding/format/protobuf/v2/protobuf_test.go
+++ b/binding/format/protobuf/v2/protobuf_test.go
@@ -6,6 +6,7 @@
 package format_test
 
 import (
+	"encoding/json"
 	"net/url"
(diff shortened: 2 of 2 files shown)


duglin:
@immanuwell thanks. Fix seems ok but I have a question, why would json Marshal() ever return bad json like that? It doesn't seem possible to me even if the data were nil

duglin:
Since we can't control all of the messages we might receive it feels like we might want to also add protection against receiving these nil port but messages, no?

kunalworldwide (commented):
Good fix. Without this, `ToProto` always sets `Data` to a `BinaryData` wrapper even when the event has no payload. That empty `BinaryData{BinaryData: nil}` gets serialized as a zero-length bytes field in protobuf, which on the Unmarshal side creates an empty `[]byte{}` instead of `nil` — and when that's subsequently marshalled to JSON, you get `"data_base64":""` instead of no `data` field at all, producing invalid JSON for events like heartbeats that intentionally have no payload.

The guard on `DataContentType == ContentTypeProtobuf` is also correct — no point trying to wrap nil data into an `Any` message.

Test is thorough: roundtrips through protobuf marshal/unmarshal, verifies `DataEncoded` is nil, then confirms the JSON output is clean and re-parseable.

LGTM.

PR #1302, opened 2026-07-09
fix(pubsub): nack message when context is cancelled during delivery

AruneshDwivedi:
When the subscriber context is cancelled while a message is still being handed to the inbound channel, the message was silently dropped without Ack or Nack. Pub/Sub then held it in-flight until the ack deadline expired, delaying redelivery by minutes under exactly-once delivery. Nacking on ctx.Done lets the server redeliver immediately.

### DIFF
--- a/protocol/pubsub/v2/protocol.go
+++ b/protocol/pubsub/v2/protocol.go
@@ -206,6 +206,12 @@ func (t *Protocol) startSubscriber(ctx context.Context, sub subscriptionWithTopi
 		select {
 		case t.incoming <- *m:
 		case <-ctx.Done():
+			// The subscriber is shutting down before the message was handed
+			// off. Nack it so Pub/Sub redelivers immediately instead of
+			// waiting out the ack deadline (which with exactly-once delivery
+			// can delay redelivery by minutes).
+			logger.Warnf("context cancelled while delivering message %s; nacking for redelivery", m.ID)
+			m.Nack()
 		}
 	})
 }


kunalworldwide (commented):
Nacking on `ctx.Done()` is the correct fix. Without it, a message in flight during subscriber shutdown sits in Pub/Sub until the ack deadline expires, which under exactly-once delivery can delay redelivery by minutes.

The change is minimal and targeted — only the `startSubscriber` loop is modified. One thing to consider: does the `m.Nack()` call itself need a short timeout or context check, or is the Pub/Sub client nack fast enough that it won't block the shutdown path? Either way, this is a clear improvement.

LGTM.

protocol/pubsub/v2/protocol.go
```diff
@@ -206,6 +206,12 @@ func (t *Protocol) startSubscriber(ctx context.Context, sub subscriptionWithTopi
 		select {
 		case t.incoming <- *m:
 		case <-ctx.Done():
+			// The subscriber is shutting down before the message was handed
+			// off. Nack it so Pub/Sub redelivers immediately instead of
+			// waiting out the ack deadline (which with exactly-once delivery
+			// can delay redelivery by minutes).
+			logger.Warnf("context cancelled while delivering message %s; nacking for redelivery", m.ID)
```
duglin, line 213: Is it "redelivery" or "delivery"? Is it always a retry?

PR #1307, opened 2026-07-21
Fix a panic when header is just "CE-"

duglin:
thanks to @KrisKennawayDD for noticing

### DIFF
--- a/v2/protocol/http/message.go
+++ b/v2/protocol/http/message.go
@@ -117,7 +117,7 @@ func (m *Message) ReadBinary(ctx context.Context, encoder binding.BinaryWriter) 
 		attr := m.version.Attribute(k)
 		if attr != nil {
 			err = encoder.SetAttribute(attr, v[0])
-		} else if strings.HasPrefix(k, prefix) {
+		} else if strings.HasPrefix(k, prefix) && len(k) > len(prefix) {
 			// Trim Prefix + To lower
 			var b strings.Builder
 			b.Grow(len(k) - len(prefix))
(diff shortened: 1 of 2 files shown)


duglin:
ping @embano1

v2/protocol/http/message_test.go
```diff
@@ -155,6 +155,26 @@ func TestMessageTransformDeleteExtension(t *testing.T) {
 	require.Equal(t, eventIn.ID(), id)
 }
 
+// TestReadBinaryEmptyCeHeaderNoPanic verifies that a header whose name is
+// exactly the "Ce-" prefix (i.e. no attribute/extension name follows it)
+// does not cause ReadBinary to panic with an index-out-of-range error.
+// See advisory-01-binary-header-panic-dos.md.
```
embano1: question: where can I find this advisory? also is it "dos" or "docs"?

v2/protocol/http/message.go
```diff
@@ -117,7 +117,7 @@ func (m *Message) ReadBinary(ctx context.Context, encoder binding.BinaryWriter)
 		attr := m.version.Attribute(k)
 		if attr != nil {
 			err = encoder.SetAttribute(attr, v[0])
-		} else if strings.HasPrefix(k, prefix) {
+		} else if strings.HasPrefix(k, prefix) && len(k) > len(prefix) {
```
embano1, line 120: what if k is "ce--"?

duglin: then the 2nd "-" should be part of the name... and normal processing (or error stuff) should kick in.

v2/protocol/http/message_test.go
duglin: let me remove that

v2/protocol/http/message.go
embano1: can you add a test :)

duglin: done

embano1 (approved):

PR #1311, opened 2026-07-21
Blind SSRF via unvalidated WebHook-Request-Callback

duglin:
Blind SSRF via unvalidated WebHook-Request-Callback when AutoACKCallback is enabled.

When a CloudEvents HTTP receiver is configured with `WebhookConfig.AutoACKCallba ck = true`, the built-in `Protocol.OptionsHandler` reads the client-supplied `We bHook-Request-Callback` header and issues an outbound `POST` to that URL via `ht tp.DefaultClient` with no validation of scheme, host, or port. Because the only gate — `WebHook-Request-Origin` validation — compares a header value the client fully controls, an unauthenticated remote attacker can make the server send HTTP
 requests to arbitrary destinations, including loopback, RFC1918, and cloud-meta
data addresses (blind SSRF).

thanks to @KrisKennawayDD for the report

### DIFF
--- a/v2/protocol/http/abuse_protection.go
+++ b/v2/protocol/http/abuse_protection.go
@@ -7,9 +7,12 @@ package http
 
 import (
 	"context"
+	"fmt"
 	cecontext "github.com/cloudevents/sdk-go/v2/context"
 	"go.uber.org/zap"
+	"net"
 	"net/http"
+	"net/url"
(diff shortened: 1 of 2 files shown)


embano1 (dismissed):

duglin:
rebased

embano1 (approved):

PR #1315, opened 2026-08-09
fix: handle zero length slice data in JSON marshalling (fixes #1193)

### LINKED ISSUE #1193: JSON marshalling produces invalid JSON when data is a zero length slice

AruneshDwivedi:
Fixes #1193

When event has data that is a non-nil, zero length slice (empty byte slice), JSON marshalling produces malformed JSON with `{"data":}` which is invalid.

The fix checks for empty DataEncoded when mediaType is JSON and writes `null` instead of raw bytes.

Added test case that reproduces the issue and verifies the roundtrip (marshal -> unmarshal) works correctly.

### DIFF
--- a/v2/event/event_marshal.go
+++ b/v2/event/event_marshal.go
@@ -152,10 +152,15 @@ func WriteJson(in *Event, writer io.Writer) error {
 
 		// If IsJSON and no encoding to base64, we don't need to perform additional steps
 		if isJSON(mediaType) && !isBase64 {
-			stream.WriteObjectField("data")
-			_, err := stream.Write(in.DataEncoded)
-			if err != nil {
-				return fmt.Errorf("error while writing data: %w", err)
+			if len(in.DataEncoded) == 0 {
+				stream.WriteObjectField("data")
+				stream.WriteNil()
+			} else {
+				stream.WriteObjectField("data")
+				_, err := stream.Write(in.DataEncoded)
+				if err != nil {
+					return fmt.Errorf("error while writing data: %w", err)
+				}
 			}
 		} else {
 			if in.Context.GetSpecVersion() == CloudEventsVersionV1 && isBase64 {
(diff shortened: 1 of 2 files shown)


v2/event/event_marshal.go
```diff
@@ -152,10 +152,15 @@ func WriteJson(in *Event, writer io.Writer) error {
 
 		// If IsJSON and no encoding to base64, we don't need to perform additional steps
 		if isJSON(mediaType) && !isBase64 {
-			stream.WriteObjectField("data")
-			_, err := stream.Write(in.DataEncoded)
-			if err != nil {
-				return fmt.Errorf("error while writing data: %w", err)
+			if len(in.DataEncoded) == 0 {
+				stream.WriteObjectField("data")
```
duglin, line 156: Can you move this to be before the "if" since it applies to both paths?

v2/event/event_marshal_1193_test.go
```diff
@@ -0,0 +1,38 @@
...
+	var raw json.RawMessage
+	if err := json.Unmarshal(dta, &raw); err != nil {
+		t.Fatalf("output is NOT valid JSON: %s\noutput: %s", err, string(dta))
+	}
+	t.Logf("JSON output: %s", string(dta))
+
+	var res event.Event
+	if err := res.UnmarshalJSON(dta); err != nil {
+		t.Fatalf("unmarshal failed: %s", err)
+	}
```
duglin, line 37: Can you verify that "data" is in the event with a null value?

duglin:
@AruneshDwivedi also, please sign your commit(s)

PR #1319, opened 2026-08-17
fix(nats_jetstream): add WithStreamName option to bypass stream lookup

### LINKED ISSUE #1240: NATS Jetstream specify stream name
I saw from the source code that the stream name will be automatically resolved based on subject filter. This is convenient but i think in the deployment that has multiple stream for specific subject, the sdk will throw [ErrMoreThanOneStream](https://github.com/cloudevents/sdk-go/blob/main/protocol/nats_jetstream/v3/protocol.go#L241-L272). Is it discouraged to allow for the user to specify the stream name manually through the option func?

AruneshDwivedi:
## Problem

When multiple NATS JetStream streams share the same subject, the SDK throws
ErrMoreThanOneStream during both Send and OpenInbound because it auto-looks up
the stream name via StreamNameBySubject. There is no way for users to specify
the stream name explicitly.

Fixes #1240

## Solution

Add WithStreamName ProtocolOption that sets an explicit stream name on the
Protocol struct. When set, both Send and createJetstreamConsumer skip the
StreamNameBySubject lookup and use the provided name directly.

## Changes

- Add WithStreamName option in options.go
- Add streamName field to Protocol struct
- Skip StreamNameBySubject in Send when streamName is set
- Skip getStreamFromSubjects when streamName is set (returns it directly)

### DIFF
--- a/protocol/nats_jetstream/v3/protocol.go
+++ b/protocol/nats_jetstream/v3/protocol.go
@@ -56,6 +56,7 @@ type Protocol struct {
 	// sender
 	publishOpts []jetstream.PublishOpt
 	sendSubject string
+	streamName  string
 }
 
 // New creates a new NATS protocol.
@@ -121,7 +122,10 @@ func (p *Protocol) Send(ctx context.Context, in binding.Message, transformers ..
 	}()
 
 	if _, err = p.jetStream.StreamNameBySubject(ctx, subject); err != nil {
-		return err
+		// If an explicit stream name was provided, skip the lookup
+		if p.streamName == "" {
+			return err
+		}
 	}
 
 	writer := new(bytes.Buffer)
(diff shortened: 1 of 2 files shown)


protocol/nats_jetstream/v3/protocol.go
```diff
@@ -121,7 +122,10 @@ func (p *Protocol) Send(ctx context.Context, in binding.Message, transformers ..
 	}()
 
 	if _, err = p.jetStream.StreamNameBySubject(ctx, subject); err != nil {
-		return err
+		// If an explicit stream name was provided, skip the lookup
+		if p.streamName == "" {
+			return err
```
embano1, line 127: Can you explain this part to me, specifically:

1) should the check come before the call (skip the `StreamNameBySubject` call?)
2) somehow I don't understand the comment/code because the comment is about "provided" whereas the code check is empty string (inside an `err != nil` check)

embano1 (commented):
Just a question to better understand the code - also, any chance we can test this change (unit/integration)?

PR #1322, opened 2026-08-31
ci: bump Go toolchain to 1.26

### LINKED ISSUE #1321: CI: dependency update workflow fails because Go toolchain is pinned to 1.25
The weekly `Update Dependencies` workflow has failed since `github.com/quic-go/quic-go` released v0.62.0, which declares `go >= 1.26.0`. Our CI pins `go-version: 1.25` and `actions/setup-go` sets `GOTOOLCHAIN=local`, so `go get -u -t ./...` cannot upgrade past that dependency and exits 1.

Failing run: https://github.com/cloudevents/sdk-go/actions/runs/33353546430/job/99371294001

```
go: github.com/quic-go/quic-go@v0.62.0 requires go >= 1.26.0 (running go 1.25.14; GOTOOLCHAIN=local)
##[error]Process completed with exit code 1.
```

`quic-go` reaches us as an indirect dependency of `samples/http`, so no module we publish needs Go 1.26 today. Once the toolchain is bumped, `go get -u` raises the `go` directive in `samples/http/go.mod` to `1.26.0` on its own, and `hack/build-test.sh` builds the samples, so the build, test, lint, format, conformance, integration, observability and release workflows need Go 1.26 as well.

The proposal is to bump `go-version` to `1.26` in all ten workflow files under `.github/workflows/`. The `go` directives in `v2/go.mod` and the protocol modules stay at `1.25.0`, so the minimum Go version for consumers of the SDK does not change.

One related inconsistency, worth a separate fix: `README.md` states a minimum Go version of 1.24, while `v2/go.mod` has required `1.25.0` for some time.

embano1:
# What changed

`go-version` moves from `1.25` to `1.26` in all ten workflow files under `.github/workflows/`. The `go` directives in `v2/go.mod` and the protocol modules are untouched, so this changes the toolchain CI runs, and not the minimum Go version consumers of the SDK need.

The lint job's golangci-lint pin moves from v2.8.0 to v2.12.2, because v2.8.0 cannot analyze code built with Go 1.26. v2.12.2 reports six issues on `./v2`, all cleared here. govet's `inline` analyzer flags three uses of `reflect.Ptr`, a deprecated alias for `reflect.Pointer` since Go 1.18, and the two constants are identical. staticcheck QF1012 flags three calls of the form `Builder.WriteString(fmt.Sprintf(...))`, now `fmt.Fprintf` on the builder, which drops the intermediate string allocation. Ignoring the `Fprintf` return is safe because `strings.Builder` never returns an error, and errcheck is disabled in `.golangci.yaml`.

The README note on the minimum Go version now reads 1.25 instead of 1.24, and points at `protocol/pubsub/v2` needing 1.25.8. The 1.24 claim has been wrong since `v2/go.mod` moved to `go 1.25.0`.

# Why

The weekly `Update Dependencies` workflow has failed since `github.com/quic-go/quic-go` released v0.62.0, which declares `go >= 1.26.0`. CI pinned `go-version: 1.25` and `actions/setup-go` sets `GOTOOLCHAIN=local`, so `go get -u -t ./...` had no way to satisfy that requirement and exited 1 with `requires go >= 1.26.0 (running go 1.25.14; GOTOOLCHAIN=local)`.

`quic-go` reaches the repository as an indirect dependency of `samples/http`. Bumping only `update-dependencies.yml` would fix the dependency job and then break the others, because `go get -u` raises the `go` directive in `samples/http/go.mod` to `1.26.0` on its own and `hack/build-test.sh` builds the samples. All ten workflows move together for that reason.

Closes #1321

# How to verify

Checked locally with go1.26.2 and `GOTOOLCHAIN=local`, which is what `actions/setup-go` gives the runner, and golangci-lint v2.12.2 to match the new pin.

- `cd samples/http && go get -u -t ./...` succeeds in a throwaway worktree, reporting `go: upgraded go 1.25.0 => 1.26.0` and `go: upgraded github.com/quic-go/quic-go v0.61.0 => v0.62.0`. On go1.25 this is the command that exits 1.
- `cd v2 && golangci-lint run` reports `0 issues.`
- `cd v2 && go build ./... && go vet ./...` is clean.
- `cd v2 && go test ./event/... ./types/... ./test/...` passes.
- The workflow change only takes effect for the `Update Dependencies` job after merge, because that job runs on a schedule against `main`. Every other job exercises Go 1.26 on this PR.

### DIFF
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -15,7 +15,7 @@ jobs:
     name: CloudEvents Conformance
     strategy:
       matrix:
-        go-version: [1.25]
+        go-version: [1.26]
         platform: [ubuntu-latest]
 
     runs-on: ${{ matrix.platform }}
--- a/.github/workflows/go-build.yaml
+++ b/.github/workflows/go-build.yaml
@@ -15,7 +15,7 @@ jobs:
     name: Build
     strategy:
       matrix:
-        go-version: [1.25]
+        go-version: [1.26]
         platform: [ubuntu-latest]
 
     runs-on: ${{ matrix.platform }}
--- a/.github/workflows/go-format.yaml
+++ b/.github/workflows/go-format.yaml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version: 1.25
+          go-version: 1.26
           cache-dependency-path: v2/go.sum
         id: go
 
--- a/.github/workflows/go-lint.yaml
+++ b/.github/workflows/go-lint.yaml
@@ -35,5 +35,5 @@ jobs:
         if: steps.golangci_configuration.outputs.files_exists == 'true'
         uses: golangci/golangci-lint-action@ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a # v9.3.0
         with:
-          version: v2.8.0
+          version: v2.12.2
           working-directory: v2
--- a/.github/workflows/go-lint.yaml
+++ b/.github/workflows/go-lint.yaml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version: 1.25
+          go-version: 1.26
           cache-dependency-path: v2/go.sum
         id: go
 
--- a/.github/workflows/go-unit-test.yaml
+++ b/.github/workflows/go-unit-test.yaml
@@ -15,7 +15,7 @@ jobs:
     name: Unit Test
     strategy:
       matrix:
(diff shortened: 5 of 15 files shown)


duglin:
Is it going to be painful for folks to force them to upgrade their golang to a release that's as recent as Feb of this year? Only about 6 months ago.

embano1:
@duglin no worries here, from the PR body

> The `go` directives in `v2/go.mod` and the protocol modules are **untouched**, so this changes the toolchain CI runs, and not the minimum Go version consumers of the SDK need.

And

> The README note on the minimum Go version now reads 1.25 instead of 1.24, and points at `protocol/pubsub/v2` needing 1.25.8. The 1.24 claim has been wrong since `v2/go.mod` moved to `go 1.25.0`.

And 1.25 is 1y old now.

duglin (approved):

embano1:
@duglin thx, btw the reason some workflows "hang" was because CI checks were waiting on now stale workflow versions. Fixed now that it's being merged.

Update: looks like you were quicker ;)

duglin:
yep - I had to edit the settings on the repo to fix it

PR #1324, opened 2026-09-03
feat: add WithStreamName option to bypass stream resolution

### LINKED ISSUE #1240: NATS Jetstream specify stream name

AruneshDwivedi:
Fixes #1240

Adds WithStreamName option to NATS JetStream protocol that allows users to
specify the stream name directly, bypassing automatic stream resolution via
subjects.

This is useful in deployments with multiple streams for the same subject where
the SDK would throw ErrMoreThanOneStream.

### DIFF
--- a/protocol/nats_jetstream/v3/options.go
+++ b/protocol/nats_jetstream/v3/options.go
@@ -88,3 +88,12 @@ func WithPullConsumerOptions(pullConsumeOpts []jetstream.PullConsumeOpt) Protoco
 		return nil
 	}
 }
+
+// WithStreamName allows the user to specify a stream name directly,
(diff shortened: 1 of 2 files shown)


embano1:
Is there a reason you're opening two PRs for this?

https://github.com/cloudevents/sdk-go/pull/1319

Issue #1326, opened 2026-09-10
Support jitter for exponential backoff retry strategy


enriquezduane:
Currently, `v2/context/retry.go` provides `BackoffStrategyConstant`, `BackoffStrategyLinear`, and `BackoffStrategyExponential`.

When multiple concurrent clients retry against a failing or recovering receiver, pure exponential backoff causes retries to sync up at identical intervals, which can result in periodic request bursts (retry storms / ["thundering herd"](https://redis.io/blog/how-to-tame-the-thundering-herd-problem/)).

Adding jitter would spreads out retries so clients don't all hit the server at once. 

Reference:
* [AWS Architecture: Exponential Backoff And Jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/)

PR #1327, opened 2026-09-10
feat: add exponential backoff with full jitter retry strategy

enriquezduane:
Addresses [#1326](https://github.com/cloudevents/sdk-go/issues/1326)

This PR implements an exponential backoff with full jitter retry strategy (`BackoffStrategyExponentialWithJitter`) to prevent [thundering herd](https://redis.io/blog/how-to-tame-the-thundering-herd-problem/) issues during retries.

Changes Made:
- **Main Logic (`v2/context`)**: 
  - Added `BackoffStrategyExponentialWithJitter = "exponential-jitter"` constant.
  - Implemented full jitter calculation in `BackoffFor()` with guards against negative periods, float overflow, and sub-nanosecond intervals.
  - Added `d <= 0` guard in `Backoff()` to avoid `time.NewTicker(0)` runtime panics when jitter evaluates to 0, while properly preserving context cancellation checks.
  - Added `WithRetriesExponentialBackoffWithJitter` helper function.
- **Top-level API (`v2/alias.go`)**: 
  - Re-exported `ContextWithRetriesExponentialBackoffWithJitter`.
- **Protocol (`v2/protocol/http`)**: 
  - Added strategy to HTTP protocol retry dispatch.
- **Samples (`samples/http`)**: 
  - Updated `sender-retry` sample to demonstrate full jitter retry usage.
- **Tests**: 
  - Added unit tests covering edge cases (overflow, zero/negative periods, canceled context) and an HTTP protocol retry integration test.

For tests:
- Ran `cd v2 && go test -v -race ./context/... ./protocol/http/...`
- Ran `./hack/unit-test.sh` across all modules
- Ran `./hack/build-test.sh` across all modules
- Ran `go vet` and `gofmt -s`

### DIFF
--- a/samples/http/sender-retry/main.go
+++ b/samples/http/sender-retry/main.go
@@ -34,6 +34,8 @@ func main() {
 	send10(cloudevents.ContextWithRetriesLinearBackoff(ctx, 10*time.Millisecond, 10), c)
 	log.Println("--- Exponential ---")
 	send10(cloudevents.ContextWithRetriesExponentialBackoff(ctx, 10*time.Millisecond, 10), c)
+	log.Println("--- Exponential with Full Jitter ---")
+	send10(cloudevents.ContextWithRetriesExponentialBackoffWithJitter(ctx, 10*time.Millisecond, 10), c)
 }
 
 func send10(ctx context.Context, c cloudevents.Client) {
--- a/v2/alias.go
+++ b/v2/alias.go
@@ -150,11 +150,12 @@ var (
 
 	// Context
 
-	ContextWithTarget                    = context.WithTarget
-	TargetFromContext                    = context.TargetFrom
-	ContextWithRetriesConstantBackoff    = context.WithRetriesConstantBackoff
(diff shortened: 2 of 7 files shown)


embano1:
Thanks for the PR, @enriquezduane, and sorry for the slow response. The change looks good, and the context and HTTP retry tests pass locally.

One thing to think about, non-blocking. Full jitter has no upper bound on the delay, so with `Period: 10ms` the 20th retry can wait up to about 3 hours. The existing exponential strategy has the same gap, so this PR doesn't make it worse. A `MaxBackoff` field on `RetryParams` could cap both strategies in a follow-up.

enriquezduane:
Thanks for taking time to test and review, @embano1! I'd be glad to open an issue and followup PR to cap both exponential retry strategies once this lands.

 I've also rebased the branch onto latest `main`. Let me know if there's anything else you'd like me to adjust.

embano1:
Thx, all good - don't think we need a follow up on this right now.

embano1 (approved):

PR #1333, opened 2026-09-21
fix(http): stop receiver stalling under concurrent request cancellation

### LINKED ISSUE #1332: HTTP receiver permanently stalls under concurrent request cancellation
HTTP receiver permanently stalls under concurrent request cancellation (orphaned unbuffered-channel send)

## Version
Observed on `github.com/cloudevents/sdk-go/v2` v2.15.2 and v2.16.2. The relevant code is unchanged on `main` at time of writing; please verify.

## Summary
The HTTP protocol receiver routes every inbound request through a single, process-shared **unbuffered** channel (`Protocol.incoming`) using a **non-cancellable** send in `Protocol.ServeHTTP`. Each request also spawns a receiver goroutine that selects on `<-incoming` vs `<-ctx.Done()`(`Protocol.Respond`).

When a client cancels a request mid-flight, that request's receiver goroutine can take the `ctx.Done()` branch and return without ever draining `incoming` — **orphaning a concurrent request's queued send**. The orphaned sender is stuck forever on the bare channel send (it does not select on `ctx.Done()`), so its `ServeHTTP` never returns. Under sustained concurrent load with occasional cancellations, these orphaned sends accumulate: healthy, instant-handler requests block for their full deadline, server goroutines leak, and the endpoint effectively wedges — even though the application handler and network are healthy.

This is easy to trigger in practice: any client that imposes a per-request timeout (and thus cancels slow requests) while sending concurrently will, over time, wedge the receiver.

## Relevant code (v2.16.2)
- `protocol/http/protocol.go:109` — `incoming: make(chan msgErr)` (unbuffered, shared)
- `protocol/http/protocol.go:426` — `p.incoming <- msgErr{msg: m, respFn: fn}` (bare, non-cancellable send)
- `protocol/http/protocol.go:296-313` — `Respond` selects `<-p.incoming` vs `<-ctx.Done()` (returns `io.EOF` on cancel, without draining)
- `client/http_receiver.go:35-44` — `EventReceiver.ServeHTTP` spawns the per-request receiver goroutine, then calls `p.ServeHTTP`

## Reproduction
Instant handler; 16 concurrent workers; a quarter of them cancel a quarter of their requests ~immediately to force mid-flight cancellation. Healthy requests use a generous deadline against an instant handler, so any multi-hundred-ms latency is pure stall.

```go
package main_test

import (
	"bytes"
	"context"
	"errors"
	"net/http"
	"net/http/httptest"
	"sync"
	"sync/atomic"
	"testing"
	"time"

	"github.com/cloudevents/sdk-go/v2/event"
	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
)

func TestReceiverConcurrentCancelStall(t *testing.T) {
	p, _ := cehttp.New()
	h, err := cloudevents.NewHTTPReceiveHandler(context.Background(), p, func(_ event.Event) {})
	if err != nil {
		t.Fatal(err)
	}
	srv := httptest.NewServer(h)
	defer srv.Close()

	post := func(ctx context.Context) error {
		req, _ := http.NewRequestWithContext(ctx, http.MethodPost, srv.URL, bytes.NewReader([]byte("x")))
		req.Header.Set("Ce-Specversion", "1.0")
		req.Header.Set("Ce-Id", "id")
		req.Header.Set("Ce-Source", "example/uri")
		req.Header.Set("Ce-Type", "example.type")
		req.Header.Set("Content-Type", "application/octet-stream")
		resp, err := http.DefaultClient.Do(req)
		if err != nil {
			return err
		}
		return resp.Body.Close()
	}

	const workers = 16
	var stalled, ok int64
	var wg sync.WaitGroup
	deadline := time.Now().Add(10 * time.Second)
	for w := 0; w < workers; w++ {
		wg.Add(1)
		go func(w int) {
			defer wg.Done()
			for i := 0; time.Now().Before(deadline); i++ {
				if w%4 == 0 && i%4 == 0 { // force mid-flight cancellation
					ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
					_ = post(ctx)
					cancel()
					continue
				}
				ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
				start := time.Now()
				err := post(ctx)
				cancel()
				if errors.Is(err, context.DeadlineExceeded) || time.Since(start) > time.Second {
					atomic.AddInt64(&stalled, 1) // instant handler took ~the full deadline
				} else if err == nil {
					atomic.AddInt64(&ok, 1)
				}
			}
		}(w)
	}
	wg.Wait()
	t.Logf("ok=%d stalled=%d", ok, stalled)
	if stalled > 0 {
		t.Fatalf("%d healthy instant-handler requests stalled under concurrent cancellation", stalled)
	}
}
```

**Actual:** `stalled > 0`; goroutine dumps show handler goroutines permanently parked at `protocol/http/protocol.go:426` (`[chan send]`), e.g.:

```
goroutine NNN [chan send, 1 minutes]:
  ...protocol/http.(*Protocol).ServeHTTP ... protocol.go:426
  ...client.(*EventReceiver).ServeHTTP ... http_receiver.go:44
```

**Expected:** healthy requests complete promptly regardless of how many concurrent requests are cancelled; no goroutine leak.

## Root cause
The send on line 426 must be cancellable, or the receive/dispatch must not depend on a shared unbuffered channel whose sender and receiver can be desynchronized by per-request cancellation. Because the send is bare, a cancelled request's abandoned receiver leaves a permanent gap between the number of sends and receives on the shared channel.

## Possible fixes
- Make the send in `Protocol.ServeHTTP` select on `req.Context().Done()` (and the rate-limit / error sends similarly), so a cancelled request cannot leave an undeliverable send parked forever.
- And/or decouple per-request delivery from the single shared channel so one request's cancellation cannot orphan another's send.

(prepared using AI)

aliok:
Fixes #1332.

The HTTP protocol receiver routed every inbound request through a single, process-shared **unbuffered** channel (`Protocol.incoming`) using a **non-cancellable** send in `Protocol.ServeHTTP`, while `EventReceiver.ServeHTTP` spawned a per-request goroutine that consumed the channel via `Protocol.Respond` (which selects on `ctx.Done()`).

When a client cancels a request mid-flight, that request's receiver goroutine can take the `ctx.Done()` branch and return **without draining** the channel, orphaning a *concurrent* request's queued send. The orphaned sender blocks forever on the bare channel send (it does not select on `ctx.Done()`), so its `ServeHTTP` never returns. Under sustained concurrent load with occasional cancellations these orphaned sends accumulate: healthy, instant-handler requests block for their full deadline, server goroutines leak, and the endpoint wedges — even though the application handler and the network are healthy.

## The change

Deliver each request to the invoker **inline within the request goroutine** instead of handing it across the shared channel:

- `Protocol.ServeHTTP` keeps its existing `Receive`/`Respond` channel behavior unchanged (the polling consumer API).
- New `Protocol.ServeHTTPWithHandler(rw, req, handle)` parses the request — preserving rate limiting and the `GET`/`OPTIONS`/`DELETE` handlers — and calls `handle` inline. Both methods share a single `serveHTTP` implementation.
- `EventReceiver.ServeHTTP` now calls `ServeHTTPWithHandler` with `invoker.Invoke`, dropping the extra goroutine and the `Respond` channel round-trip.

With no cross-request channel handoff, one request's cancellation can no longer strand another request's delivery.

## Test

Adds `client/receiver_cancel_test.go`: concurrent workers, a fraction of them forcing mid-flight cancellation, asserting that no healthy instant-handler request hits its deadline. The test **wedges (times out) on the previous code** and passes with this change. Full `go test ./...` on the `v2` module is green.

## Notes

- Public API addition: `Protocol.ServeHTTPWithHandler`. `ServeHTTP` and the `Receive`/`Respond` channel path are behaviorally unchanged.
- Scope: this fixes the `EventReceiver`-as-`http.Handler` usage. The default `Client.StartReceiver`/`OpenInbound` path uses the same unbuffered channel but is **not** subject to #1332 — its consumers are long-lived rather than per-request, so a cancelled request can't orphan a concurrent send there. Left unchanged deliberately.


embano1:
Thanks for the detailed report and the fix, @aliok. Delivering each request inline is the right approach. I confirmed locally that `TestReceiverConcurrentCancelStall` hangs on the old code and passes with this change, including repeated runs under `-race`.

A few non-blocking nits.

1) The two new files carry a 2021 copyright header. Newer files in the repo use the year they were added, so 2026 fits here.
2) The doc comments on `ServeHTTPWithHandler` and in `serveHTTP` are long. Trimming them to the contract (call `respFn` exactly once, before the method returns) makes them easier to keep accurate.
3) On a regression, `TestReceiverConcurrentCancelStall` hangs until the test binary times out, because `srv.Close` waits on the stuck handlers. A bounded wait that calls `t.Fatal` would fail faster and point at the cause.

